PR #120, opened 2019-10-18
Log server version

dev-gto:
When running different krill instances on several machines using either docker or rsyslog, it is useful to log each version at startup, as well as their configuration filenames

### DIFF
--- a/src/daemon/config.rs
+++ b/src/daemon/config.rs
@@ -17,9 +17,8 @@ use rpki::uri;
 use crate::commons::api::Token;
 use crate::commons::util::ext_serde;
 use crate::daemon::http::ssl;
-
-pub const APP_NAME: &str = "NLnet Labs RRDP Server";
-pub const APP_VERSION: &str = "0.2.1";
+use crate::daemon::version::APP_NAME;
+use crate::daemon::version::APP_VERSION;
 
 //------------ ConfigDefaults ------------------------------------------------
 
--- a/src/daemon/config.rs
+++ b/src/daemon/config.rs
@@ -18,6 +18,9 @@ use crate::commons::api::Token;
 use crate::commons::util::ext_serde;
 use crate::daemon::http::ssl;
 
+pub const APP_NAME: &str = "NLnet Labs RRDP Server";
+pub const APP_VERSION: &str = "0.2.1";
+
 //------------ ConfigDefaults ------------------------------------------------
 
 pub struct ConfigDefaults;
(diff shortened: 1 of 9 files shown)


src/daemon/config.rs
```diff
@@ -215,6 +217,12 @@ impl Config {
         let config_file = matches
             .value_of("config")
             .unwrap_or("./defaults/krill.conf");
+        format!("{}", config_file)
+    }
+
+    /// Creates the config (at startup). Panics in case of issues.
+    pub fn create() -> Result<Self, ConfigError> {
+        let config_file = &*Self::get_config_filename();
```
timbru: Why do we need &* here?

timbru:
Thank you for the pull request.

Could you move the constants to src/lib.rs and have only one VERSION constant? And change the name for the krill app. I.e. add the following:

```
pub const KRILL_VERSION: &str = "0.2.1";
pub const KRILL_SERVER_APP: &str = "Krill";
pub const KRILL_CLIENT_APP: &str = "Krill Client";
```

Other than that I left a comment regarding the use of '&*'. Can't that just be left out?

I am happy to merge it when these things are changed.

src/daemon/config.rs
dev-gto: actually I'm a newbie in rust world... I will fix this

dev-gto:
I've just finished the changes... Please let me know if there are any issues left.

src/lib.rs
```diff
@@ -1,3 +1,7 @@
+pub const KRILL_VERSION: &str = "0.2.1";
+pub const KRILL_SERVER_APP: &str = "NLnet Labs RRDP Server";
```
timbru: Can you name the server "Krill"

src/daemon/krillserver.rs
```diff
@@ -82,6 +84,8 @@ impl KrillServer {
         let token = &config.auth_token;
         let ca_refresh_rate = config.ca_refresh;
 
+        info!("{} v{}", KRILL_SERVER_APP, KRILL_VERSION);
```
timbru: Can you change these three lines to:
```
info!("Starting {} v{}", KRILL_SERVER_APP, KRILL_VERSION);
info!("{} uses configuration file: {}", KRILL_SERVER_APP, Config::get_config_filename());
info!("{} uses service uri: {}", KRILL_SERVER_APP, service_uri);
```

dev-gto: ok, done

timbru:
Thanks! I will merge as soon as the Travis build comes through. (network in commuter train permitting).

PR #179, opened 2020-01-15
Base the Docker image on Alpine Linux.

ximon18:
For a smaller Docker image and consistency with NLnet Labs Routinator.
Image size is reduced from almost 200 MiB to < 50 MiB.
Removing the `entrypoint.sh` script and its `bash` and `util-linux` dependencies would reduce it even more.

Before merging:
- [ ] Merge https://github.com/NLnetLabs/rpki-deploy/pull/20


ximon18:
Merging of NLnetLabs/rpki-deploy#20 is no longer needed as an updated base image has been pushed to Docker Hub, instead that branch can be discarded. Basic `krillc` commands using `krillc` from within the Alpine image against `krill` daemon in the Alpine image work. As discussed with @timbru I will merge this branch.

PR #181, opened 2020-01-21
Move the Krill E2E tests from the rpki-deploy repo to the Krill repo.

ximon18:
Merge https://github.com/NLnetLabs/rpki-deploy/pull/21 before merging this PR.

This PR moves the Krill E2E tests from the rpki-deploy repo to the Krill repo. Note: The Python file will not work on its own, it depends on being executed by the Krill `main.yml` GitHub Actions workflow as part of the E2E test suite defined in the rpki-deploy repo.

Python code has been refactored such that framework / helper / util code is in the rpki-deploy repo and the test logic is in this repo. Functionally, there should be no change compared to the current E2E test run.

The current mechanism for using the tests is simply to copy any files in the defined test directory from the Krill checkout to the rpki-deploy checkout, such that PyTest `conftest.py` special behaviour works as expected. There might be a cleverer or more Pythonic way to do this but this works for now.

Before merging:
- [x] See the E2E test work in this branch vs the branch in the rpki-deploy repo.
- [x] Merge https://github.com/NLnetLabs/rpki-deploy/pull/21


.github/workflows/main.yml
```diff
@@ -31,6 +31,7 @@ jobs:
       with:
         repository: nlnetlabs/rpki-deploy
         path: rpki-deploy
+        ref: externalize-tests
```
ximon18: This change is just temporarily to test against the as yet unmerged branch in the rpki-deploy repo. This line should be removed before merging this PR.

ximon18:
The create if not already existing approach is a legacy from when the tests were running in Bash and I ran them repeatedly locally without destroying the Docker container. Whether or not this approach is needed now depends on the lifetime of ur Krill container that is desirable to test against, I’m fine with assuming a clean Krill each time and terminated and recreated Krill (Matthew with a different config file) as needed.

timbru:
I am also fine with leaving this until later, I don't think it's a blocking issue.

ximon18:
If it’s readable as is let’s leave it for now so I can merge it and do it other stuff :-)

ximon18:
I’d like to extract the RP version detecting code but that can also wait. Basically either run an explicit version grabbing test first, or capture versions in any test where that RP version has not already been captured, the latter works better if we ever start running tests selectively.

PR #183, opened 2020-01-22
Use matching E2E test framework branch if exists

ximon18:
Each time a new Krill repo branch is made that changes something that depends on a modification in the E2E test framework in the rpki-deploy repo the Krill `main.yml` file has to be temporarily changed in the PR to reference a corresponding branch in the rpki-deploy repo, and changed back before the PR merges (thus breaking the E2E test in the PR checks report).

This change attempts first to checkout the E2E test framework from a branch by the same name as the current Krill branch, and only if that fails checkout the master E2E test framework branch.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,6 +32,7 @@ jobs:
         fetch-depth: 1
 
     - name: Checkout the E2E test framework (matching branch if exists)
+      continue-on-error: true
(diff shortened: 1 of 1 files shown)


ximon18:
Looks good. Tested by creating a dummy branch in the rpki-deploy repo and re-running the E2E test job and it then checked out the dummy branch instead of master.

PR #214, opened 2020-03-11
Enable support for CTRL-C from outside the container.

ximon18:
Same fix as proposed for Routinator in https://github.com/NLnetLabs/routinator/pull/277.
I'm not sure if `krill` needs this but certainly `krillc` when run via this container needs it.
The "proper" fix is to handle the signal sent to the process, though Tini also handles other "pid 1" responsibilities as well (e.g. reaping finished background processes).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,8 +51,7 @@ EXPOSE 3000/tcp
 # Use Tini to ensure that krillc responds to CTRL-C when run in the
 # foreground without the Docker argument "--init" (which is actually another
 # way of activating Tini, but cannot be enabled from inside the Docker image).
-ADD https://github.com/krallin/tini/releases/download/v0.18.0/tini /tini
-RUN chmod +x /tini
-
-ENTRYPOINT ["/tini", "--", "/opt/entrypoint.sh"]
+RUN apk add --no-cache tini
+# Tini is now available at /sbin/tini
+ENTRYPOINT ["/sbin/tini", "--", "/opt/entrypoint.sh"]
 CMD ["krill", "-c", "/var/krill/data/krill.conf"]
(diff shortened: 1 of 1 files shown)


ximon18:
@wk: any thoughts on this?

ximon18:
One example of where this makes a difference is when running the `history` command via `krillc` in the container. Without this change you cannot interrupt the command and must wait which could take a long time:
```
$ docker run --rm -it -e KRILL_CLI_TOKEN=2700DB12-D463-4E08-8CBA-4300525A56C0 -e KRILL_CLI_SERVER=https://ca.rpki.nlnetlabs.nl/ nlnetlabs/krill:v0.5.0-pre krillc history --ca ca
^C^C^Cid: ca version: 0 details: Initialised with ID key hash:
...
```

Note the multiple ignored "^C" attempts. With this change you _can_ interrupt the command:
```
$ docker run --rm -it -e KRILL_CLI_TOKEN=2700DB12-D463-4E08-8CBA-4300525A56C0 -e KRILL_CLI_SERVER=https://ca.rpki.nlnetlabs.nl/ nlnetlabs/krill:ctrlc krillc history --ca ca
^C
$ 
```

ximon18:
@timbru: FYI @wk linked to something for Rust on this topic: (from Routinator PR #277):

> There's a pretty clean manner to handle this in-process in Rust [documented here](https://piers.rocks/rust/signal/sigint/container/docker/2019/01/06/rust-catch-ctrl-c-container.html).

timbru:
I saw the comment for routinator but I also remember that Martin said something about this being supported by the new Tokio. That would probably be cleaner.

Please feel free to add an issue for this. Then I can see if I can pick this up after my holidays.

ximon18:
Thx, but I wonder if this makes any sense in Krill or Routinator themselves. The issue AFAIK doesn't exist unless they are run as PID 1, e.g. inside a Docker container, for which there exist Docker specific solutions such as `--init` and `tini` and friends.

PR #269, opened 2020-06-26
Proper pkg naming

ximon18:
These GitHub Actions Workflow changes ensure that packages for different releases of the same O/S (e.g. Debian or Ubuntu) are unique in their name/version/arch triple, as required by [Debian policy](https://www.debian.org/doc/debian-policy/ch-controlfields.html#version) and enforced by [Aptly](http://aptly.info/) when managing a conforming repo.

Without these changes we cannot store the different flavours of Krill DEB that we produced on master for Krill v0.7.1 in a policy compliant APT repo server as the triples of the 5 different packages (Ubuntu 16.04/18.04.20.04 and Debian 9/10) are all the same. These physically cannot be stored in a policy compliant repo because the packages are stored by unique triplet in a common 'pool' directory irrespective of which 'distribution' they are served for (e.g. 16.04 or 18.04) and so conflict with or overwrite each other if the triplet is the same for each package.

The packages produced can safely be installed by users who already previously installed the packages created for Krill v0.7.1 from master, doing so will treat the installation as an upgrade and if done using a repo URL the users can then in future do `apt update` to get the newest version (assumes that we publish the created DEBs to a repo server of course).

As a proof-of-concept the packages already produced by this branch have been published to the PoC repo server.

### DIFF
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -148,7 +148,7 @@ jobs:
           debian:10)    OS_REL=buster ;;
           *)            echo 2>&1 "ERROR: Unexpected matrix image"; exit 1 ;;
         esac
-        DEB_VER="krill_${KRILL_VER}-1${OS_REL}_amd64"
+        DEB_VER="${KRILL_VER}-1${OS_REL}"
         cargo deb --variant $DEB_NAME --deb-version $DEB_VER
       env:
         MATRIX_IMAGE: ${{ matrix.image }}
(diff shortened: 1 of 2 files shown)


ximon18:
jq is the standard linux tool for json command line parsing, using shell script or sed or awk is less readable and more error prone, so I would stick with jq for now. If you have no other issues I suggest we merge this.

timbru:
Works for me! Feel free to merge!

PR #276, opened 2020-07-02
Do as Debian advises

ximon18:
Use Debian package maintainer script systemd fragments from Debian dh_installsystemd autoscripts instead of home-grown commands.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,6 +97,7 @@ assets = [
     ["defaults/krill.conf", "/usr/share/doc/krill/krill.conf", "644"],
     ["doc/krill.1", "/usr/share/man/man1/krill.1", "644"],
     ["doc/krillc.1", "/usr/share/man/man1/krillc.1", "644"],
+    ["debian/assets/krill.service.preset", "/lib/systemd/system-preset/50-krill.preset", "644"],
     ["debian/assets/krill.service.ubuntu2004", "/lib/systemd/system/krill.service", "644"]
 ]
 
@@ -110,6 +111,7 @@ assets = [
     ["defaults/krill.conf", "/usr/share/doc/krill/krill.conf", "644"],
     ["doc/krill.1", "/usr/share/man/man1/krill.1", "644"],
     ["doc/krillc.1", "/usr/share/man/man1/krillc.1", "644"],
+    ["debian/assets/krill.service.preset", "/lib/systemd/system-preset/50-krill.preset", "644"],
     ["debian/assets/krill.service.ubuntu1804", "/lib/systemd/system/krill.service", "644"]
 ]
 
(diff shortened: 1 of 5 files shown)


ximon18:
I noticed another issue with the unit files which is [affecting master](https://github.com/NLnetLabs/krill/runs/831032193?check_suite_focus=true#step:11:242) and is still the case in this branch: `systemctl status` shows the unit as `vendor preset: enabled` but we don't enable it, we want it initially disabled so this is confusing. Let's see if we can improve this as well...

ximon18:
The vendor preset issue appears [to be fixed now](https://github.com/NLnetLabs/krill/pull/276/checks?check_run_id=831257549#step:11:241) by this PR.

ximon18:
TO DO: Add a table to this PR showing `systemctl status` state changes before and after this PR while moving from clean system to installed, removed then purged, and also on a system that previously had the released .deb installed that has the warning this PR was originally created to remove.

ximon18:
So, `systemctl status` on `apt remove` is "masked" and that seems to be correct.

ximon18:
So, actually, we need to consider what the maintainer scripts will do on first install, install after remove, and upgrade after install, according to the [lifecycle diagrams here](https://www.debian.org/doc/debian-policy/ap-flowcharts.html).

On first install we do NOT want Krill to be started because the user has to review the config file and probably set `service_uri`. Maybe we should force this by giving the config file a name like `/etc/krill.conf.template` or something so that the user has to rename it to `/etc/krill.conf` before the service will start? Currently on first install in `postinst configure` we create a user and config file if needed but do not start Krill.

On install after remove, I think we should not start Krill as it wasn't running prior to install and just because the config file may exist (as it isn't removed until apt purge) that doesn't mean the config file is configured yet according to the users wishes. Currently on install after remove we behave the same way as on first install but due to the existence of a config file none is generated.

On upgrade after install, we should keep the existing config file, and if the service is running, stop it and restart it (to use the upgraded binary). Currently in `prerm upgrade` we stop the Krill service if it was running, and in `postinst configure` we ensure that if the service was enabled that it remains enabled. We do **NOT** start the service again after stopping it. If you want an upgrade to minimize downtime this is probably not the right behaviour, we **SHOULD** start Krill again.

ximon18:
For the record this is a good read: https://linux.m2osw.com/how-prevent-service-installed-debian-system-automatically-being-enabled. According to that article because our unit file contains `WantedBy=multi-user.target` that means that systemd will force Krill to become enabled and start on boot even though we don't want it to be enabled and started until the user enables it manually. Is this really what happens?

Another unrelated TO DO item: use the lifecycle diagram to understand and check what will happen when upgrading from the currently released package to this new package version with modified maintainer scripts.

ximon18:
Obsoleted by PR #288.

PR #278, opened 2020-07-02
Run on branch or PR, not both

ximon18:
See: https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662


ximon18:
Looks good:
image
No duplicate packaging invocations there.

PR #286, opened 2020-07-20
Expose Krill server major.minor.patch version as gauge metrics

ximon18:
Note: also changes the initialisation of the `KRILL_VERSION` constant to use Cargo compile-time environment variable [`CARGO_PKG_VERSION`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) because then using the similar `_MAJOR`, `_MINOR` and `_PATCH` compile-time env vars is easy.

The alternative is to use something like [`std::sync::Once` ](https://doc.rust-lang.org/std/sync/struct.Once.html) and [`semver_parser::version::parse()`](https://docs.rs/semver-parser/0.9.0/semver_parser/version/fn.parse.html) which is overkill when the values are available during compilation as Cargo env vars AND that removes the need to manually keep the constant in sync with the Cargo.toml version key value.

With this PR the following metric output is added:

```
# HELP krill_version_major krill server major version number
# TYPE krill_version_major gauge
krill_version_major 0

# HELP krill_version_minor krill server minor version number
# TYPE krill_version_minor gauge
krill_version_minor 8

# HELP krill_version_patch krill server patch version number
# TYPE krill_version_patch gauge
krill_version_patch 0
```

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -1,4 +1,7 @@
-pub const KRILL_VERSION: &str = "0.8.0-RC-dev";
+pub const KRILL_VERSION: &str = env!("CARGO_PKG_VERSION");
+pub const KRILL_VERSION_MAJOR: &str = env!("CARGO_PKG_VERSION_MAJOR");
+pub const KRILL_VERSION_MINOR: &str = env!("CARGO_PKG_VERSION_MINOR");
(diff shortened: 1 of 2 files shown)


ximon18:
Confirmed with @timbru that this can be merged.

PR #287, opened 2020-07-23
Support 'pre-release' build labels in .deb packages

ximon18:
Correctly handle dashes in the Krill version number so that 'pre-release' 'tags' appended to the version number do not cause the package to be considered newer than a final untagged version number.

### DIFF
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -60,6 +60,9 @@ jobs:
         ]
     env:
       CARGO_DEB_VER: 1.23.1
+      # A Krill version of the form 'x.y.z-plus' denotes a dev build that is
+      # newer than the released x.y.z version but is not yet a new release.
+      NEXT_VER_LABEL: plus
     name: deb-pkg
     runs-on: ubuntu-latest
     # Build on the oldest platform we are targeting in order to avoid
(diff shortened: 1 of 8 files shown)


ximon18:
This PR is ready to merge to master.

However,  currently the new "upgrade" test jobs in the pkg build will fail because the master Krill Cargo.toml `version` (0.7.3) is the same as the last release version (0.7.3) and attempting to install the new package over the old one causes `apt` to complain about an unexpected downgrade. While the version numbers are the same, a reinstall of the exact same package skips the already installed package so in this case apt sees that the version is the same but the binary package is not (as it contains new commits from master). Why it treats this as a downgrade I don't know, but the "fix" would be to put `0.7.3-plus` or something in Cargo.toml on master instead of the same version as was last released.

@timbru: How do you want to proceed? We can just merge and ignore the failing upgrade tests, but they are clearly warning about ever letting those packages out into the wild, it might be wise to fix the issue by making sure the Cargo.toml version is distinct from that that was released.

ximon18:
Agreed with @timbru to set `version` in Cargo.toml to `0.7.3-plus` as plus is a convention that the project has already been using and denotes "next" without saying what version that will be, only that master is higher/newer than the last release. This is better than using "0.7.3-dev" as that could be understood to mean working toward a 0.7.3 release. Using "0.7.4-dev" or "0.8.0-dev" instead avoids that problem but then you have to already decide what the next release number will be. This can then be merged.

Issue #275, opened 2020-07-02
systemd warning after apt remove


ximon18:
With Krill v0.7.2 on Ubuntu 18.04 LTS:

```
$ sudo apt remove krill -y
...
$ sudo systemctl status krill
Warning: The unit file, source configuration file or drop-ins of krill.service changed on disk. Run 'systemctl daemon-reload' to reload units.
```

After running the suggested command the warning is gone.

Looking at the nginx-common package I see with `dpkg -e` that the `postrm` script does:
```
# In case this system is running systemd, we make systemd reload the unit files
# to pick up changes.
if [ -d /run/systemd/system ] ; then
        systemctl --system daemon-reload >/dev/null || true
fi
```

These lines were added by `dh_installsystemd` invoked by debhelper when preparing the NGINX package. You can see this by looking at the patches attached to [this debhelper improvement tracker issue](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=690399). The actual debhelper autoscript that adds these lines [can be seen here](https://salsa.debian.org/debian/debhelper/-/blob/master/autoscripts/postrm-systemd-reload-only).

As Cargo Deb doesn't take care of it for us, we should probably do whatever [dh_installsystemd](https://salsa.debian.org/debian/debhelper/-/blob/master/dh_installsystemd) _(debian source code)_ does in its [autoscripts](https://salsa.debian.org/debian/debhelper/-/tree/master/autoscripts) _(debian source code)_. Even better would be if Cargo Deb did this itself, somewhat like [fpm](https://github.com/jordansissel/fpm) does something like this.

It might also be worth looking at [how fpm was extended to support systemd](https://github.com/jordansissel/fpm/pull/952/files). It also does daemon-reload as part of package removal.

ximon18:
After a lot of ... effort, I managed to get `dh_installsystemd -v` on the Krill package running on Ubuntu 18.04 LTS to generate maintainerscript additions for systemd actions. The end result was that it generated these files:

**debian/.debhelper/generated/krill/postinst.service:**
```
# Automatically added by dh_installsystemd/11.1.6ubuntu2
if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ] ; then
	# This will only remove masks created by d-s-h on package removal.
	deb-systemd-helper unmask 'krill.service' >/dev/null || true

[...]

PR #288, opened 2020-07-23
Deb pkg fixes

ximon18:
Work in progress. Includes the `rc-deb-support` branch changes.


ximon18:
Current status
- Issue #275: not reproducible with 0.7.3 or with this PR.
- Issues identified in PR #276:
  - Incorrect systemd unit vendor preset: fixed
  - Introduced systemd unit masking.
  - On first install we do NOT want Krill to be started: **FIXED** with this PR
  - On install after remove, I think we should not start Krill: **NOT** FIXED with this PR
  - On upgrade after install, we should keep the existing config file, and if the service is running, stop it and restart it (to use the upgraded binary): **FIXED** with this PR

ximon18:
> On install after remove, I think we should not start Krill: NOT FIXED with this PR
This does not seem to be supported by the current debhelper dh_installsystemd autoscript fragments and dh_installsystemd logic, 
 and would be difficult and hacky to add while using them. As it's a non-trivial corner case I'm going to ignore it for now.

ximon18:
Lintian complains with:
```
E: krill: debian-changelog-file-missing
W: krill: copyright-without-copyright-notice
```

I'll take a quick look at these. Cargo deb has options for managing the changelog, I'll quickly look at that, but we don't want to manually maintain a changelog for the deb package as well as in GitHub release notes, so maybe we can just point to the GH release notes somehow?

PR #299, opened 2020-08-04
Test minimal systemd Cargo.toml syntax

ximon18:
(and restart after upgrade = true)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,13 +99,7 @@ assets = [
     ["doc/krillc.1", "/usr/share/man/man1/krillc.1", "644"],
     ["debian/krill.service.preset", "/lib/systemd/system-preset/50-krill.preset", "644"],
 ]
-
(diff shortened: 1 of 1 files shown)


ximon18:
@timbru: This change simplifies the Cargo.toml for the systemd configuration for the .deb package, but it also changes the logic used by the upgrade process to the Debian modern default of leaving the process running until after the package upgrade is complete, then restarting it, rather than the current legacy approach of stopping the process, upgrading the package then restarting the process. Do you see any issues with the Krill binaries being replaced on disk while the Krill service is running? Currently the package doesn't touch any other files AFAIK and the only possible issue with the modern approach is if the upgrade changes something on disk that the running process doesn't expect/can't handle. Any thoughts on this?

ximon18:
Discussed and agreed with @timbru that indeed Krill should be fine with restart after package upgrade rather than stop, upgrade, start.

Issue #306, opened 2020-08-18
Add testbed TA and repo UI for easier testing


timbru:
So that we can run a service where Krill users can just register their test Krill instance as a child and publisher.

ximon18:
Handy links for review of the initial POC:
- https://github.com/NLnetLabs/krill/compare/0.8.0-RC-dev...testbed-poc
- https://github.com/NLnetLabs/lagosta/compare/testbed-poc
- https://testbed-poc.aws.nlnetlabs.nl/#/testbed

Review feedback gathered so far from @timbru and @AlexanderBand:
- [ ] Don't link from the top-left Krill logo to the Krill root page.
- [ ] Add a disclaimer that this is for testing, established CA/repo parent-child relationships may disappear at any time and can be deleted by anyone. Should also cover that you can claim in this test hierarchy any resources with ROAs, even those of other people, as being able to claim real ROAs permits testing of the BGP integration in Krill
- [ ] Make any not-yet-translatable strings translatable.
- [ ] Blog about it and link to the blog from the testbed page (to explain that it is a separate test RPKI hierarchy for example)
- [ ] Use a dedicated deployment (e.g. on AWS) with its own monitoring (can re-use current Prometheus and Grafana instance but have separate Grafana dashboard) and logging (use a new AWS S3 bucket)
- [ ] Final deployment should be at https://testbed.rpki.nlnetlabs.nl/ (redirecting to /#/testbed)
- [ ] Make the page layout it a bit wider (even on a 960 wide browser window it becomes too narrow).
- [ ] Grant the testbed CA all resources:
```
asns = "0-4294967295"
v4 = "0.0.0.0-255.255.255.255"
v6 = "::0/0"
```
- [ ] Final deployment should be at https://testbed.rpki.nlnetlabs.nl/ (redirecting to /#/testbed)
 Make the TAL available with filename `test.tal` so that relying party software such as Routinator that report the Trust Anchor name based on the filename report a more useful name than 'ta'.

PR #308, opened 2020-08-24
Add testbed mode

ximon18:
Adds support for a new `testbed_enabled` boolean flag to configure Krill to permit child/parent CA/repo RFC-8183 relationships to be established by *anonymous* clients with a CA called `testbed` and enables a `/#/testbed` UI page for doing so visually ala other RPKI testbeds such as the APNIC RPKI testbed. Note: the `testbed` CA must be created manually, it is **not** created as part of this PR Implements the s/w parts of issue #306.


ximon18:
Note: merge to master should probably wait on merge of the Lagosta `testbed-poc` branch to Lagosta master first. That branch is included (in release compiled form) in this Krill branch.

timbru:
> Note: merge to master should probably wait on merge of the Lagosta `testbed-poc` branch to Lagosta master first. That branch is included (in release compiled form) in this Krill branch.

I don't mind this.. we are merging to a dev branch and can update lagosta again later.

Also the failing test may be due to the issue with the message queue scheduling fixed in '9b46f5ab60b49485d6e18e6770e0478c73e52b85'

So, it's probably fine to mere.

ximon18:
Thanks @timbru. FYI this work currently lacks any modification of or addition to the automated test suite. I should probably extend it at least to check that the disabling of authorisation checks only happens as and when it should, right?

timbru:
> Thanks @timbru. FYI this work currently lacks any modification of or addition to the automated test suite. I should probably extend it at least to check that the disabling of authorisation checks only happens as and when it should, right?

That sounds like a really good idea :)

Other than that it would also be nice to have an end to end test for this so we can catch regressions.

ximon18:
The CI build failure ([`reached the type-length limit while instantiating`](https://github.com/NLnetLabs/krill/pull/308/checks?check_run_id=1020520324#step:4:663)) is only on Rust beta, not on Rust stable, so will ignore for now and assume it will soon solve itself.

src/daemon/http/server.rs
```diff
@@ -1285,6 +1261,61 @@ async fn api_ca_rta_oneoff(req: Request, path: &mut RequestPath, ca: Handle) ->
     }
 }
 
+//------------ Support acting as a testbed -------------------------------------
+
+async fn testbed(req: Request) -> RoutingResult {
+    if !CONFIG.testbed_enabled {
+        render_not_found(req).await
```
timbru, line 1268: Use `Err(req)` so that the req gets returned and routing continues.

src/daemon/http/server.rs
```diff
@@ -1285,6 +1261,61 @@ async fn api_ca_rta_oneoff(req: Request, path: &mut RequestPath, ca: Handle) ->
...
+    } else {
+        let mut path = req.path().clone();
+        match path.next() {
+            Some("enabled")    => testbed_enabled(req).await,
+            Some("children")   => testbed_children(req, &mut path).await,
+            Some("publishers") => testbed_publishers(req, &mut path).await,
+            _                  => render_not_found(req).await,
```
timbru, line 1275: The api uses `render_unknown_method()` whenever the URI maps to a specific part of the API, but there is no method to map to. As opposed to `render_unknown_resource()` when the path is okay but the actual resource at that location cannot be found (e.g. a CA does not exist). The first results in a 400 response, the second in a 404 with a json error, whereas `render_not_found(req)` results in a 404 with HTML which is intended to be browser user facing.

Whether this is all the best way of doing things is of course debatable, but we should still use `render_unknown_method()` here to be consistent.

ximon18:
TODO: update to use Lagosta master as the Lagosta testbed-poc branch has been merged.

ximon18:
Agreed with @timbru to merge this branch.

PR #313, opened 2020-09-02
Fix FORT Validator incid-hashalg-has-params incidence check failure.

ximon18:
Enabling the FORT Validator [`incid-hashalg-has-params` incidence check](https://nicmx.github.io/FORT-validator/incidence.html#signed-objects-hash-algorithm-has-null-object-as-parameters) currently causes the Krill end-to-end test to fail if enabled.

The [underlying cause has been fixed](https://github.com/NLnetLabs/rpki-rs/commit/8101ee966fa27d3929d3ac8ce977ade288b59db2) in the `rpki` Rust crate dependency used by Krill.

This PR updates Krill with that fix but the fix is not yet available in a release version of the `rpki` crate and so this is currently only a DRAFT PR. Once a release version of the rpki crate with the fix is available then Cargo.toml should be updated in this PR to reference the new rpki crate dependency version and then this PR can be merged.

When this change has been merged the Krill end-to-end test should be updated to enable the FORT Validator `incid-hashalg-has-params` incidence check. See https://github.com/NLnetLabs/rpki-deploy/pull/27.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ openssl         = { version = "^0.10", features = ["v110"] }
 pretty          = "0.5.2" # for testing
 rand            = "^0.5"
 reqwest         = { version = "0.10.8", features = [ "json"] }
-rpki            = { git = "https://github.com/NLnetLabs/rpki-rs.git", branch = "rta-01", version = "0.9.2-bis" }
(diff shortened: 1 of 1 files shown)


timbru:
Fixed in rpki-rs 0.10.0 used in Krill 0.8.0

PR #368, opened 2021-01-06
Release 0.9.0-rc1

ximon18:
The Krill 0.8.1-bis development branch contains the so-called "multi-user" support. This PR is intended to facilitate review of that functionality as a step toward **_beta_** release.

### A note on branching strategy

The multi-user branch was already merged for pragmatic reasons to the 0.8.1-bis branch, the set of changes includes **_ALL_** changes to Krill made in the development branch, not just the "multi-user" changes. However, this PR is only intended at this point for review of the "multi-user" changes, not the other changes that exist in the 0.8.1-bis branch.

### Feature gate

The new "multi-user" functionality is "hidden" behind a Rust feature flag. By default builds of Krill do not include the new "mutli-user" functionality. To enable it build either with `cargo build --features multi-user` or `cargo build --all-features`.

### What is "multi-user" support?

The "multi-user" support extends Krill to support:
- **Authentication** of multiple different end users, not just the single master-token based authentication that exists in all versions of Krill up to the latest v0.8.2 release. This includes showing the logged in user in the web UI and attributing actions in the Krill event history to the logged in user (or as before to the internal "krill" user if the action were initiated by Krill itself).
- More fine grained **Authorisation** of the requested action than the authenticated-or-not approach supported upto and including Krill v0.8.2.

I will refer to Authentication and Authorisation collectively as "Auth".

The implementation of these features adds:
- The notion of an Actor which has certain rights and whose id/name is recorded in the event history.
- Users and user metadata defined inline in the Krill configuration file via the new **ConfigFileAuthProvider**.
- Users and user metadata defined in a 3rd party system that implements the OpenID Connect Core 1.0 specification via the new **OpenIDConnectAuthProvider**. Supports regular expression and JMESPath based matching of provider "claim" data in order to select e.g. email or first name or friendly name etc as the users ID, and to "lookup" the users rights (e.g. their "role" or "team" membership) in the provider metadata.
- Configurable and extensible authorisation policy defined in terms of [Oso](https://www.osohq.com/) policy rules which can optionally consider values of arbitrary user metadata properties. For example users can be denied access to certain CAs, or permitted to read but not write, or to work with CAs but not to create ROAs. Users can be grouped by role or by team or both or by whatever scheme matches your needs.
- Logout of Krill itself (and the 3rd party provider in the case of OpenID Connect), rather than just forgetting the login token in the web UI as was done until now.
- Server-side login session expiration (and also browser-side logout) e.g. based on OpenID Connect token expiration time, rather than just browser-side logout after a period of inactivity.
- [Cypress](https://www.cypress.io/) based UI driven automation tests, for end-to-end testing of the auth related features _(**note:** requires Docker and is not yet integrated into the GitHub Actions continuous-integration setup)_.
- A mock OpenID Connect provider for automation testing of the OpenID Connect functionality.
- Continued support for master token authentication for `krillc` and 3rd party API clients even if Krill is configured to use config file or OpenID Connect based auth. Note: `krillc` **only** supports master token authentication, there is no support yet e.g. for OpenID Connect device flow based authentication for `krillc`.

### Status

- With the exception of one UI test ("Add ROA for ...") that is currently unstable (I am working to address this) all `cargo test` tests pass both with and without the "multi-user" feature enabled.
- Security aspects of the code require review and possible further implementation (e.g. OpenID Connect CSRF token checking, the nonce and AAD values used for session encryption).

### Where to start?

Note: The implementation is designed to be stateless on the Krill server side. We may in future choose to move session state to the server. State when needed is currently stored encrypted in the bearer token issued to and passed back to Krill by the Lagosta UI.

- End user documentation does not exist yet but will be added to the [krill-0.9 branch](https://github.com/NLnetLabs/rpki-doc/tree/krill-0.9) in the rpki-doc repository.
- All of the UI changes are already in the [Lagosta master branch](https://github.com/NLnetLabs/lagosta).
- `defaults/krill-multi-user.conf` documents the additional configuration file options that relate to the multi-user feature.
- `defaults/*.polar` define the built-in Oso authorisation policy rules and contain a **LOT** of inline sanity-check self-tests.
- `fn map_requests` in `src/daemon/http/server.rs` is the starting point for handling of all API requests. The call into `Request::new()` initiates logic to inspect the request and to create an `Actor` object. API call handling functions then use an `aa!` macro (_a_uthentication and _a_uthorisation) to restrict access via the `Actor::is_allowed()` function which in turn, if the feature is enabled, asks the Oso policy engine to decide if the request should be permitted or denied.
- `src/daemon/http/auth.rs` contains the HTTP endpoint handlers that are specific to auth related communication with clients such as `krillc` and the Lagosta web UI.
- UI based tests can be run with `cargo test --feature multi-user` (requires Docker and rights to use Docker as a non-root user, only tested on Linux). These produce a video in `target/ui/videos` which can be quite educational to watch. They run an instance of Krill listening on localhost:3000 and a headless instance of the Google Chrome browser inside a Docker container which connects to via the host network interface to Krill to serve the Krill UI and access the Krill API. The test logic and configuration is spread across `multi_user_xxx.rs` files and correspondingly named browser "scripts" in "multi_user_xxx.js" files and Krill configuration files in "multi_user_xxx.conf" files. For the OpenID Connect tests some configuration of users and rights metadata is hard-coded in the `openid_connect_mock.rs` module.
- `src/commons/actor.rs` defines the new `Actor` type.
- `src/daemon/auth/authorizer.rs` defines the border between Krill code and auth specific code.
- `src/daemon/auth/providers/` contains the individual AuthProvider implementations.

### Visualisation of the key components & relationships

multi-user-class-diagram-simplified

### Open TODO items

This list is incomplete. It will be expanded during the review process.

- [ ] Add/mrpove RustDoc comments.
- [ ] Reformat to 120 character max line length (most of the current code was written using an 80 character limit).


ximon18:
I’m going to toggle this review to the ready state to see if GH will run the Actions against it then.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -0,0 +1,1123 @@
...
+
+        Ok(logout_url)
+    }
+
+    fn try_refresh_token(&self, session: &ClientSession) -> Option<Auth> {
+        match &session.secrets.get(0) {
+            Some(refresh_token) => {
+                debug!("OpenID Connect: Refreshing token for user: \"{}\"", &session.id);
+                trace!("OpenID Connect: Submitting RFC-6749 section 6 Access Token Refresh request");
+                match self.conn.read() {
```
ximon18, line 304: **TODO:** Review this locking code, it feels clumsy. Is the Arc needed given that the provider is accessed through Request::state which is itself `Arc<RwLock<KrillServer>>` ?

Cargo.toml
```diff
@@ -18,44 +18,50 @@ build = "build.rs"
...
 hyper           = "^0.13"
 intervaltree    = "0.2.6"
-lazy_static     = "1.4.0"
+jmespatch       = { version = "^0.3", features = ["sync"], optional = true }
 libflate        = "^1.0"
 log             = "^0.4"
+openidconnect   = { version = "^1.0", optional = true }
+#openidconnect = { git = "https://github.com/ximon18/openidconnect-rs.git", branch = "krill-hack", optional = true }
 openssl         = { version = "^0.10", features = ["v110"] }
+oso             = { version = "^0.8", optional = true }
```
ximon18, line 39: Latest is now 0.9.0 but I'm not aware of any particular bug fixes or features that would be an urgent reason to upgrade.

src/daemon/auth/common/session.rs
```diff
@@ -0,0 +1,340 @@
...
+                        },
+                    );
+                }
+                Err(err) => warn!("Unable to cache decrypted session token: {}", err),
+            },
+            Err(err) => warn!("Unable to cache decrypted session token: {}", err),
+        }
+    }
+
+    pub fn encode(
```
ximon18, line 169: This type, `LoginSessionCache`, is a sort of user session store. That it decodes on the fly and the decoded sessions are short lived are implementation details. I wonder then whether with the right trait and naming we could use that trait as a facade behind which we could later swap the implementation with say an on-disk db or a distributed key-value store etc?

src/daemon/auth/providers/openid_connect/util.rs
```diff
@@ -0,0 +1,226 @@
...
+// to, for example, define a "role" member field of our custom additional claims
+// struct, serde_json would fail to deserialize it if the the field is not
+// present or not structured as expected. Using this approach we can inspect the
+// structure when we receive it from the provider.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct CustomerDefinedAdditionalClaims(serde_json::Value);
+impl AdditionalClaims for CustomerDefinedAdditionalClaims {}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct CustomerDefinedExtraTokenFields(serde_json::Value);
```
ximon18, line 38: The oauth2-rs and openidconnect-rs crates do this slightly differently, maybe their way would be better?
E.g. see [here](https://github.com/ramosbugs/oauth2-rs/blob/176fbdd4ae97868f20757690329fe05b71a23652/examples/google_devicecode.rs#L26) and [here](https://github.com/ramosbugs/openidconnect-rs/blob/f4a7abef527a22a905521a198a98b6f1b435c422/src/user_info.rs#L539). Rather than `(serde_json::Value)` they use `(HashMap<String, serde_json::Value)`.

ximon18:
I'm going to briefly toggle this PR to non-draft and push a commit to it to see if that triggers the e2e test or not.

timbru (approved):

Issue #371, opened 2021-01-12
Trigger creation of a code coverage report via GitHub Actions


ximon18:
E.g. use [cargo tarpaulin](https://crates.io/crates/cargo-tarpaulin) triggered after a successful `CI` build.

ximon18:
Linked pull request #372 creates a cargo tarpaulin code coverage HTML report and uploads it to GitHub as a workflow artifact. Code coverage is reported both for default features and for all features. In the image below you can see hte difference in the report for default features vs all features at the top level of the source tree:

![image](https://user-images.githubusercontent.com/3304436/104383485-31f71680-5530-11eb-8382-dcab426a9dfb.png)

These reports can be drilled into, e.g.:

![image](https://user-images.githubusercontent.com/3304436/104383557-4f2be500-5530-11eb-85e6-83c355af6e80.png)

ximon18:
One "issue" with the current implementation is that while it is a separate build job to the CI build job, it is still part of the CI workflow which means that:
- the CI workflow will fail if generating the code coverage report fails
- the CI workflow will take longer to complete than it does without code coverage

Perhaps it would be good to investigate if it can bs separate out to a separate workflow which is triggered after CI workflow success?

ximon18:
Closing as all new successful (and test) runs on the dev branch now cause a coverage report job to run.

PR #372, opened 2021-01-12
Extend the CI job with code coverage reports by cargo tarpaulin. (#371)

ximon18:


ximon18:
When reviewing see the comments on issue #371.

.github/workflows/ci.yml
```diff
@@ -35,28 +38,40 @@ jobs:
...
       with:
         rust-version: ${{ matrix.rust }}
-    - run: cargo build --verbose
-    - run: cargo test --verbose 2>&1
-  allfeatures:
-    name: allfeatures
+    - run: cargo build --verbose ${{ matrix.args }}
+    - run: cargo test --verbose ${{ matrix.args }} 2>&1
+
+  coverage:
```
ximon18, line 52: One issue with this approach is that while it uses a separate job to the main `build` CI job, it is still part of the CI workflow which means that:

- the CI workflow will fail if generating the code coverage report fails
- the CI workflow will take longer to complete than it does without code coverage

Perhaps it would be good to investigate if it can be separated out to a separate workflow which is triggered after CI workflow success?

.github/workflows/ci.yml
```diff
@@ -35,28 +38,40 @@ jobs:
...
     steps:
     - name: Checkout repository
       uses: actions/checkout@v1
     - name: Install Rust
       uses: hecrj/setup-rust-action@v1
       with:
         rust-version: ${{ matrix.rust }}
-    - run: cargo build --verbose --all-features
-    - run: cargo test --verbose --all-features 2>&1
+    - run: cargo install cargo-tarpaulin
```
ximon18, line 68: Perhaps use https://github.com/cypress-io/github-action here instead of compiling tarpauling via `cargo install`?

.github/workflows/ci.yml
```diff
@@ -35,28 +38,40 @@ jobs:
...
+    # use a long timeout with tarpaulin as UI tests have to docker pull the
+    # cypress.io image which can cause a test timeout with the default tarpaulin
+    # timeout of 1 minute.
```
ximon18, line 71: We should solve this by pre-pulling the docker image rather than increasing a timeout.

.github/workflows/ci.yml
ximon18: Done, but the new coverage workflow is not visible in the Actions tab.

I suspect that GH won’t create the new coverage workflow until the new YML file is merged to master. As such we have no way of knowing if the new coverage YML is correct or is invoked by GH when intended, not will coverage reports be generated... 

@timbru: Unless you object I will add the new coverage.yml workflow file directly to master to see if it gets triggered and behaves as expected.

ximon18: The new coverage workflow is now working. I have tested that it runs on the correct Git commit when triggered by any of the following:
- A successful CI workflow run for a push to master (also if the CI workflow is re-run). See [this example workflow run](https://github.com/NLnetLabs/krill/actions/runs/486345094)
- A successful CI workflow run for a push to a pull request branch (also if the CI workflow is re-run). See [this example workflow run](https://github.com/NLnetLabs/krill/actions/runs/486372881)

A couple of additional PROs and CONs of this approach occurred to me:
- PRO: the coverage report workflow job doesn't add noise to the "checks" report shown in the GitHub view of a PR.
- CON: if we later wanted to fail builds because of insufficient or regressing coverage this approach would not work. However, with the current tooling and implementation the coverage report takes so long to generate that I don't think we'll want to make normal CI builds wait for a coverage report, and we could in theory (with more work), if we wanted to, make the separate coverage workflow post a comment on a PR about coverage issues.

This branch also might have solved the flakey test issue reported in #377, at least I pushed a couple of tweaks for #377 to this branch (because #377 was preventing the CI or coverage jobs to fail which blocked my work on this issue) and I haven't encounted the failure since.

.github/workflows/ci.yml
ximon18: This isn't going to work as the action runs the tests (it doesn't just pull in the Cypress test runner Docker image) and currently the Cypress test runner is invoked by Rust integration test code (not by the GH workflow) which also does pre and post steps around the invocation of the Cypress runner (e.g. start and stop Krill, start and stop the mock OpenID Connect provider, inspect the Krill event history after the test, etc).

ximon18: I won't investigate this at the moment as the win isn't big enough to warrant the additional time to test the tarpaulin action and to work out where it writes the output report and to upload it correctly. Currently the workflow run for all features against the v0.8.1-bis branch takes 28 minutes of which only 3 minutes is spent installing cargo tarpaulin.

ximon18:
I think this is good enough for now. If interested click the "Show Resolved" links above for some issues I thought of and why I propose not to do anything more at this point about them. The only pain point in my opinion that remains is that the workflow run in the GitHub Actions tab is always called "Coverage" and doesn't indicate which branch or PR was tested, for that you have to dig into the run logs to see which code was checked out. Unfortunately I don't know how to improve this.

tests/ui/cypress_specs/multi_user_config_file_with_ta.js
```diff
@@ -211,6 +211,9 @@ describe('Config File Users with TA', () => {
             if (ts.o) {
               cy.contains('Success')
               cy.contains('Error').should('not.exist')
+              // wait for the parent registration to complete inside Krill and
```
ximon18, line 214: This change is actually for #377 but was made here because #377 was blocking testing for #371.

tests/ui/cypress_specs/multi_user_config_file_with_ta.js
```diff
@@ -238,13 +241,18 @@ describe('Config File Users with TA', () => {
 
   add_roa_test_settings.forEach(function (ts) {
     it('Add ROA for CA ' + ts.ca + ' as ' + ts.d + ' user should ' + (ts.o ? 'succeed' : 'fail'), () => {
+      cy.intercept('GET', '/api/v1/cas/' + ts.ca + '/routes/analysis/full').as('analyzeRoutes')
```
ximon18, line 244: This change is actually for #377 but was made here because #377 was blocking testing for #371.

tests/ui/cypress_specs/multi_user_config_file_with_ta.js
```diff
@@ -238,13 +241,18 @@ describe('Config File Users with TA', () => {
...
+
       cy.visit('/')
       cy.get('input[placeholder="Your username"]').type(ts.u)
       cy.get(':password').type(ts.p)
       cy.contains('Sign In').click()
       cy.contains(ts.u)
       cy.contains('Sign In').should('not.exist')
 
+      // wait for Lagosta to finish fetching the route analysis details
```
ximon18, line 253: This change is actually for #377 but was made here because #377 was blocking testing for #371.

timbru (approved):

ximon18:
There's still an issue with the 'Add ROA for ...' UI test:

```
Add ROA for CA ca_readonly as readonly user should fail:
     CypressError: Timed out retrying after 5000ms: `cy.wait()` timed out waiting `5000ms` for the 1st request to the route: `analyzeRoutes`. No request ever occurred.
```

However, that test already has issues in the 0.8.1-bis branch so I'm going to merge this anwyay and continue to investigate test issues as they are not directly related to adding supporrt for generating a test coverage report.

PR #376, opened 2021-01-13
Add SessionStatus::NeedsRefresh to status check

density215:
Since the try\_refresh\_token return None if the Session can't be extended, we can safely add the condition SessionStatus::NeedsRefresh here.

This is sort of the future-proof version of this conditional check (I realise that right now, the whole conditional that deals with the SessionStatus variant can be removed here (since there's only three variants). As such this PR is a reminder, not a suggestion for the final code.

### DIFF
--- a/src/daemon/auth/providers/openid_connect/provider.rs
+++ b/src/daemon/auth/providers/openid_connect/provider.rs
@@ -591,9 +591,12 @@ impl AuthProvider for OpenIDConnectAuthProvider {
                 let status = session.status();
 
                 let mut new_auth = None;
+                if log_enabled!(log::Level::Trace) {
+                    trace!("Status: {:?}", status);
+                }
                 if status != SessionStatus::Active {
                     new_auth = self.try_refresh_token(&session);
-                    if new_auth.is_none() && status == SessionStatus::Expired {
+                    if new_auth.is_none() && (status == SessionStatus::Expired || status == SessionStatus::NeedsRefresh) {
                         return Err(Error::ApiInvalidCredentials(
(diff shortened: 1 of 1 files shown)


src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -591,9 +591,12 @@ impl AuthProvider for OpenIDConnectAuthProvider {
                 let status = session.status();
 
                 let mut new_auth = None;
+                if log_enabled!(log::Level::Trace) {
+                    trace!("Status: {:?}", status);
+                }
                 if status != SessionStatus::Active {
                     new_auth = self.try_refresh_token(&session);
-                    if new_auth.is_none() && status == SessionStatus::Expired {
+                    if new_auth.is_none() && (status == SessionStatus::Expired || status == SessionStatus::NeedsRefresh) {
```
ximon18, line 599: Hi @density215,

If the session is not active that does not mean it is expired, rather that it is either expired or far enough into its token lifetime that it is worth trying to refresh it (because we try to refresh it before it expires). If refresh fails but the session is not expired that means that the session is in fact still okay and the user need not be logged out yet, perhaps a future attempt to refresh the token will succeed where this attempt failed.

It does not seem correct to me to error here if the status is needs refresh. Do you concur? Perhaps this code would be better if it used an explicit match block showing that `try_refresh_token()` is executed for both the needs refresh and expired cases?

Ximon

ximon18:
@density215: let’s talk about this today, I’d like to understand your points about future proofing and the redundant conditional.

density215:
Superseded by #396

PR #394, opened 2021-01-15
Use an official Alpine docker image release rather than Alpine Edge.

### LINKED ISSUE #387: Upgrade Docker image to Alpine 3.13 base
Currently the `v0.8.1-bis` branch uses an Edge version of the Alpine base Docker image. This is because the latest release at the time, 3.12.x, only had support for Rust versions that were too old to compile the `v0.8.1-bis` code.

[Alpine 3.13 was released yesterday](https://alpinelinux.org/posts/Alpine-3.13.0-released.html) and is now available via Docker Hub. This release includes the Rust compiler version that was previously only available via Alpine Edge. We can therefore upgrade to 3.13.0, unless we instead want to investigate reducing the minimum version of Rust required.

```
$ docker run -it alpine:3.12.3 sh -c 'apk -q add rust ; rustc -V'
rustc 1.44.0

$ docker run -it alpine:3.13.0 sh -c 'apk -q add rust ; rustc -V'
rustc 1.47.0
```

The issue encountered with the older compiler was:
```
   Compiling reqwest v0.10.8
error[E0658]: use of unstable library feature 'str_strip': newly added
   --> src/commons/api/roas.rs:380:49
    |
380 |             } else if let Some(stripped) = line.strip_prefix("A:") {
    |                                                 ^^^^^^^^^^^^
    |
    = note: see issue #67302 <https://github.com/rust-lang/rust/issues/67302> for more information
error[E0658]: use of unstable library feature 'str_strip': newly added
   --> src/commons/api/roas.rs:383:49
    |
383 |             } else if let Some(stripped) = line.strip_prefix("R:") {
    |                                                 ^^^^^^^^^^^^
    |
    = note: see issue #67302 <https://github.com/rust-lang/rust/issues/67302> for more information
error: aborting due to 2 previous errors
For more information about this error, try `rustc --explain E0658`.
error: could not compile `krill`.
To learn more, run the command again with --verbose.
```

The `str_strip` feature was stabilized Rust 1.45.0.

The `strip_prefix` function is used in Krill as follows:

```
$ git grep strip_prefix
src/commons/api/roas.rs:            } else if let Some(stripped) = line.strip_prefix("A:") {
src/commons/api/roas.rs:            } else if let Some(stripped) = line.strip_prefix("R:") {
src/commons/util/file.rs:    let rel = path.strip_prefix(base_path).map_err(|_| Error::PathOutsideBasePath)?;
```

ximon18:
Resolves #387.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,7 @@
 # Make the base image configurable so that the E2E test can use a base image
 # with a prepopulated Cargo build cache to accelerate the build process.
 # Use Ubuntu 16.04 because this is what the Travis CI Krill build uses.
-#
-# Use Alpine edge (speific dated build 20201218) to get Rust >= 1.45.0 as needed
-# to build Krill, as latest Alpine release contains Rust 1.44.0.
-ARG BASE_IMG=alpine:20201218
+ARG BASE_IMG=alpine:3.13
 
 #
 # -- stage 1: build krill and krillc
@@ -17,7 +14,7 @@ RUN apk add rust cargo openssl-dev
 WORKDIR /tmp/krill
 COPY . .
 
-RUN cargo build --target x86_64-alpine-linux-musl --release
+RUN cargo build --target x86_64-alpine-linux-musl --release --features multi-user
 
 #
 # -- stage 2: create an image containing just the binaries, configs &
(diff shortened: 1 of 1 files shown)


ximon18:
As `build.rs` in the v0.8.1-bis branch already requires Rust 1.45.0 the decision has already been made to move to a newer Rust version, so by moving the Docker image from Alpine Edge to an official Alpine official release we're only stabilizing the Docker image base, we're not changing anything about how Krill uses Rust.

density215 (approved):

Issue #375, opened 2021-01-13
Client side login session is still considered valid by Krill if the token is removed at the OpenID Connect provider


density215:
Note: This relates to the `multi-user` feature in the `v0.8.1-bis` branch.

If the session is removed on the AuthProvider the client-side still can successfully perform API requests that require authentication, even though krill receives a 400 from the AuthProvider:

```
OpenID Connect: Refreshing token for user: "jasper@plainspace.com"
OpenID Connect: Submitting RFC-6749 section 6 Access Token Refresh request
[...]

Issue #87, opened 2019-09-26
Explicitly trigger background jobs for testing.


timbru:
This should provide a quick fix for the failing builds resulting from concurrency errors. The real fix for this is planned in #64.

PR #396, opened 2021-01-15
authenticate()/try_refresh_token() rfc-6749 style errors

density215:
This supersedes the PR for #375.

It changes the return type of try\_refresh\_token (an Option) to a Result with the `CoreErrorResponseType` as its Error type. This type maps to the rfc 6749 5.2 Error codes for Access Token Request Errors.

The caller, provider.authenticate() can then decide based on the Error variant whether it wants to retry a refresh or report back to the user (this is a TODO, right now it always returns a HTTP 401).

The Return Error also has been changed from 403 Forbidden to 401 Unauthorized, since the most obvious action when reported back to the end-user would be to login again. Also all CoreErrorResponseTypes indicate that authentication is lost  (due to changes in Scopes, Realms, Users and/or Clients on the OIDC Provider side).


src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -367,9 +391,14 @@
                     "OpenID Connect: Could not extend login session for user '{}': no refresh token available",
                     &session.id
                 );
+                return Err(CoreErrorResponseType::Extension(
+                    "OpenID Connect: Unknwon Error with existing token".to_string(),
+                ));
             }
         }
-        None
+        return Err(CoreErrorResponseType::Extension(
```
ximon18, line 399: clippy::needless_return warning here

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -587,20 +617,41 @@
...
+                                return Err(Error::ApiAuthError("There is a unknown Authentication Problem. You may try to Login Again.".to_string()))
+                            },
+                            // If changes are made to the roles of the user, the client or the scope on the side of the OpenID Connect Provider,
+                            // the token refresh may get one of these errors.
+                            CoreErrorResponseType::UnauthorizedClient | CoreErrorResponseType::UnsupportedGrantType | CoreErrorResponseType::InvalidScope => {
+                                return Err(Error::ApiAuthError("The Authorization was revoked for this user, client or action. Please ask you Administrator".to_string()))
+                            },
+                            // The Extension Type Errors are used by the try_refresh_token method to signal generic problems with either the
+                            // current token, or the freshly received one.
+                            CoreErrorResponseType::Extension(e) => {
```
ximon18, line 647: clippy unused_variables warning here

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -338,10 +341,24 @@
...
-                                            &session.id, err
-                                        );
+                                        match &err {
+                                            // this is where the RFC-6749 5.2 Error Response is received and
+                                            // return to the caller. It's the responsibility of the caller
+                                            // to decide whether to retry or report back to the user.
+                                            openidconnect::RequestTokenError::ServerResponse(r) => {
+                                                warn!(
+                                                    "OpenID Connect: Could not extend login session for user '{}': {}",
+                                                    &session.id, "Connection to OpenID Connect provider not yet established"
```
ximon18, line 351: This warning message cannot be correct as if we hadn't already established a connection to the provider we would not be handling a provider response.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -338,10 +341,24 @@
...
                                     }
                                     Err(err) => {
-                                        warn!(
-                                            "OpenID Connect: Could not extend login session for user '{}': {}",
-                                            &session.id, err
-                                        );
+                                        match &err {
+                                            // this is where the RFC-6749 5.2 Error Response is received and
+                                            // return to the caller. It's the responsibility of the caller
+                                            // to decide whether to retry or report back to the user.
```
ximon18, line 347: Could this be moved up into the Err() match above thus avoiding the need for an extra match?

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -338,10 +341,24 @@
...
+                                            openidconnect::RequestTokenError::ServerResponse(r) => {
+                                                warn!(
+                                                    "OpenID Connect: Could not extend login session for user '{}': {}",
+                                                    &session.id, "Connection to OpenID Connect provider not yet established"
+                                                );
+                                                return Err(r.error().clone());
```
ximon18, line 353: The call to r.error() discards the error description string - is it really of no value?

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -338,10 +341,24 @@
...
+                                            }
+                                            _ => {
+                                                return Err(CoreErrorResponseType::Extension(
```
ximon18, line 356: None of the information about the actual error is returned here, is there nothing of value to pass back about the error?

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -587,20 +617,41 @@
...
+                if status == SessionStatus::Active {
+                    return Ok(Some(ActorDef::user(session.id, session.attributes, None)));
                 }
 
-                Ok(Some(ActorDef::user(session.id, session.attributes, new_auth)))
+                let new_auth = match self.try_refresh_token(&session) {
+                    Ok(auth) => auth,
+                    Err(e) => {
+                        warn!("RFC-6749 5.2 Error Response returned: {:?}", e);
+                        match e {
```
ximon18, line 631: These match handlers discard the original error details, is there nothing useful from the original error to return to the caller?

ximon18:
Please run `cargo fmt`.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -587,20 +617,41 @@
...
+                // Token found in cache and active; all good, do an early return
+                if status == SessionStatus::Active {
+                    return Ok(Some(ActorDef::user(session.id, session.attributes, None)));
                 }
 
-                Ok(Some(ActorDef::user(session.id, session.attributes, new_auth)))
+                let new_auth = match self.try_refresh_token(&session) {
+                    Ok(auth) => auth,
+                    Err(e) => {
+                        warn!("RFC-6749 5.2 Error Response returned: {:?}", e);
```
ximon18, line 630: Is RFC-6479 5.2 the only way that try_refresh_token() can fail?

src/daemon/auth/providers/openid_connect/provider.rs
density215: This is what I am still working on, there's other variants in here that need to be handled, I think. If this turns out to be only usefull variant, I'll collapse this one into its parent.

src/daemon/auth/providers/openid_connect/provider.rs
density215: I don't think so, the description is a summary of the rfc06749 5.2 text, it contains no details about this context

src/daemon/auth/providers/openid_connect/provider.rs
density215: See my comment above, I need to figure out more details about the other variants here

src/daemon/auth/providers/openid_connect/provider.rs
density215: Probably yes, I assumed this would be embedded in a more debug level message about the context

src/daemon/auth/providers/openid_connect/provider.rs
density215: According to the docs of the openidconnect crate, yes. I still have to figure out where the 4.2.2.1 Error Responses turn up

src/daemon/auth/providers/openid_connect/provider.rs
density215: No, sure. Any suggestions for a better one?

src/daemon/auth/providers/openid_connect/provider.rs
ximon18: I was meaning that `try_refresh_token()` is an internal fn, not part of the openidconnect crate, and so may have its own logic and cause other errors than those produced by the openidconnect crate. Since you posted this PR I haven't had a chance to look at the actual code of `fn try_refresh_token()`.

density215:
A more elaborate description of what changed/is new:

*   Introduces a `commons::error::ApiAuthError` enum, as well as the corresponding variants in the `commons::error:Error` enum. They can be converted to and from by implementing the \`From\` and  `To` traits.
*   Changes the methods that took an error string as an argument/return value to taking an `ApiAuthError` for both.
*   Changes the mapping from `ApiAauthErrors` to `HttpResponse` errors.
*   Implements the rfc-6749 5.2 errors in `OpenIDConnectAuthProvider.try_refresh_token()`, as well as some more generic errors (problems with tokens and network errors).
*   Implements the rfc-6749 5.2 errors in the `OpenIDConnectAuthProvider.authenticate()`. Note that additional errors that do not fit the 5.2 section are returned using `CoreErrorReponseType::Extension()`.
*   Adds cypress tests for situations where login is successful but a rfc-6749 5.2 style error gets thrown on trying to refresh the token.

These are the mappings from OIDC Provider errors returned to krill and the corresponding errors krill throws in its API. Note that the last row is out of scope for this PR / repo and should be implemented in Lagosta.

| **OIDC Provider error** | **Krill Error label** | **Krill HTTP Code** | **Lagosta Action Suggestion** |
| --- | --- | --- | --- |
| invalid\_scope, unsupported\_grant\_type, unauthorized\_client | api-insufficient-rights | 403 FORBIDDEN | Retry Login / Contact Admin |
| invalid\_grant | api-invalid-credentials | 401 UNAUTHORIZED | Retry Login |
| invalid\_client, invalid\_request | api-auth-permanent-error | 401 UNAUTHORIZED | Contact Admin |
| 500, 503, Network Failure | api-auth-transient-error | 401 UNAUTHORIZED | Retry Action |

tests/ui/cypress/specs/multi_user_openid_connect.js
```diff
@@ -0,0 +1,289 @@
...
+
+    // Try to create a CA, by typing in the input, clicking the 'Create CA' button
+    // and then clicking 'Ok'. This should fail, since the token can't be refreshed.
+    cy.intercept('POST', '/api/v1/cas').as('createCA')
+    cy.get('main input').type('some-handle-name')
+    cy.get('main button').click()
+    cy.get('.el-message-box__btns button:nth-child(2)').click()
+    cy.wait('@createCA').its('response.statusCode').should('eq', 401)
+  })
+  ;[...create_ca_settings_401, ...create_ca_settings_403].forEach((t) =>
```
density215, line 216: Interesting, prettier (the JS formatter) really, really wants this if you disable semis.

tests/ui/cypress/specs/multi_user_openid_connect.js
```diff
@@ -0,0 +1,289 @@
...
+    cy.intercept('GET', '/api/v1/authorized').as('isAuthorized')
+    cy.contains('Sign In').click()
+
+    cy.wait('@postLoginIndexFetch').its('response.statusCode').should('eq', 200)
+    cy.wait('@isAuthorized').its('response.statusCode').should('eq', 200)
+    cy.url().should('include', Cypress.config('baseUrl'))
+    cy.contains('Sign In').should('not.exist')
+    cy.get('#userinfo').click()
+    cy.get('#userinfo_table').contains(shorttoken.u)
+    cy.contains(shorttoken.u)
```
ximon18, line 145: In `multi_user_config_file_with_ta.js` this is done differently:

```
// create a CA
cy.contains('CA Handle')
cy.get('form input[type="text"]').type(ts.ca)
cy.contains('Create CA').click()
cy.contains('OK').click()
```

The intercept you have added here would be good to add in the `with_ta` JS script as a future improvement (let's not mix changes to that test in with this PR, I just want to say I like your use of the intercept here).

However, I find the locators used in the `with_ta` test JS script easier to understand (click "Create CA" instead of click "main button" for example) so perhaps it would be good to use the locator style from the `with_ta` JS script here.

tests/ui/cypress/specs/multi_user_openid_connect.js
```diff
@@ -0,0 +1,289 @@
...
+
+    // the token has a lifetime of 5 second and no refresh token
+    // wait 6 seconds...
+    // note: a shorter token with a 1 second lifetime doesn't work in the GitHub
+    // Action runner environment because the token has sometimes already expired
+    // by the time Krill verifies it!
+    cy.wait(6000)
+
+    // Try to create a CA, by typing in the input, clicking the 'Create CA' button
+    // and then clicking 'Ok'. This should fail, since the token can't be refereshed.
```
ximon18, line 156: Minor improvement suggestion: move `[valid refresh]` to the end of the test name otherwise when viewing the list of tests all of the other test names start with an English description but these tests do not which makes the list of test names harder to read (for me at least).

tests/ui/cypress/specs/multi_user_openid_connect.js
```diff
@@ -0,0 +1,289 @@
...
+    cy.url().should('not.include', Cypress.config('baseUrl'))
+    cy.contains('Mock OpenID Connect login form')
+    cy.get('input[name="username"]').clear().type(shorttoken.u)
+
+    cy.intercept('GET', '/index.html').as('postLoginIndexFetch')
+    cy.intercept('GET', '/api/v1/authorized').as('isAuthorized')
+    cy.contains('Sign In').click()
+
+    cy.wait('@postLoginIndexFetch').its('response.statusCode').should('eq', 200)
+    cy.wait('@isAuthorized').its('response.statusCode').should('eq', 200)
```
ximon18, line 193: Missing line break here to separate these tests out from each other?

tests/ui/cypress/specs/multi_user_openid_connect.js
```diff
@@ -0,0 +1,289 @@
...
+    cy.get('input[name="username"]')
+  })
+
+  it('Login with short-lived non-refreshable token and try to refresh page', () => {
+    cy.intercept('GET', '/api/v1/authorized').as('isAuthorized')
+    cy.visit('/')
+
+    cy.wait('@isAuthorized').its('response.statusCode').should('eq', 403)
+    cy.url().should('not.include', Cypress.config('baseUrl'))
+    cy.contains('Mock OpenID Connect login form')
```
ximon18, line 132: You added 'try to create a CA" to the name of the next test but also extended this test to try to create a CA. Actually, I must be missing something, how does this test differ from the next one? What is meant by `[valid refresh]` ? I don't see any difference in the logic of this test and the next except that this test also checks that the user is shown the OIDC login page at the end of the test?

tests/ui/cypress/specs/multi_user_openid_connect.js
```diff
@@ -0,0 +1,289 @@
...
+    cy.get('input[name="username"]').clear().type(shorttoken.u)
+
+    cy.intercept('GET', '/index.html').as('postLoginIndexFetch')
+    cy.intercept('GET', '/api/v1/authorized').as('isAuthorized')
+    cy.contains('Sign In').click()
+
+    cy.wait('@postLoginIndexFetch').its('response.statusCode').should('eq', 200)
+    cy.wait('@isAuthorized').its('response.statusCode').should('eq', 200)
+    cy.url().should('include', Cypress.config('baseUrl'))
```
ximon18, line 194: lived with a 'd'.

tests/ui/cypress/specs/multi_user_openid_connect.js
```diff
@@ -0,0 +1,289 @@
...
+
+      cy.intercept('GET', '/index.html').as('postLoginIndexFetch')
+      cy.intercept('GET', '/api/v1/authorized').as('isAuthorized')
+      cy.contains('Sign In').click()
+
+      cy.wait('@postLoginIndexFetch').its('response.statusCode').should('eq', 200)
+      cy.wait('@isAuthorized').its('response.statusCode').should('eq', 200)
+      cy.url().should('include', Cypress.config('baseUrl'))
+      cy.contains('Sign In').should('not.exist')
+      cy.get('#userinfo').click()
```
ximon18, line 234: Minor: "try" -> "try to"

tests/ui/cypress/specs/multi_user_openid_connect.js
density215: Do you want all the bracketed usernames at the end? I put them in front because the username is what makes them discernable.

tests/ui/cypress/specs/multi_user_openid_connect.js
```diff
@@ -0,0 +1,289 @@
...
+    // by the time Krill verifies it!
+    cy.wait(6000)
+
+    // Try to create a CA, by typing in the input, clicking the 'Create CA' button
+    // and then clicking 'Ok'. This should fail, since the token can't be refereshed.
+    cy.intercept('POST', '/api/v1/cas').as('createCA')
+    cy.get('main input').type('some-handle-name')
+    cy.get('main button').click()
+    cy.get('.el-message-box__btns button:nth-child(2)').click()
+    cy.wait('@createCA').its('response.statusCode').should('eq', 401)
```
ximon18, line 161: The appearance of the error message could be improved. That's not something to fix in this PR but I noticed it when seeing what the end user experience looks like with Cypress running the test interactively:
image
At the very least perhaps we should make the text left-aligned so that it doesn't space out so much.

tests/ui/cypress/specs/multi_user_openid_connect.js
density215: See my remark about the semis above, it's prettier doing the weird thing with the line-break and the semicolon. Made a workaround for that

tests/ui/cypress/specs/multi_user_openid_connect.js
ximon18: If you don't immediately think it would be better the other way I'd say leave it as you have a reason for the naming strategy you chose and the most important thing is the test itself, I can live with the name.

tests/ui/cypress/specs/multi_user_openid_connect.js
ximon18: Ah okay, I didn't understand the comment about the prettifier :-)

tests/ui/cypress/specs/multi_user_openid_connect.js
density215: I think Lagosta should be modified to handle the various errors more gracefully, like always supplying a suggested action to the user. Ideally it should not dump the API error it receives on the user directly (in the console is fine I guess)

ximon18: I think indeed a review of the current error handling and presentation strategy could yiled improvement, but I don't know if we have time for that before we release this, so I was thinking of a minor improvement for now.

tests/ui/cypress/specs/multi_user_openid_connect.js
density215: You're right, I copied and extended the first one, but then left the original one there. There's a slight difference though, the first one (starting 126 'Login with short-lived non-refreshable token and try to refresh page') tries to refresh the page, a legit action that will receive a redirect back to login, whereas the second one (starting 179 '\[valid refresh\] Login with short-lived non-refreshable token and try to create a CA') does not refresh and therefore receives an API error and will not redirect to login.

The \[valid refresh\] is wrong though, I'll remove it, should be \[invalid refresh\].

tests/ui/cypress/specs/multi_user_openid_connect.js
density215: Maybe instead of `[${ts.u}] Login with short-lived non-refreshable token and try to create a CA` do \[${ts.u}\] `Login as ${ts.u} and try to create a CA` ? It's more concise and better for humans?

tests/ui/cypress/specs/multi_user_openid_connect.js
density215: Fair enough, I'll change it to do that. I wish though that Lagosta would have meaningful id attributes on the html elements involved. Selecting based on textContent is still a bit brittle.

tests/ui/cypress/specs/multi_user_openid_connect.js
ximon18: To be clear, I don't think this PR should be blocked by these issues, the primary value of this PR is handling OpenID Connect provider errors properly, and it would be unfair to say that this PR introduces the issues with the way errors are communicated to the Lagosta web UI and presented to the user.

tests/ui/cypress/specs/multi_user_openid_connect.js
ximon18: We can add them, I have done so in some cases while working on the multi-user work. A related approach that I stumbled upon yesterday is to add `cy-name` attributes.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -587,20 +616,61 @@ impl AuthProvider for OpenIDConnectAuthProvider {
...
+                            | CoreErrorResponseType::UnsupportedGrantType
+                            | CoreErrorResponseType::InvalidScope => {
+                                warn!("OpenID Connect:  rfc-6749 5.2 {:?}", err);
+                                return Err(Error::ApiInsufficientRights(
+                                    "The Authorization was revoked for this user, client or action.".to_string(),
+                                ));
+                            }
+                            // The Extension Type Errors are used by the try_refresh_token method to signal generic problems with either the
+                            // current token, or the freshly received one.
+                            CoreErrorResponseType::Extension(err) => {
```
ximon18, line 665: According to an [Errata for RFC 6749](https://www.rfc-editor.org/errata/eid4745) two error codes are missing from section 5.2 of the RFC: `server_error` and `temporarily_unavailable`. However I see that the underlying `openidconnect` Rust create enum `BasicErrorResponseType` doesn't define these either, so I guess these will be presented to our client as `::Extension(str)`. It might be worth also noting that here.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -587,20 +616,61 @@ impl AuthProvider for OpenIDConnectAuthProvider {
...
+                        );
+                        auth
+                    }
+                    Err(err) => {
+                        trace!("OpenID Connect: rfc-6749 5.2 Error Response returned...");
+                        debug!(
+                            "OpenID Connect: Refreshing the token for user '{}' failed: {}",
+                            &session.id, &err
+                        );
+                        match err {
```
ximon18, line 640: Minor point: I would prefer not to use initial capitalization for terms such as Session and Authentication Problem within these error strings. Note: these messages will only appear in the Krill log, the actual messages presented to end users are controlled by `en.json` etc in the Lagosta web UI code (assuming that all the used error labels have assigned values in the Lagosta translation files).

src/commons/error.rs
```diff
@@ -551,7 +600,9 @@ impl Error {
 
             Error::ApiInvalidCredentials(e) => ErrorResponse::new("api-invalid-credentials", &self).with_cause(e),
 
-            Error::ApiAuthError(e) => ErrorResponse::new("api-auth-error", &self).with_cause(e),
+            Error::ApiAuthPermanentError(e) => ErrorResponse::new("api-auth-error", &self).with_cause(e),
```
ximon18, line 603: These labels need adding to the Lagosta `en.json` translation file. Without them the hard-coded English message strings in the Krill code will be shown to the end user. I don't see a corresponding branch at https://github.com/NLnetLabs/lagosta/branches, we should create one for this work.

src/commons/actor.rs
```diff
@@ -207,15 +208,15 @@ impl Actor {
         A: ToPolar + Display + Clone,
         R: ToPolar + Display + Clone,
     {
-        if let Some(error_msg) = &self.auth_error {
+        if let Some(api_error) = &self.auth_error {
             trace!(
-                "Unable to check access: actor={}, action={}, resource={}: {}",
+                "Unable to check access(#1): actor={}, action={}, resource={}: {}",
```
ximon18, line 213: Rather than #1 and #2 we should change this message as it is not about being unable to check access, it's that the process of authenticating the user failed and so is not permitted access. So this should be "Access denied: ..."

src/commons/error.rs
```diff
@@ -281,7 +317,8 @@ impl fmt::Display for Error {
             Error::PostTooBig => write!(f, "POST body exceeds configured limit"),
             Error::PostCannotRead => write!(f, "POST body cannot be read"),
             Error::ApiInvalidCredentials(e) => write!(f, "Invalid credentials: {}", e),
-            Error::ApiAuthError(e) => write!(f, "Authentication error: {}", e),
+            Error::ApiAuthPermanentError(e) => write!(f, "Authentication error: {}", e),
+            Error::ApiAuthTransientError(e) => write!(f, "Transient Authentication error: {}", e),
```
ximon18, line 321: Lowercase authentication please.

src/daemon/auth/authorizer.rs
```diff
@@ -137,7 +138,10 @@ impl Authorizer {
             Ok(None) => self.actor_from_def(ACTOR_DEF_ANON),
 
             // error during authentication
-            Err(err) => self.actor_from_def(ACTOR_DEF_ANON.with_auth_error(err.to_string())),
+            Err(err) => {
+                // reveives a commons::error::Error, but we need an ApiAuthError
+                self.actor_from_def(ACTOR_DEF_ANON.with_auth_error(ApiAuthError::from(err)))
```
ximon18, line 143: What about using into() here?

And while you could fix "reveives" I suggest instead just dropping the comment entirely and make the block consistent with the others.

So something like this:

```
Err(err) => self.actor_from_def(ACTOR_DEF_ANON.with_auth_error(err.into()))
```

tests/ui/openid_connect_mock.rs
```diff
@@ -191,6 +233,117 @@ fn run_mock_openid_connect_server() {
...
+                role: Some("readonly"),
+                exc_cas: Some("ta,testbed"),
+                token_secs: Some(5),
+                refresh: true,
+                on_refresh_token_failure: Some(RefreshTokenFailureMode::Error500Response),
+                ..Default::default()
+            },
+        );
+        known_users.insert(
+            "user-with-503-on-refresh",
```
ximon18, line 336: Perhaps you didn't mean to leave these as "warnings", they seem to be diagnostic in nature?

tests/ui/openid_connect_mock.rs
```diff
@@ -530,11 +683,15 @@ fn run_mock_openid_connect_server() {
 
             // we skip over verifying the Authorization HTTP header but perhaps
```
ximon18, line 684: Ditto.

tests/ui/openid_connect_mock.rs
```diff
@@ -596,10 +753,53 @@ fn run_mock_openid_connect_server() {
...
+                                        .respond(
+                                            Response::empty(StatusCode(400))
+                                                .with_header(
+                                                    Header::from_str("Content-Type: application/json").unwrap(),
+                                                )
+                                                .with_data(format!("{{\"error\":\"{}\"}}", err_mode).as_bytes(), None),
+                                        )
+                                        .map_err(|err| err.into());
+                                }
+                            }
```
ximon18, line 802: You could just write this as b"..." since it's a constant string?
Or use the serde_json json! macro to ensure you are producing correct JSON.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -587,20 +616,61 @@ impl AuthProvider for OpenIDConnectAuthProvider {
...
+                let new_auth = match self.try_refresh_token(&session) {
+                    Ok(auth) => {
+                        trace!(
+                            "OpenID Connect: Succesfully refreshed token for user \"{}\"",
+                            &session.id
+                        );
+                        auth
+                    }
+                    Err(err) => {
+                        trace!("OpenID Connect: rfc-6749 5.2 Error Response returned...");
```
ximon18, line 635: I'm curious about the "don't shout rfc" commit you made that lowercased "RFC" in these log statements.

This change of case makes these log statements inconsistent with other log statements in this file that use uppercase for RFC. If you look at https://www.ietf.org/standards/rfcs/ you'll see that RFC is rendered in capital letters, in English at least acronyms are always in all capitals, and Googling a bit I found [this](https://tools.ietf.org/html/draft-carpenter-rfc-citation-recs-01#section-5.2) which also states that RFCs should be referenced in uppercase:

> The correct typography includes a blank space.  Both "RFC2026" and "RFC-2026" are incorrect as a descriptive term.

Although it also states that using a dash is incorrect so in that case the existing log comments are also incorrect...

density215: I find a mixture of lowercase and uppercase harder to read. Not a big deal, so if the standard is RFC, RFC it is

src/daemon/auth/providers/openid_connect/provider.rs
density215: commit cf3fe15 deals with it in both try\_refresh\_token() and authenticate()

src/daemon/auth/authorizer.rs
density215: Crossed my mind, let's see if I can make that work. 031f5ce handles it. Note that I had to pull in the `crate::commons::` and `crate::daemons::` stuff into focus for all features. Either that or have an explicit `crate::commons::error::Error` as argument type in `with_auth_error`

src/commons/actor.rs
density215: How about `Authentication denied` since actual access also needs a policy?

src/commons/actor.rs
```diff
@@ -103,8 +101,8 @@ impl ActorDef {
     }
 
     // store an error string instead of an Error because Error cannot be cloned.
```
ximon18, line 103: This comment should be removed now as it is obsoleted by this PR.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -357,49 +363,84 @@ impl OpenIDConnectAuthProvider {
...
-                                                    &session.id, err
-                                                );
+                                                return Err(CoreErrorResponseType::Extension(format!(
+                                                    "Error while encoding the refreshed token {}",
+                                                    err
+                                                )));
                                             }
                                         }
                                     }
                                     Err(err) => {
```
ximon18, line 378: I think the code here is the same for any request to the OpenID Connect provider as I suspect they all return error responses of the same structure and with the same error categories. E.g. whether it be the call to exchange the temporary token for an access and id token, or the call to refresh a token, or the call to revoke a token (see PR #410). As such I wonder if this logic shouldn't be moved into something like `logging_http_client` in `util.rs` which is used for all calls to the provider made via the OpenID Connect crate?

density215: Maybe, yes, but not sure. The way I read it is that the Errors in Section 4.2.2 are only valid as responses from the redirect URL from the initial Authorization Request (4.2.1). Subsequential access token requests can have Error responses from Section 5.2. These two error sections are subtly different.

tests/ui/openid_connect_mock.rs
density215: 5e96304 converts it to using json\_serde

tests/ui/openid_connect_mock.rs
density215: afa3581 And eab7197. Using `dev_logger` crate to log to stdout (for now), warn, info and trace messages. Hope it's ok to pull in another dep, if not it can be incorporated into the mock server code itself.

src/daemon/auth/providers/openid_connect/provider.rs
ximon18: Discussed with @density215. This might be the way to go but is a wider change that can be investigated as a separate PR.

tests/ui/openid_connect_mock.rs
ximon18: My initial thought is why not use the existing Fern logger dependency?

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -357,49 +363,84 @@ impl OpenIDConnectAuthProvider {
...
+                                            // Note that [Errata for RFC 6749](https://www.rfc-editor.org/errata/eid4745)
+                                            // defines two additional error responses, `server_error` and
+                                            // `temporarily_unavailable`, that don't have variant counterparts
+                                            // in the openid-connect crate. These two error messages will
+                                            // therefore **not** end up in the `ServerReponse` variant.
+                                            openidconnect::RequestTokenError::ServerResponse(r) => {
+                                                return Err(r.error().clone());
+                                            }
+                                            openidconnect::RequestTokenError::Request(r) => {
+                                                return Err(CoreErrorResponseType::Extension(format!(
```
ximon18, line 393: One thing I'm wondering about here is error chaining. In effect you are chaining two errors together by concatenating their error messages. However, the oauth2 crate from which RequestTokenError comes uses the failure crate and the [docs for that crate include](https://docs.rs/failure/0.1.8/failure/struct.Causes.html) this:

> Causes | A iterator over the causes of a Fail

Maybe the current approach is fine for now but I wanted to raise this in case it's really worth looking at. I have no idea.

tests/ui/openid_connect_mock.rs
density215: Overlooked fern, will use that.

src/daemon/auth/providers/openid_connect/provider.rs
density215: The ServerResponse(T) variant's T type has bounds `ErrorResponse + 'static`, so no `Fail`. It doesn't have a cause() method. The ErrorResponse which wraps the StandardErrorResponse trait, which feature the error\_description() method. According to the docs "`... is used to assist the client developer in understanding the error that occurred."`

So I don't think that should be propagated up the error chain, since it doesn't give the end user anything. We could add a `trace!` statement for it.

The other variants here are all non-RFC 6749 error messages, more generic messages, like known unknowns and unknown unknowns.

So I think the current approach is good enough for now

.gitignore
```diff
@@ -7,3 +7,4 @@ krill.log
 target/
 tmp
 work
+examples/openid_connect_mock.rs
```
ximon18, line 10: Did you mean to push this change?

src/commons/actor.rs
```diff
@@ -207,15 +205,15 @@ impl Actor {
         A: ToPolar + Display + Clone,
         R: ToPolar + Display + Clone,
     {
-        if let Some(error_msg) = &self.auth_error {
+        if let Some(api_error) = &self.auth_error {
             trace!(
-                "Unable to check access: actor={}, action={}, resource={}: {}",
+                "Authentication denied: actor={}, action={}, resource={}: {}",
```
ximon18, line 210: Looking at this again I think both the original log message and the new log message are incorrect.

In the original log message saying that we were unable to check access leaves it unclear (from that log statement alone) whether access is being granted or denied.

With the new log message it is also not clear, and it's also not true to say that authentication is/was denied. Instead what's happened is that the user was able to authenticate against the OpenID Connect 3rd party system (i.e. login to company X) but we were unable to complete the login to Krill itself perhaps due to missing information in the backend provider metadata about the user or a problem with connection to the backend after the user was authenticated by the 3rd party system but before Krill finished its part of the login process.

Any return from this function leads to one of two possible outcomes for the end user: access denied or access granted. I suppose you could also try to handle a 3rd scenario whereby a retry of a failure could be done, but you'd have to understand the root cause for why you didn't complete the login process properly which can't be known from here.

A nice simplification of this function would be to have only one log statement that mentions the affected actor, action being requested and resource to be affected, a statement of access granted or access denied, and in the denied case a reason why.

src/commons/error.rs
```diff
@@ -111,6 +111,45 @@ impl fmt::Display for RoaDeltaError {
     }
 }
 
+// ApiAuthError is *also* implemented as a separate enum,
```
ximon18, line 114: A few suggestions here:
  - rewrap this comment to 120 characters.
  - change "keep these errors as a separate container" to "keep these errors separate" as I find the reference to a container confusing.
  - it's untrue that "they all originate in interactions with the OIDC provider" as e.g. the MasterTokenAuthProvider and ConfigFileAuthProvider also both return the ApiInvalidCredentials error, and the ApiInsufficientRights error is returned from multiple places.

src/commons/error.rs
```diff
@@ -111,6 +111,45 @@ impl fmt::Display for RoaDeltaError {
...
+// so that we don't have to implement the Clone trait for
+// all of the Error enum.
+// Also it makes kind of sense to keep these errors as a separate container,
+// since they all originate in interactions with the OIDC provider (or lack thereof)
+#[derive(Debug, Clone)]
+pub enum ApiAuthError {
+    ApiInvalidCredentials(String),
+    ApiLoginError(String),
```
ximon18, line 122: I don't have a better suggestion but some of the other errors could be considered to be "login errors" so I'm thinking we should revisit this name.

src/commons/error.rs
```diff
@@ -281,10 +321,11 @@ impl fmt::Display for Error {
             Error::PostTooBig => write!(f, "POST body exceeds configured limit"),
             Error::PostCannotRead => write!(f, "POST body cannot be read"),
             Error::ApiInvalidCredentials(e) => write!(f, "Invalid credentials: {}", e),
-            Error::ApiAuthError(e) => write!(f, "Authentication error: {}", e),
+            Error::ApiLoginError(e) => write!(f, "Login error: {}", e),
```
ximon18, line 324: See above comment about "login error" being too generic and overlapping with other errors, e.g. "invalid credentials". Or is this error supposed to be generic and covering cases not covered by more specific errors?

src/daemon/http/server.rs
```diff
@@ -766,7 +766,10 @@ macro_rules! aa {
                 // which causes the error message to nest, e.g.
                 //   "Invalid credentials: Invalid credentials: Session expired"
                 match err {
-                    Error::ApiInvalidCredentials(_) | Error::ApiInsufficientRights(_) => {
+                    Error::ApiInvalidCredentials(_)
```
ximon18, line 769: One oddity here is that HttpResponse::forbidden() calls self::response_from_error(ApiInsufficientRights), i.e. is the same as one of the cases in the other part of the match.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -617,20 +661,76 @@ impl AuthProvider for OpenIDConnectAuthProvider {
...
+                        );
+                        match err {
+                            // This is the Error returned by the OpenID Connect Provider if the session was terminated
+                            // by them. The user should be able to create a new session by logging in again.
+                            CoreErrorResponseType::InvalidGrant => {
+                                warn!("OpenID Connect: invalid_grant {:?}", err);
+                                return Err(Error::ApiInvalidCredentials("The session has ended.".to_string()));
+                            }
+                            CoreErrorResponseType::InvalidRequest | CoreErrorResponseType::InvalidClient => {
+                                warn!("OpenID Connect:  RFC 6749 5.2 {:?}", err);
```
ximon18, line 693: The double space after "Connect:" in "Connect:  " should be a single space.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -617,20 +661,76 @@ impl AuthProvider for OpenIDConnectAuthProvider {
...
+                                    "There is a unknown authentication problem.".to_string(),
+                                ));
+                            }
+                            // If changes are made to the roles of the user, the client or
+                            // the scope on the side of the OpenID Connect Provider,
+                            // the token refresh may get one of these errors.
+                            CoreErrorResponseType::UnauthorizedClient
+                            | CoreErrorResponseType::UnsupportedGrantType
+                            | CoreErrorResponseType::InvalidScope => {
+                                warn!("OpenID Connect:  RFC 6749 5.2 {:?}", err);
```
ximon18, line 704: The double space after "Connect:" in "Connect:  " should be a single space.

tests/ui/openid_connect_mock.rs
```diff
@@ -106,14 +161,34 @@ type KnownUsers = HashMap<KnownUserId, KnownUser>;
...
+    })
+    .level(log::LevelFilter::Trace)
+    .chain(std::io::stdout())
+    .apply()
+    .map_err(|e| format!("Failed to init stderr logging: {}", e));
+    
+    start(2500).await;
+}
+
+pub async fn start(delay_secs: u64) -> Option<task::JoinHandle<()>> {
```
ximon18, line 184: What was the motivation for making this configurable, especially since the caller uses a hard-coded value?

tests/ui/openid_connect_mock.rs
```diff
@@ -74,13 +77,65 @@ type CustomIdTokenFields = IdTokenFields<
 type CustomTokenResponse = StandardTokenResponse<CustomIdTokenFields, CoreTokenType>;
 // end cascade
 
+// This failure mode can be invoked for a user at user creation time.
+// The failure mode will be activated on a 'refresh_token' request
+// from krill to the OIDC Provider
```
ximon18, line 82: s/krill/Krill/.

tests/ui/openid_connect_mock.rs
```diff
@@ -74,13 +77,65 @@ type CustomIdTokenFields = IdTokenFields<
...
+#[derive(Clone, Debug)]
+enum RefreshTokenFailureMode {
+    // These are the RFC 6749 5.2 Errors
+    InvalidRequestErrorResponse,
+    InvalidClientErrorResponse,
+    InvalidGrantErrorResponse,
+    InvalidScopeErrorResponse,
+    UnauthorizedClientErrorResponse,
+    UnsupportedGrantTypeErrorResponse,
+    // These various generic failure modes,
```
ximon18, line 92: I'm not sure I understand this comment, I would expect "these ... <insert verb>" but the verb missing missing. What were you trying to say here?

tests/ui/openid_connect_mock.rs
```diff
@@ -74,13 +77,65 @@ type CustomIdTokenFields = IdTokenFields<
...
+    // without OAUTH/OpenID Connect specific errors
+    NoResponse,
+    SlowResponse(Duration),
+    Error500Response,
+    Error503Response,
+}
+
+impl SerdeSerialize for RefreshTokenFailureMode {
```
ximon18, line 100: Huh, interesting way to do this. I learned something :-)

.gitignore
density215: I did, but then clippy barks all over the place, although it compiles just fine. It was just a link to the mock auth provider, so that I could `cargo run --examples openid_connect_mock`

src/commons/error.rs
density215: I wanted to differentiate between errors _only_ occurring at login and errors that could occur _both_ at login and at (re)authentication time.

src/commons/error.rs
density215: See the above comment

tests/ui/openid_connect_mock.rs
density215: These errors are not mentioned in Section 5.2, They are error codes that originate from the generic network or server-side problems.

tests/ui/openid_connect_mock.rs
density215: My take at the lowest-effort-way of making the file openid\_connect\_mock.rs into a stand-alone server, that would run for the specified amount of seconds (See the main() function).

tests/ui/cypress/specs/multi_user_openid_connect.js
```diff
@@ -122,6 +152,14 @@ describe('OpenID Connect users', () => {
     // by the time Krill verifies it!
     cy.wait(6000)
 
+    // Try to create a CA, by typing in the input, clicking the 'Create CA' button
+    // and then clicking 'Ok'. This should fail, since the token can't be refereshed.
+    cy.intercept('POST', '/api/v1/cas').as('createCA')
+    cy.get('main input').type('some-handle-name')
+    cy.get('main button').click()
+    cy.get('.el-message-box__btns button:nth-child(2)').click()
+    cy.wait('@createCA').its('response.statusCode').should('eq', 401)
```
ximon18, line 161: I was wondering from reading the test what the UI experience is for the user, and that it would be good to verify that here. However, on trying it myself I find a problem. If I login as user `shorttokenwithoutrefresh@krill` then I receive a short-lived access token and no refresh token. When the create CA form is submitted after the token has been allowed to expire I therefore see an error in the UI:

> A (temporary) error occured while trying to authenticate your request. Please try again later.

Unfortunately this guidance is incorrect. No amount of retrying will help as the token is invalid and cannot be refreshed and so remains invalid. In this case we need to direct the user back to the login form. We should perhaps even do so transparently and automatically in case the provider is able to use user-agent cookies to magically "extend" the session by automatically issuing a new access token.

tests/ui/cypress/specs/multi_user_openid_connect.js
```diff
@@ -138,7 +176,90 @@ describe('OpenID Connect users', () => {
     cy.get('input[name="username"]')
   })
 
-  it('Login receives short-lived refreshable token', () => {
+  it('Login with short-lived non-refreshable token and try to create a CA', () => {
```
ximon18, line 179: I'm confused about this test. I have a feeling of deja vu like we discussed this before, sorry if that's the case. A diff of this test with the test above seems to only show that this one is a subset of the one above, as the one above also calls `visit('/')` at the end and this one does not. The other differences are that the test above uses updated lookup selectors and this test does not. It seems to me like this test should be removed and keep only the one above, but I have the feeling you said there was a valuable difference, but I can't see it at the moment. What am I missing?

tests/ui/openid_connect_mock.rs
```diff
@@ -106,14 +161,34 @@ type KnownUsers = HashMap<KnownUserId, KnownUser>;
 const DEFAULT_TOKEN_DURATION_SECS: u32 = 3600;
 static MOCK_OPENID_CONNECT_SERVER_RUNNING_FLAG: AtomicBool = AtomicBool::new(false);
 
-pub async fn start() -> Option<task::JoinHandle<()>> {
+#[tokio::main]
+pub async fn main() {
+    // Log to stdout.
+    #[cfg(feature = "multi-user")]
```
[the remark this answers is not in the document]
density215: Leftover over from when I tried to use use `env_logger` based on the `mutli-user` feature. removed.

tests/ui/openid_connect_mock.rs
```diff
@@ -106,14 +161,34 @@ type KnownUsers = HashMap<KnownUserId, KnownUser>;
...
+                "{} [{}] {}",
+                chrono::Local::now().format("%Y-%m-%d %H:%M:%S"),
+                record.level(),
+                message
+            ))
+    })
+    .level(log::LevelFilter::Trace)
+    .chain(std::io::stdout())
+    .apply()
+    .map_err(|e| format!("Failed to init stderr logging: {}", e));
```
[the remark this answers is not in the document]
density215: This would only when run as a stand alone mock server with `--example` and only when logging to stdout, so you would see this in the console where you started that. All other uses of the mock server would not go through `main`

I included this mainly to silence clippy, btw.

tests/ui/cypress/specs/multi_user_openid_connect.js
density215: I didn't want to convey to the user that they might get away with trying the button again, but that they may also have to start all over again, that's why I said 'try again later' instead of 'retry'. That might not be clear enough, we may rephrase to something like: 'please try again later of try to log in again'?

This is the thing with transient errors, right? They may be transient, but they may also be take way too long to resolve

density215: Maybe in this particular case we may also throw a ApiAuthPermanentError instead of ApiAuthTransientError

ximon18: In this particular case however no amount of waiting or retrying will ever work, only logging in again will work.

ximon18: > Maybe in this particular case we may also throw a ApiAuthPermanentError instead of ApiAuthTransientError

No, I don't think that's right either as the issue is not permanent. What needs to happen is that the user needs to re-login. Is it okay on HTTP 401 to redirect the user to the login process with a return-here-url-param to continue where they were before, perhaps transparently which is nice, or in the case that it isn't transparent because a login form is displayed or a provider error occurs, is it not nice because the user will think hang on why am I looking at a login form or an error from the provider? The middle ground is to include a login link in the error message. One way to handle this could be to cause the Login view to show as that has the ability to show only an error message with a link to the login form, the only downside being it takes the user away from the current thing they are seeing. Another alternative is to literally include a login link in the error message that is shown in the context of what the user is doing right now.

ximon18: I think what we can do in this PR is update this test to verify that the message is shown thereby highlighting the current behaviour, and leave a TODO comment at that point in the test script saying that we should in some way automatically re-login the user or guide them to relogin.

tests/ui/cypress/specs/multi_user_openid_connect.js
density215: The one above doesn't need to try to create a CA (doesn't hurt either?), think, but the refresh is worthwhile to test: If the token has expired and can't be refreshed, a refresh should _not_ show you this page, but the OIDC login page.

But yes, at least I could used the cy.contains function. updated.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -357,49 +363,84 @@ impl OpenIDConnectAuthProvider {
...
+                        ));
                     }
                 }
             }
             None => {
-                debug!(
-                    "OpenID Connect: Could not extend login session for user '{}': no refresh token available",
-                    &session.id
-                );
+                return Err(CoreErrorResponseType::Extension(
```
ximon18, line 439: As discussed move this case to the caller, i.e. don't try to refresh if we don't have a refresh token rather than attempting to encode this as an error.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -357,49 +363,84 @@ impl OpenIDConnectAuthProvider {
...
+                                                // and early returned.
+                                                return Ok(Auth::Bearer(new_token));
                                             }
                                             Err(err) => {
-                                                warn!(
-                                                    "OpenID Connect: Could not extend login session for user '{}': {}",
-                                                    &session.id, err
-                                                );
+                                                return Err(CoreErrorResponseType::Extension(format!(
+                                                    "Error while encoding the refreshed token {}",
```
ximon18, line 372: Suggestion: Prefix the error message with `krill_internal_error: ` to make it easy to distinguish from any actual provider error messages.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -357,49 +363,84 @@ impl OpenIDConnectAuthProvider {
...
+                                            }
+                                            openidconnect::RequestTokenError::Other(err_string) => {
+                                                match err_string.as_str() {
+                                                    "temporarily_unavailable" | "server_error" => {
+                                                        return Err(CoreErrorResponseType::Extension(
+                                                            err_string.to_string(),
+                                                        ))
+                                                    }
+                                                    _ => {
+                                                        return Err(CoreErrorResponseType::Extension(format!(
```
ximon18, line 412: Make `Error` lowercase. Also, remove the #1 as this is not similar to the #2 case below, in this case this is an unexpected response from the 3rd party system or the openid connect Rust crate, while #2 is an internal error in Krill.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -357,49 +363,84 @@ impl OpenIDConnectAuthProvider {
...
                                     }
                                 }
                             }
                             None => {
                                 // should be unreachable
-                                warn!(
-                                    "OpenID Connect: Could not extend login session for user '{}': {}",
-                                    &session.id, "Connection to OpenID Connect provider not yet established"
-                                );
+                                return Err(CoreErrorResponseType::Extension(
```
ximon18, line 425: Suggestion: Prefix the error message with `krill_internal_error: ` to make it easy to distinguish from any actual provider error messages.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -357,49 +363,84 @@ impl OpenIDConnectAuthProvider {
...
+                                ));
                             }
                         }
                     }
                     Err(err) => {
-                        warn!(
-                            "OpenID Connect: Could not extend login session for user '{}': {}",
-                            &session.id, err
-                        );
+                        return Err(CoreErrorResponseType::Extension(
```
ximon18, line 432: Suggestion: Prefix the error message with `krill_internal_error: ` to make it easy to distinguish from any actual provider error messages. Also make Error lowercase and remove the #2 (see comment above regarding #1).

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -597,13 +638,16 @@ impl OpenIDConnectAuthProvider {
 }
 
 impl AuthProvider for OpenIDConnectAuthProvider {
-    // TODO: handle error responses from the provider as per RFC 6749 and OpenID
     // Connect Core 1.0 section 3.1.26 Authentication Error Response
+    // OAuth 2.0 RFC-674 4.1.2.1 (Authorization Request Errors) & 5.2 (Access Token Request Errors)
 
+    /// Tries to extend the session with the OIDC Provider once or multiple times (retries and/or back-off).
```
ximon18, line 644: This comment isn't quite correct. This fn is also responsible for completing the handshake for initial creation of a session, not just for attempting to extend an existing session.

src/commons/actor.rs
ximon18: Deferred to #420.

tests/ui/cypress/specs/multi_user_openid_connect.js
ximon18: I've updated the test to verify the message shown to the user. I decided not to add a TODO, rather any error experience improvements should be handled as part of #419. Still to do:
- [ ] Add a translation string for this new `api-auth-session-expired` error label.

ximon18: I added `api-auth-session-expired` to Lagosta PR #87 and synced this PR with the modified Lagosta `new-api-error-strings` branch.

ximon18 (approved):

ximon18:
The Lagosta `new-api-error-strings` branch has been merged to master. This branch already incorporates those changes. Merging this branch to master.

Issue #402, opened 2021-01-20
Testbed mode is broken in the v0.8.1-bis branch


ximon18:
There are two issues with testbed mode at present in the `v0.8.1-bis` branch:

1. None of the "Register CA", "Unregister CA", etc tabs are selected on first load of /index.html#/testbed, and so you see this:
![image](https://user-images.githubusercontent.com/3304436/105194584-08df1300-5b3a-11eb-8e0e-2cc680439335.png)

2. The handling of /tested/ API errors is broken as `APIService.js` in Lagosta was changed to return the `data` part of the error response and not the `response` object itself but the `Testbed.vue` code was not updated to follow this change. This results in errors in the console and the testbed UI "hanging" if there is a problem with the JSON (converted from XML end-user input) sent to Krill by the testbed UI.

ximon18:
Improvements in this area have been merged to the Krill dev branch but not yet released or deployed on testbed.rpki.nlnetlabs.nl.

PR #403, opened 2021-01-20
Testbed mode should be more liberal in handling formatting variations in child and publisher registration XML. (#369)

### LINKED ISSUE #369: Gracefully handle line breaks in child request XML
When a child request XML has line breaks, in particular the `<child_bpki_ta>` block, Krill will not accept it. Krill should probably be more liberal in accepting variations in formatting.

ximon18:
Testbed UI improvements for #369:
- Don't show JSON error messages if Krill cannot parse the submitted "XML" (which the testbed UI converted first to a JSON form), instead just warn that the XML could not be parsed correctly.
- Trim leading and trailing whitespace in XML element text nodes.
- Remove embedded whitespace in Base64 encoded xxx_bpki_ta XML element text values (as Krill does not accept embedded whitespace in the Base64)
- Warn about incorrectly Base64 encoded xxx_bpki_ta XML element text values. (does not warn about embedded whitespace as JS atob() accepts it);
- Warn about non-ASCII characters in xxx_bpki_ta XML element text values.
- Cypress based UI tests covering all of the above cases for both child and publisher requests to the testbed.

Depends on and includes the compiled form of the changes in Lagosta at https://github.com/NLnetLabs/lagosta/pull/85.

This PR also fixes the issues reported in #402 as this was necessary to fix #369.


ximon18:
@Alex, @timbru : to summarize the end-user experience regarding the updated behaviour:

### For `xxx_bpik_ta` values containing whitespace

Prior to submitting the values to the Krill server, the UI will trim leading, trailing and embedded ASCII whitespace in the certificate string, so even a multi-line Base64 certificate with leading and trailing line breaks will "just work".

### For problems detected by the UI

If the child or publisher request XML is invalid XML, lacks an expected element or attribute, or the base64 certificate contains non-ASCII characters or cannot be decoded as base64, when the user pushes the Register button a _translatable_ error in red text is written beneath the XML input field, one of (quoting from the English translation string file):
```
Invalid XML: {err}
Missing element {el}
Missing child element {el} of element {parent}
Missing attribute {attr} on element {el}
Element {el} cannot be empty
Attribute {attr} on element {el} cannot be empty
Element {el} cannot contain non-ASCII characters
Child Handle is required
Publisher Handle is required
Element {el} must contain a correctly Base64 encoded self-signed X.509 BPKI certificate
```
For example:
image

### For problems detected by the Krill server
If the XML is accepted by the UI a JSON request is constructed from elements of the XML and is sent to the Krill API. Krill may then reject it with the following error messages which are written in red _above_ the XML input field *AND* are shown in a popup in the top right of the window:
```
Invalid JSON: {}
Submitted bytes are invalid UTF8
Input contains non-ASCII chars (maybe whitespace?)
```
For example:
image

**Note:** While Krill can generally respond with the "Invalid JSON: {}" to any HTTP POST API call for which Krill is unable to parse a received JSON body, in this specific case the testebd UI detects the error label `api-json` and instead shows a (translatable) "Your registration details could not be parsed correctly" error instead. This is because the user provides XML so getting an error back about JSON is highly confusing and the actual error message even gives a cryptic location to where the issue is (the user can't make use of the specified row and column as they relate to a JSON text the user has never seen).

### Missing translations
While writing this I realize that the following error messages are translatable but _lack translations_, and so these English strings will be shown as is:

| Error label | English message returned by Krill |
|----|----|
| `api-json` | Invalid JSON: {} |
| `api-invalid-utf8` | Submitted bytes are invalid UTF8 |
| `api-non-ascii` | Input contains non-ASCII chars (maybe whitespace?) |

ximon18:
If you want to test this locally, one way is as follows:
```
$ git clone
$ git checkout 369-testbed-should-trim-input
$ cargo build
$ KRILL_TEST=1 KRILL_TESTBED_RRDP=https://localhost:3000/rrdp/ KRILL_TESTBED_RSYNC=rsync://localhost/repo/ target/debug/krill -c test-resources/ui/testbed_ui.conf
```

Then browse to https://localhost:3000/index.html#/testbed.

You can find some good and bad test XMLs in the [test-resources/testbed](https://github.com/NLnetLabs/krill/tree/369-testbed-should-trim-input/test-resources/testbed) subdirectory of the branch.

timbru (approved):
Nice! Thanks

PR #410, opened 2021-01-28
Support OAuth 2.0 token revocation (resolves #385)

### LINKED ISSUE #385: Finalize support for OpenID Connect revocation_endpoint

Support for detecting claimed support in the OpenID Connect discovery response for OAuth 2,0 Token Revocation [RFC-7009](https://tools.ietf.org/html/rfc7009) is implemented (e.g. as advertized here https://accounts.google.com/.well-known/openid-configuration via `revocation_endpoint` in the response).

However, the revocation endpoint is never actually contacted. `build_logout_url()` in `providers/openid_connect/provider.rs` just stores the Krill public UI URI for later returning in `logout()` to the browser, so the revocation endpoint is never contacted.

ximon18:
This PR extends Krill so that it doesn't just detect that the provider supports [RFC 7009 OAuth 2.0 Token Revocation](https://tools.ietf.org/html/rfc7009) but actually contacts the revocation endpoint.

Changes include:
- Not building the logout URL on OpenID Connect initialization but at the moment of logout, as the URL is no longer static but now needs to include (in the token revocation case) a dynamic component, i.e. the token to revoke.
- Code to construct the HTTP POST request to be made to the token revocation endpoint.
- Extension of the tests and the OpenID Connect mock to support testing with both a provider that supports RP-Initiated Logout based logout and a provider that supports OAuth 2.0 Token Revocation based logout.
- Default to directing the user to the Krill UI index page if the provider does not offer any of the supported logout mechanisms and no `logout_url` has been provided by the operator in the Krill config file.

Depends on:
- PR #396 for the base on which to implement provider error response handling.

Still to do:
- [x] Try with an access token if we don't have a refresh token.
- [x] Extend the automated tests to simulate the error below.
- [x] Extend the automated tests to try both with a provider that does and a provider that does not offer a refresh token.
- [x] Handle revocation endpoint errors, currently logout results in a HTTP 400 from the provider, the logout button just seems to do nothing and the browser console shows an error of the form:
    ```
    Failed to load resource: the server responded with a status of 400 ()
    Object: data: {
        label: "general-error",
        msg: "Unable to revoke OAuth 2.0 refresh token: Token revocation failed with status code 400 Bad Request",
        args: ...
    }
    ```
- [x] Authenticate the request sent to the revocation endpoint, currently the provider responds with an error in the HTTP body:
    ```
    {
        "error":"invalid_request",
        "error_description":"client id parameter is missing"
    }
    ```
- [ ] See if `TokenRevocationRequest` can be used more like the built-in request types supported by the `openidconnect` crate. See https://github.com/ramosbugs/openidconnect-rs/issues/37. **Update**: my changes to support token revocation have been incorporated into released alpha versions of the `openidconnect-rs` and `oauth2-rs` crates.
- [x] Unstable UI test works locally but [fails in GitHub Actions](https://github.com/NLnetLabs/krill/runs/1787327191?check_suite_focus=true#step:5:4507) with `AssertionError: Timed out retrying after 4000ms: expected 'https://localhost:3000/index.html#/onboarding' to not include 'https://localhost:3000'`.


ximon18:
Currently a test is failing, `CypressError: Timed out retrying after 5000ms: `cy.wait()` timed out waiting `5000ms` for the 1st request to the route: `getLoginForm`. No request ever occurred.`. However, I think it makes sense to put this PR on hold until PR #396 is finished and merged into `v0.8.1-bis`, then rebase/merge those changes into this PR, adjust this PR in light of those changes and then see where we are again.

ximon18:
Obsoleted by PR #436.

PR #412, opened 2021-02-02
Ignore the functional test by default as it takes >1 minute to run and enable testing of ignored tests in GH jobs.

ximon18:


timbru (approved):

ximon18:
Hmm, it seems that `--include-ignored` only wors if `-Z` is passed to the test binary. I'm not comfortable using `-Z` for tests we depend on. I'll abandon this PR as this was meant to be a quick improvement, it's not something I want to invest time in at the moment.

PR #415, opened 2021-02-04
Fix master token access restrictions when not using the multi-user feature.

ximon18:
When not in possession of a valid bearer token it is still possible, in the `v0.8.1-bis` branch, to call restricted REST APIs such as create CA. This can be seen by running `krillc add` with an invalid API token.

This PR ensures that, irrespective of whether the `multi-user` feature is enabled or not, that anonymous users are unable to perform restricted actions, and adds a minimal test that catches the failure by trying to create a CA using the wrong API token. We could expand the test to verify that all restricted API endpoints are indeed restricted but that can be done separately, not having it now in this PR doesn't diminish the value of this PR.

Note: This regression was introduced in the `v0.8.1-bis` branch, it was never released. It wasn't detected until now because all of the existing tests either only use a correct master token (hard-coded to `secret` in the default integration test Krill config and in the Rust test suite helper functions) or only call restricted REST APIs after first logging in and obtaining a valid bearer token.


ximon18:
The new `auth_check` test is correctly "failing", e.g. see [this GH Actions log fragment](https://github.com/NLnetLabs/krill/pull/415/checks?check_run_id=1834599160#step:5:793).

density215 (approved):

density215 (approved):

PR #424, opened 2021-02-16
Check the OpenID Connect CSRF token (resolves #383).

### LINKED ISSUE #383: Check the OpenID Connect CSRF token.

ximon18:
When logging in to an OpenID Connect provider via a 3rd party login form, the URL to which the user is HTTP redirected contains security features including a so-called CSRF token query parameter called `state`.

After the login form has been submitted and accepted by the 3rd party provider the provider redirects the end user back to Krill via their user agent (aka browser). This second redirect also includes a `state` query parameter and its value should be the same as that generated by Krill when the initial redirect URL is generated.

Prior to this PR the state value was generated per login request but was not checked upon receiving the redirect from the 3rd party provider. This PR:
- Adds a short lived secure Set-Cookie HTTP header to the initial redirect to the 3rd party login form, with its value being the hash of the CSRF token.
- On receipt of the 3rd party provider redirect to Krill, retrieves the cookie and checks if the hash of the `state` parameter matches the cookie value.
- Adds a test which deliberately returns the wrong `state` query parameter value from the OpenID Connect mock server and verifies that the end user is shown an error.

This PR has also been manually tested including with AWS Cognito and Azure ActiveDirectory.

### DIFF
--- a/src/daemon/auth/authorizer.rs
+++ b/src/daemon/auth/authorizer.rs
@@ -4,6 +4,7 @@ use std::any::Any;
 use std::collections::HashMap;
 use std::sync::Arc;
 
+use crate::{commons::api::Token, daemon::auth::common::permissions::Permission};
 use crate::commons::error::Error;
 use crate::commons::KrillResult;
 use crate::constants::ACTOR_DEF_ANON;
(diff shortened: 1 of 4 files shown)


ximon18:

PR #429, opened 2021-02-23
Trap startup error in tests

ximon18:
When not running from a test a startup failure in Krill is logged by `fn main()`, but when run from a test such errors are lost. This change detects and logs them.


ximon18:
Merging as agreed as the target branch is not master and this only affects tests, not Krill itself.

PR #430, opened 2021-02-23
Log types passed to oso when trace is enabled

ximon18:
Sometimes when investigating an Oso policy problem you need to know which type was passed to the rule engine, not just the display value of that type. This change logs the Debug representation of the types passed when trace level logging is enabled.


ximon18:

PR #431, opened 2021-02-23
Stop Cypress when UI test is aborted

ximon18:
Krill tests that drive the UI do so via a Docker container instance of the Cypress test runner and headless Chrome browser.

Unfortunately, if CTRL-C is pressed while `cargo test` is running the test suite is stopped but the Docker container keeps running.

Use a CTRL-C handler to issue a `docker kill` command if the test is aborted while running so that we don't leave the container running.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -423,6 +423,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctrlc"
+version = "3.1.8"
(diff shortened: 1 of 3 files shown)


ximon18:
Merging as agreed as the target branch is not `master` and this only affects tests, not Krill itself.

Issue #428, opened 2021-02-19
Remove second reqwest dependency added by the `multi-user` feature


ximon18:
Krill depends on `reqwest` v0.10.x. If the `multi-user` feature is enabled a second dependency on `reqwest` v0.9.x is introduced. This happens because by default the `openidconnect-rs` v1.x crate depends on `reqwest` v0.9.x. As `openidconnect-rs` v1.x has a feature flag for `reqwest-010` and `openidconnect-rs` v2.x depends on `reqwest` v0.10.x by default, one might assume we can just switch to using `reqwest` v0.10.x throughout Krill.

However it isn't so straightforward:
- Enabling `reqwest` v0.10.x with `openidconnect-rs` causes a runtime panic: `panicked at 'Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.' reqwest blocking`. I believe this is because the implementation of the blocking API in v0.10.x is implemented internally as an async runtime wrapper around the async `reqwest` API and you cannot have nested async runtimes.
- Switching to the async `reqwest` API, to match the rest of Krill, would involve changing a lot of code as making a function deep inside the OpenID Connect Krill AuthProvider use a call to `.await` will require it and all of its callers to be `async` functions, and currently the `async` code stops higher up the call tree. This is doable, it's just work.
- A bigger problem however is that the functions of the `AuthProvider` trait cannot be changed to be `async` because Rust doesn't yet support async traits. The workaround is to use the [`async-trait`](https://docs.rs/crate/async-trait/) crate which is apparently recommended as long as you don't mind a little extra heap allocation.
- Going down the `async-trait` path eventually hits a problem that the OpenID Connect AuthProvider is not `Send` because of something in the `openidconnect-rs` crate `Client` type. To work around this, and maybe this is a good idea anyway to address other issues (e.g. multiple (almost) concurrent requests causing overlapping requests to exchange a refresh token for a new access token)), we might need to make a background worker thread that does the communication with the 3rd party OpenID Connect provider system and use a `mpsc` model to communicate with the worker thread. Some the [spike-incomplete-one-async-reqwest-dependency](https://github.com/NLnetLabs/krill/tree/spike-incomplete-one-async-reqwest-dependency) branch for a partial exploration of this approach.
[...]

PR #436, opened 2021-03-03
Logout enhancements (closes #385, closes #397, fixes #408, closes #425)

### LINKED ISSUE #385: Finalize support for OpenID Connect revocation_endpoint

### LINKED ISSUE #397: Upgrade to openid-connect crate 2.x version?
See: https://github.com/ramosbugs/openidconnect-rs/issues/35#issuecomment-761250629

### LINKED ISSUE #408: OpenID Connect RP-Initiated Logout 1.0 implementation is not spec compliant
https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout says:

> post_logout_redirect_uri ... An `id_token_hint` is also REQUIRED when this parameter is included.

And https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RedirectionAfterLogout says:

> An `id_token_hint` carring an ID Token for the RP is also REQUIRED when requesting post-logout redirection; if it is not supplied with post_logout_redirect_uri, the OP **MUST NOT** perform post-logout redirection

However, the current implementation of support for RP-Initiated logout in Krill in branch `v0.8.1-bis` does NOT add the `id_token_hint` to the logout URL to which the user agent is redirected and thus post logout redirect will be ignored by spec compliant OpenID Connect providers (the `OP` in the spec text above).

### LINKED ISSUE #425: Support config defined logout URL
Logout via OpenID Connect discovered mechanisms such as [RP-Initiated Logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) and [OAuth 2.0 Token Revocation](https://tools.ietf.org/html/rfc7009) is nice, if supported by the provider.

If not supported the customer may still want, as a mirror to the customer specific login experience, a customer specific logout experience as well rather than just returning to the login page post logout.

This could be supported through a configuration option to specify a post logout URL. The user would be directed by Lagosta to this URL after being logged out of Lagosta/Krill. If RP-Initiated logout support is discovered and the config file specifies a logout URL then RP-Initiated logout will not be used (as that would already take the user to a provider controlled logout experience).

A manually defined logout URL can also be used, at least in some cases such as AWS Cognito, to perform a logout [using a mechanism that Krill is not aware of](https://docs.aws.amazon.com/cognito/latest/developerguide/logout-endpoint.html). 

The behaviour in the various scenarios should then be as follows:

- _Note: Discovered RP-Initiated and/or token revocation endpoints will be ignored if not spec compliant (e.g. if not HTTPS URLs)._
- _Note: In all logout scenarios Lagosta will forget the bearer token that it was using to authenticate with the Krill API._

| Config File logout URL specified? | RP-Initiated Logout supported? | Token Revocation supported? | Action taken by Krill |
|---|---|---|---|
| :x: | :x: | :x: | Direct the user to the Krill login page |
| :x: | :x: | :heavy_check_mark: | Revoke the token then direct the user to the Krill login page. |
| :x: | :heavy_check_mark: | :x: | Direct the user to the customer logout portal. If redirected back to Krill, direct the user to the login page. |
| :x: | :heavy_check_mark: | :heavy_check_mark: | Behave as if only RP-Initiated Logout is supported. |
| :heavy_check_mark: | :x: | :x: | Direct the user to the configured logout URL. |
| :heavy_check_mark: | :x: | :heavy_check_mark: | Revoke the token then direct the user to the configured logout URL. |
| :heavy_check_mark: | :heavy_check_mark: | :x: | Direct the user to the configured logout URL. |
| :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | Direct the user to the configured logout URL. |

ximon18:
- Formalize support for different logout strategies and add the fallback strategy.
- Generate the logout URL at logout time in preparation for supporting dynamic logout requests (as needed by token revocation).
- Secure the connection to the mock OpenID Connect provider with a self-signed TLS certificate.
- Allow self-signed certificates for HTTPS connections to localhost (same policy as elsewhere in Krill).
- Upgrade openidconnect-rs to latest v2.0.0 alpha to gain contributed support for OAuth 2.0 Token Revocation. (#385 and #397)
- Use reqwest 0.9.x directly instead of via the openidconnect-rs crate (we cannot use the v0.11.x reqwest that comes with the crate as (a) it doesn't permit self-signed certificates, (b) the blocking implementation was changed to be async which causes problems when inside an existing async runtime, and (c) switching the OpenID Connect client code over to be async is non-trivial - see #428).
- Pass the ID token as `id_token_hint` to the OpenID Connect RP-Initiated Logout 1.0 endpoint. (#408)
- Refined logic for the various logout mechanism permutations. (#425).
- Require OpenID Connection RP-Initiated Logout 1.0 and OAuth 2.0 Token Revocation endpoints to be HTTPS per the specs.
- Passes manual testing with Microsoft Azure Active Directory RP-Initiated Logout support and Google Compute Cloud OAuth 2.0 Token Revocation support.
- Added a Google Cloud Platform example to the comments in the default config file.
- Updated and added tests.
- Fixed logout and token revocation in the mock OpenID Connect provider to actually terminate login sessions.
- Handle a race condition in Lagosta where null user data was accessed that was just deleted due to logout.
- Handle errors from the Krill logout endpoint in Lagosta.


ximon18:

Issue #28, opened 2018-12-21
Multiple CAs in single server.


timbru:
Support having multiple CAs in one application, let user/API specify which one to use.

Also implies we think about the user -> organisation -> CAs authorisation mapping.

timbru:
(Long) done.

PR #439, opened 2021-03-12
Improvements and fixes for e2e testing on the Krill dev branch

ximon18:
(depends on [rpki-deploy PR #28](https://github.com/NLnetLabs/rpki-deploy/pull/28))

- OpenAPI YML corrections.
- Support the new way to activate the TA (in the Krill Docker image).
- FIX: Include the correct ta.cer URI in the TAL.
- Use two seprarate Python libraries for Krill: one for CA REST SAPIs and one for PUB REST APIs.
- FIX: Test for the actual CAs and resources to create, as the presence of the testbed CA violates the previous check assumptions.
- FIX: Resource ASN, v4 and v6 values can no longer be assumed to have the same sort order as when given to Krill.
- Factor out test suite code into helper functions for better readability and maintainability.
- Retry RTR fetching (needed for ROAs obtained from Rcynic Lihttpd server served JSON as otherwise connecting too early results in SyncTimeout).

Note:
- Works locally, not yet tested via GH Actions.
- Currently depends on a personal test build of a Docker image of the main RTRTR branch. v0.1.1 could not be used as it requires the `signature` metadata field to be present which is not the case for the custom scripting that creates ROA JSON output for Rcynic and rpk-client in the Krill e2e test. With goRTR it worked because goRTR supports using unsigned ROA JSON as input.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -72,11 +72,6 @@ EOF
     fi
 fi
 
-if [ "${KRILL_USE_TA}" == "true" ]; then
-    export KRILL_TESTBED_RSYNC="rsync://${KRILL_FQDN}/repo/"
-    export KRILL_TESTBED_RRDP="https://${KRILL_FQDN}/rrdp/"
-fi
-
 # Launch the command supplied either by the default CMD (krill) in the
 # Dockerfile or that given by the operator when invoking Docker run. Use exec
 # to ensure krill runs as PID 1 as required by Docker for proper signal
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -63,16 +63,18 @@ if [ "$1" == "krill" ]; then
         # script via a "-e KRILL_FQDN=some.domain.name" argument to
         # "docker run".
         cat << EOF >> ${KRILL_CONF}
-rsync_base  = "rsync://${KRILL_FQDN}/repo/" ${MAGIC}
-service_uri = "https://${KRILL_FQDN}/" ${MAGIC}
 log_level   = "${KRILL_LOG_LEVEL}" ${MAGIC}
-use_ta      = ${KRILL_USE_TA} ${MAGIC}
 EOF
 
         log_info "Dumping ${KRILL_CONF} config file"
         cat ${KRILL_CONF}
         log_info "End of dump"
     fi
+fi
+
+if [ "${KRILL_USE_TA}" == "true" ]; then
+    export KRILL_TESTBED_RSYNC="rsync://${KRILL_FQDN}/repo/"
+    export KRILL_TESTBED_RRDP="https://${KRILL_FQDN}/rrdp/"
 fi
 
 # Launch the command supplied either by the default CMD (krill) in the
--- a/src/daemon/krillserver.rs
+++ b/src/daemon/krillserver.rs
@@ -221,7 +221,7 @@ impl KrillServer {
 
                     let repo_info: RepoInfo = pubserver.repo_info_for(&ta_handle)?;
 
-                    let ta_uri = format!("{}ta/ta.cer", uris.rrdp_base_uri());
+                    let ta_uri = format!("{}ta/ta.cer", service_uri);
                     let ta_uri = uri::Https::from_string(ta_uri).unwrap();
 
                     let ta_aia = format!("{}ta/ta.cer", uris.rsync_jail().to_string());
--- a/tests/e2e/simple_rp_roa_test.py
+++ b/tests/e2e/simple_rp_roa_test.py
@@ -5,7 +5,8 @@ import rtrlib
 from retrying import retry, RetryError
 from time import time
 from operator import attrgetter
-from krill_api import *
+import krill_ca_api as krill_ca_api_lib
+import krill_pub_api as krill_pub_api_lib
 
 from tests.util import krill
 from tests.util.docker import docker_project, class_service_manager, function_service_manager, run_command, docker_host_fqdn
@@ -29,7 +30,7 @@ def krill_with_roas(docker_project, krill_api_config, class_service_manager):
     #
     def no_retry_if_forbidden(e):
         """Return True if we should retry, False otherwise"""
-        return not (isinstance(e, ApiException) and e.status == 403)
+        return not (isinstance(e, krill_ca_api_lib.ApiException) and e.status == 403)
 
     def retry_if_not(result):
         """Return True if we should retry, False otherwise"""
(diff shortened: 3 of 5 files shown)


ximon18:
The e2e test is experiencing a failure that I don't see locally:

`    Traceback (most recent call last):
1179
      File "<string>", line 1, in <module>
1180
      File "/tmp/krill-6d94d628/pip-build-3zar8oqj/cryptography/setup.py", line 14, in <module>
1181
        from setuptools_rust import RustExtension
1182
    ModuleNotFoundError: No module named 'setuptools_rust'`

ximon18:
I fixed the Python cryptography module issue mentioned above and after that the tests run but fail. I can't reproduce the issue on my development machine but I am able to reproduce it in a fresh Ubuntu 18.04 LXC container. Am investigating...

ximon18:
The issue turned out to be failure to connect to RTR ports on "localhost" because "localhost" was being resolved to the IPv6 address `::1` rather than the IPv4 address `127.0.0.1`. I have hard-coded `127.0.0.1` for now, not ideal but it seems to work. The Krill end-to-end test is now passing on this branch!

docker/entrypoint.sh
```diff
@@ -75,6 +72,11 @@ EOF
     fi
 fi
 
+if [ "${KRILL_USE_TA}" == "true" ]; then
```
ximon18, line 75: @timbru: We should decide what we want to do here. Released versions of Krill until now supported a KRILL_USE_TA environment variable and when the Docker image detected that the env var was set it used it to set the equivalent but no longer support `use_ta = true` setting in krill.conf. In this edit I kept support for KRILL_USE_TA but made it set the new env vars that need to be set to activate TA + testbed mode, the nearest equivalent to what we had before. Keeping support for KRILL_USE_TA in the Docker image but not in Krill itself feels confusing and wrong. We could just remove this if block and advise people that if they want to use the TA they have to pass `-e KRILL_TESTBED_RSYNC=...` and `-e KRILL_TESTBED_RRDP=...` when invoking `docker run` ? What do you think?

src/daemon/krillserver.rs
```diff
@@ -221,7 +221,7 @@ impl KrillServer {
 
                     let repo_info: RepoInfo = pubserver.repo_info_for(&ta_handle)?;
 
-                    let ta_uri = format!("{}ta/ta.cer", uris.rrdp_base_uri());
+                    let ta_uri = format!("{}ta/ta.cer", service_uri);
```
ximon18, line 224: @timbru: Can you double check that this is okay here, perhaps it wasn't what you had in mind?

docker/entrypoint.sh
timbru: I think:

Remove this if block and advise people that if they want to use the TA they have to pass -e KRILL_TESTBED_RSYNC=... and -e KRILL_TESTBED_RRDP=... when invoking docker run.

Alternatively we could have a testbed binary which takes uris as arguments in addition to krill, or we could have an API / CLI endpoint to set up the testbed on a running Krill instead (and fail in case it's not a clean install). But to be honest, given the amount of use of TESTBED I can live with the env approach for now.

tests/e2e/simple_rp_roa_test.py
```diff
@@ -228,30 +238,46 @@ def test_setup(self):
...
-            # are each of the TEST_ROAS items in r?
-            # i.e. is the intersection of the two sets equal to that of the TEST_ROAS set?
-
-            logging.info(f'Comparing {len(received_roas)} received ROAs to {len(TEST_ROAS)} expected ROAs...')
-            expected_roas = set([roa_to_roa_string(r) for r in TEST_ROAS])
-            assert received_roas == expected_roas
-        except rtrlib.exceptions.SyncTimeout as e:
-            logging.error(f'Timeout (>{service.rtr_timeout_seconds} seconds) while syncing RTR with {service.name} at {docker_host_fqdn}:{service.rtr_port}')
+        @retry(
+            stop_max_attempt_number=3,
```
ximon18, line 251: This 3 retry attempts setting was added for the rcynic -> custom script to produce JSON -> Lighthttpd case which for some reason librtr fails to connect to and to see that there are VRPs available if it connects too soon, but on a retry it works just fine. It seems it might be related to the way RTRTR produces an error in this situation that GoRTR does not but Martin said he believes that the RTRTR behaviour is correct, in combination with a timing difference with the other tests that also use the same "custom script to produce JSON -> Lighthttpd" pattern but don't exhibit a problem.

However, writing it like this causes the other RP tests to also be retried if they fail, but for those RPs that just delays the test result, especially in the case of the RIPE NCC Validator 3 as that is configured in the test framework (see the rpki-deploy repo) to wait for 240 seconds *per try*. With this in mind the test framework also defines the number of retries per RP and sets them all to 1 except for Rcynic which it sets to 3. However, that number isn't used here. I'm not sure if I can pass a variable to `stop_max_attempt_number` here.

docker/entrypoint.sh
ximon18: I'll remove the block.

src/daemon/krillserver.rs
timbru: I think this work: https://github.com/NLnetLabs/krill/pull/441

Is the right direction, but needs to be extended a bit.

Indeed base the default on 'service_uri' but allow overriding it. We could also go for env variables instead of config - perhaps this would be more inline with the 'KRILL_TESTBED_RSYNC' and 'KRILL_TESTBED_RRDP' settings.

timbru: Or rather, I am also fine with just accepting this change here now. And fixing this in the other PR #441 that I just made.

tests/e2e/simple_rp_roa_test.py
```diff
@@ -77,7 +78,58 @@ def wait_until_ca_has_resources(ca_handle, asn, v4, v6):
...
+    #
+    def add_ca(ca_handle):
+        logging.info(f'-> Adding CA "{ca_handle}"')
+        krill_ca_api.add_ca(krill_ca_api_lib.AddCARequest(ca_handle))
+
+        # rfc8183_request = krill_ca_api.get_ca_publisher_request("ta", format="json")
+        # logging.info(f'XIMON: req={rfc8183_request}')
+        krill_ca_api.update_ca_repository(
+            ca_handle,
+            rfc8183='embedded')
```
ximon18, line 94: Note that I made changes in this PR to support NOT using embedded mode (e.g. making Krill use our custom root CA certificate so that it can trust the connection out via NGINX back to itself) but hit some final problem and reverted to using "embedded" when I realized that actually still works. I should switch this back to non-embedded and find out what didn't work and why.

ximon18:
Merging after discussion with @timbru as this helps us make progress with finalizing the work ongoing in the dev branch. We can do more fine tuning later if needed.

PR #441, opened 2021-03-15
Let users configure the AIA to use with TA certificates. Set 'ta_aia'…

timbru:
This adds an option to configure the (rsync) ta AIA in config. This should probably be extended to allow overriding the ta HTTPS uri as well.


src/daemon/config.rs
```diff
@@ -61,6 +61,11 @@ impl ConfigDefaults {
     fn use_ta() -> bool {
         env::var(KRILL_ENV_USE_TA).is_ok()
     }
+
+    fn ta_aia() -> uri::Rsync {
+        uri::Rsync::from_str("rsync://localhost/ta/ta.cer").unwrap()
```
ximon18, line 66: Is localhost the correct default here? Or should it use the `service_uri`?

ximon18:
For context when reviewing, which problem is this intended to solve?

timbru:
@ximon18 Can you have another look? I hope things are clean and clear now!

timbru:
@ximon18 I see this breaks the multi user UI tests - which I cannot (yet) run on my machine. Can you please have look at those?

ximon18:
> @ximon18 I see this breaks the multi user UI tests - which I cannot (yet) run on my machine. Can you please have look at those?

Fixed.

defaults/krill-testbed.conf
```diff
@@ -0,0 +1,63 @@
...
+# content will be overwritten.
+#
+# Manually retrieve the TA certificate from krill and copy it
+# over - it won't change again. You can get it at:
+# https://<yourkrill>/ta/ta.cer
+ta_aia = "rsync://testbed.example.com/ta/ta.cer"
+
+# TA URI
+#
+# Like above, copy the TA certificate to a location where it may
```
ximon18, line 60: @timbru: As this is served via HTTPS it could access to it could also be via a proxy, it doesn't _have_ to be copied, right?

src/daemon/config.rs
```diff
@@ -922,6 +980,27 @@ mod tests {
...
+    }
+
+    #[test]
+    fn should_parse_testbed_config_file() {
+        // Config for auth token is required! If there is nothing in the conf
+        // file, then an environment variable must be set.
+        env::set_var(KRILL_ENV_AUTH_TOKEN, "secret");
+
+        let c = Config::read_config("./defaults/krill-testbed.conf").unwrap();
+        let expected_socket_addr: SocketAddr = ([127, 0, 0, 1], 3000).into();
```
ximon18, line 993: Why test the socket addr here? It's presumably tested elsewhere as it is part of the main Krill config and not part of the testbed config?

ximon18 (approved):

src/daemon/config.rs
timbru: quite right! removed this..

defaults/krill-testbed.conf
timbru: rephrased

src/daemon/config.rs
ximon18: Ah it depends on the publisher URI which it doesn't know until it is configured with the command line tools.

PR #444, opened 2021-03-21
Enable the multi-user feature by default.

ximon18:


.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         rust: [stable, 1.45.0, beta]
-        args: ["--features functional-tests,ui-tests", "--all-features"] 
+        args: ["--features functional-tests,ui-tests", "--all-features", "--no-default-features"]
```
partim, line 46: Both `--all-features` and `--no-default-features`? Isn’t the former making the latter redundant?

partim (approved):

.github/workflows/ci.yml
ximon18: It’s supposed to be testing all and none as separate permutations of the matrix but maybe that’s the wrong syntax, can’t see easily on my phone, will check later.

ximon18: @partim : I think this is correct. Each entry in the array contributes another permutattion to the matrix. If multiple "args" were wanted to be passed to Cargo that would be done within a single string element of hte array, e.g. "--all-features --no-default-features", but here we have "--all-features" in one array element and "--no-default-features" in another array element. For example the [GitHub Actions documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) gives the example of testing against multiple node versions like so: `node: [6, 8, 10]`. If we look at the jobs that were actually run for this configuration we see separate jobs for "--no-default-features" and for "--all-features" so I think it is correct:

image

ximon18: One more thing to note here is that `"--features functional-tests,ui-tests"` is actually equivalent to `""`. This is because the enabled features only enable additional types of test, they don't change the functionality that is compiled into Krill. I'll add a comment about this to the workflow.

PR #458, opened 2021-03-25
Retry removing files from old repositories X(==5) times before giving up

timbru:


timbru:
I will merge this to the dev branch, so that I can (more easily) get it into the issue-440 branch and include it there with other changes that touch a lot of the same code. So it can be reviewed more when I do the pull request for that branch.

Issue #460, opened 2021-03-27
Limit max nr of deltas


timbru:
Limit the maximum number of deltas in a notification file to a configurable hard upper limit.

While we are at it, perhaps also have a hard upper limit on the time to keep deltas for. Right now we have a setting to keep deltas at least X (default 2 hours) time. We could also have a measure that says, at most Y time..

We need some measurements but... downloading 200 deltas for RPs that haven't synced in a while seems cumbersome.

PR #462, opened 2021-03-31
Limit the number of deltas in RRDP notification.xml file. (#460)

timbru:


defaults/krill-pubd.conf
```diff
@@ -122,6 +122,59 @@
...
+# this because:
+#  1) big notification files are expensive to serve and parse
+#  2) it's not just size, because of latency parsing many small deltas may
+#     be more time consuming than falling back to a delta
+#
+# So, the objective would be to optimize things such that the notification.xml files
+# remain relatively small, while still allowing the vast majority of Relying Party
+# implementations to use deltas, under normal operations.
+#
+# We have use the following strategy to prune excessive deltas:
```
ximon18, line 148: Remove "have" ?

defaults/krill-pubd.conf
```diff
@@ -122,6 +122,59 @@
...
+#  - always keep 'retention_delta_files_min_nr' files
+#  - always keep 'retention_delta_files_min_seconds' files
+#  - beyond this:
+#    - never keep more than 'retention_delta_files_max_nr'
+#    - never keep older than 'retention_delta_files_max_seconds'
+#    - keep the others
+#
+# With these defaults:
+# retention_delta_files_min_nr = 5
+# dflt_retention_delta_files_min_seconds = 1200 (20 mins: twice the interval of commonly used RPs)
```
ximon18, line 160: Why do these settings have `dflt_` in the name? If you edit them does it change only a default but have no impact, or does it actually change behaviour and thus they should not be called defaults? Or are these just comments and the settings you can change are defined somewhere else?

defaults/krill-pubd.conf
```diff
@@ -122,6 +122,59 @@
...
+# dflt_retention_delta_files_max_seconds = 7200 (2 hours: twice the interval of slowest RPs)
+# dflt_retention_delta_files_max_nr = 50
+
+# Time to keep any files still referenced by notification files updated up to X seconds ago. We
+# should not delete these files too eagerly or we would risk that RPs with an old notification file
+# try to retrieve them, without success.
+#
+# retention_old_notification_files_seconds = 600 (10 mins)
```
ximon18, line 168: Does the value format of this setting support "(10 mins)" - I assume not. I prefer it when commented out settings just work if the leading comment character is removed.

src/daemon/config.rs
```diff
@@ -303,40 +303,57 @@ pub struct IssuanceTimingConfig {
 pub struct RepositoryRetentionConfig {
     #[serde(default = "RepositoryRetentionConfig::dflt_retention_old_notification_files_seconds")]
     pub retention_old_notification_files_seconds: i64,
-    #[serde(default = "RepositoryRetentionConfig::dflt_retention_delta_files_seconds")]
-    pub retention_delta_files_seconds: i64,
     #[serde(default = "RepositoryRetentionConfig::dflt_retention_delta_files_min_nr")]
     pub retention_delta_files_min_nr: usize,
```
ximon18, line 307: These settings don't seem to be documented in `krill-pubd.conf`. They are only present with the `dflt_` prefix, not without it.

src/daemon/config.rs
```diff
@@ -508,8 +525,10 @@ impl Config {
 
         let repository_retention = RepositoryRetentionConfig {
             retention_old_notification_files_seconds: 1,
-            retention_delta_files_seconds: 1,
+            retention_delta_files_min_seconds: 0,
             retention_delta_files_min_nr: 5,
+            retention_delta_files_max_seconds: 1,
```
ximon18, line 530: max seconds 1? Is this correct?

src/pubd/repository.rs
```diff
@@ -556,25 +556,41 @@ impl RrdpServer {
...
+        let max_nr = config.retention_delta_files_max_nr;
+        let max_secs = config.retention_delta_files_max_seconds;
 
         for delta in &self.deltas {
             size += delta.elements().size();
-            if size > snapshot_size || (keep > minimum && delta.older_than_seconds(retain_secs)) {
-                break;
+
+            if size > snapshot_size {
+                // never keep more than the size of the snapshot
```
ximon18, line 582: Given that size will only keep increasing as you iterate over the loop, maybe break out here?

src/pubd/repository.rs
```diff
@@ -556,25 +556,41 @@ impl RrdpServer {
...
+            } else if keep < min_nr || delta.younger_than_seconds(min_secs) {
+                // always keep 'retention_delta_files_min_nr' files
+                // always keep 'retention_delta_files_min_seconds' file
+                keep += 1
+            } else if keep == max_nr || delta.older_than_seconds(max_secs) {
```
ximon18, line 587: Is self.deltas ordered by age? If so you break out of the loop here. If not, what happens if one delta is older sthan max seconds so is skipped but we go round the loop again and find the next delta is _younger than max seconds and so is kept, but now keep has incremented by 1 referring to keeping the last delta, not this one??? This looks correct but feels like there is potential for errors here.

defaults/krill-pubd.conf
timbru: ack

defaults/krill-pubd.conf
timbru: removed dflt_

defaults/krill-pubd.conf
timbru: indeed, removed

src/daemon/config.rs
timbru: removed dflt_

src/pubd/repository.rs
timbru: yep, now breaking

src/pubd/repository.rs
timbru: yep, now breaking

src/daemon/config.rs
timbru: this is test config, so yes

ximon18: Ah, oops :-)

ximon18 (approved):

PR #463, opened 2021-04-01
Apply dependabot fixes to lagosta

ximon18:


timbru:
I think we can just merge this. If we need to review lagosta changes then we should probably review them in that project. The diff of the resulting huge .js files is impossible to review.

PR #474, opened 2021-04-13
Push password field down so username error does not overlap.

ximon18:
See: https://github.com/NLnetLabs/lagosta/pull/97.


timbru (approved):
Can't really judge from the compiled lagosta files. But I think it's probably good to merge to dev.. we will double check the UI before the release.

ximon18:
Please see the [linked Lagosta PR](https://github.com/NLnetLabs/lagosta/pull/97). I'll wait a bit to see if @enricoberti or anyone else comments on that.

ximon18:
The E2E test failure is not related to this change. I will merge this as it is based on latest Lagosta master and so doesn't accidentally merge anything out.

Issue #366, opened 2021-01-06
Upgrade dependencies for 0.9 (tokio, reqwest, rpki-rs, bcder..)


timbru:
- [x] run `cargo audit`
- [x] minimum rust 1.44 see NLnetLabs/routinator#444
- [x] build with min. rust version

ximon18:
See also #447.

ximon18:
@timbru: Regarding `minimum rust 1.44` the Krill minimum is already 1.45.

PR #476, opened 2021-04-13
Ran cargo update (#366)

timbru:
Checked versions to be updated and ran `cargo update`. This leaves only 4 warnings for `cargo audit`.

We should plan to upgrade hyper, reqwest, tokio, bytes, rpki-rs and bcder in a future release. But for the moment there are no pressing security or functional reasons to do so and this is a significant amount of work (e.g. the inbound HTTPS connections). So, in short - I believe that this is enough for 0.9.0-rc1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1492,9 +1492,9 @@ dependencies = [
 
 [[package]]
 name = "oso"
-version = "0.11.3"
+version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


timbru:
Before approving.. we need to look at the minimum rust version as mentioned in issue #366

timbru:
There are newer versions of fern and syslog, but looking at the documentation and changelogs there does not seem to be an urgent reason to upgrade - and upgrading will involve some code changes. So, I prefer to stick with what we have for now and merge this,

PR #491, opened 2021-04-20
RFC 8439 compliant nonce generation for login session encryption.

ximon18:
This PR improves the nonce generation to use a rising nonce counter prefixed with a random value generated on Krill startup, so it will differ per Krill invocation and per Krill instance (e.g. if we ever run two Krill instances in a cluster they will not generate the same none sequences). It does not persist the last used nonce to disk but due to the use of a random prefix per invocation of Krill this is better than using a nonce that just rises from zero every time. The PR also cleans up some naming.


ximon18:
I'll fix up the PR to be mergable and mark it ready for review when that's done.

timbru:
I find it difficult to judge. Even it it's not perfect I think this is an improvement. Nothing jumps at me from looking at the code. There are tests. And it works in my browsers (safari and firefox). So approving..

timbru (approved):

ximon18:
>I find it difficult to judge. Even it it's not perfect I think this is an improvement. Nothing jumps at me from looking at the code. There are tests. And it works in my browsers (safari and firefox). So approving..

I feel the same way about it, it's better to have this I think than not. Will merge.

Issue #495, opened 2021-04-21
Report *which* file/dir has an issue with I/O errors


timbru:
Unfortunately Rust's `io::Error` does not report *which* file or directory has an issue when an error occurs (or at least this seems to be the case very often).

This is unhelpful when debugging issues.

So, find all uses of io::Error and return something which does include the path that had the issue.

PR #506, opened 2021-04-29
Use more descriptive KrillIoError where feasible (#495)

timbru:


src/commons/util/file.rs
```diff
@@ -46,81 +55,95 @@ pub fn file_path(base_path: &Path, file_name: &str) -> PathBuf {
...
+            .map_err(|e| KrillIoError::new(format!("Could not remove file: {}", path.to_string_lossy()), e))?;
 
         let mut parent_opt = path.parent();
 
         while parent_opt.is_some() {
             let parent = parent_opt.unwrap();
-            if parent.read_dir()?.count() == 0 {
+            if parent
+                .read_dir()
+                .map_err(|e| KrillIoError::new(format!("Could not read directory: '{}'", parent.to_string_lossy()), e))?
```
ximon18, line 140: The same error message is used in a different function above, will it be possible to tell which of the two cases occurred or should more context be included in these messages?

src/commons/util/file.rs
```diff
@@ -46,81 +55,95 @@ pub fn file_path(base_path: &Path, file_name: &str) -> PathBuf {
...
 }
 
-/// Loads a files and deserializes as json for the expected type. Maps json
-/// errors to io::Error
-pub fn load_json<O: DeserializeOwned>(full_path: &Path) -> Result<O, io::Error> {
+/// Loads a files and deserializes as json for the expected type. Maps json errors to KrillIoError
+pub fn load_json<O: DeserializeOwned>(full_path: &Path) -> Result<O, KrillIoError> {
     let bytes = read(full_path)?;
-    serde_json::from_slice(&bytes).map_err(|_| io::Error::new(io::ErrorKind::Other, "could not deserialize json"))
+    serde_json::from_slice(&bytes).map_err(|_| {
```
ximon18, line 76: Did you intend to ignore the actual error object here rather than include it in the returned error object?

src/commons/util/file.rs
```diff
@@ -46,81 +55,95 @@ pub fn file_path(base_path: &Path, file_name: &str) -> PathBuf {
...
     let path = path_with_rsync(base_path, uri);
     save(content, &path)
 }
 
 /// Reads a file to Bytes
-pub fn read(path: &Path) -> Result<Bytes, io::Error> {
-    let mut f = File::open(path).map_err(|_| Error::cannot_read(path))?;
+pub fn read(path: &Path) -> Result<Bytes, KrillIoError> {
+    let mut f =
+        File::open(path).map_err(|e| KrillIoError::new(format!("Could not read: '{}'", path.to_string_lossy()), e))?;
```
ximon18, line 101: Will `e` be sufficient to distinguish the failure to open from the failure to read, or should this message be different to that a few lines below?

src/commons/util/file.rs
```diff
@@ -46,81 +55,95 @@ pub fn file_path(base_path: &Path, file_name: &str) -> PathBuf {
...
+    fs::remove_file(full_path)
+        .map_err(|e| KrillIoError::new(format!("Could not remove file: {}", full_path.to_string_lossy()), e))
 }
 
-pub fn clean_file_and_path(path: &Path) -> Result<(), io::Error> {
+pub fn clean_file_and_path(path: &Path) -> Result<(), KrillIoError> {
     if path.exists() {
-        fs::remove_file(&path)?;
+        fs::remove_file(&path)
+            .map_err(|e| KrillIoError::new(format!("Could not remove file: {}", path.to_string_lossy()), e))?;
```
ximon18, line 132: The error message here is the same as in `fn delete()` above, will it be possible to distinguish these two cases from each other in the logs?

src/commons/util/softsigner.rs
```diff
@@ -27,12 +29,25 @@ pub struct OpenSslSigner {
 
 impl OpenSslSigner {
     pub fn build(work_dir: &Path) -> Result<Self, SignerError> {
-        let meta_data = fs::metadata(&work_dir)?;
+        let meta_data = fs::metadata(&work_dir).map_err(|e| {
+            KrillIoError::new(
+                format!("Could not get meta data from '{}'", work_dir.to_string_lossy()),
```
ximon18, line 34: "meta data" -> "metadata" ?

src/daemon/auth/common/crypt.rs
```diff
@@ -35,8 +35,14 @@ pub(crate) fn load_or_create_key(key_path: &Path) -> KrillResult<Vec<u8>> {
         openssl::rand::rand_bytes(&mut key_bytes)
             .map_err(|err| Error::Custom(format!("Unable to generate symmetric key: {}", err)))?;
 
-        let mut f = File::create(key_path)?;
-        f.write_all(&key_bytes)?;
+        let mut f = File::create(key_path)
+            .map_err(|e| KrillIoError::new(format!("Could not create key file '{}'", key_path.to_string_lossy()), e))?;
```
ximon18, line 39: The same error message is returned by `fn load_key()` above, will it be possible to distinguish the two cases from each other in the logs?

src/daemon/config.rs
```diff
@@ -724,8 +727,10 @@ impl Config {
 
     pub fn read_config(file: &str) -> Result<Self, ConfigError> {
         let mut v = Vec::new();
-        let mut f = File::open(file)?;
-        f.read_to_end(&mut v)?;
+        let mut f =
+            File::open(file).map_err(|e| KrillIoError::new(format!("Could not read config file '{}'", file), e))?;
+        f.read_to_end(&mut v)
+            .map_err(|e| KrillIoError::new(format!("Could not read config file '{}'", file), e))?;
```
ximon18, line 733: Same comment as above, will it be possible to distinguish the open and read cases from `e` alone?

src/pubd/repository.rs
```diff
@@ -395,13 +403,39 @@ impl RsyncdStore {
...
+                    e,
+                )
+            })?;
         }
 
-        fs::rename(&new_dir, &current_dir)?;
+        fs::rename(&new_dir, &current_dir).map_err(|e| {
+            KrillIoError::new(
+                format!(
+                    "Could not rename rsync dir from '{}' to '{}' while publishing",
```
ximon18, line 421: This error message is the same as the one above, should something be added to make them distinct or will it be clear from the different dir names in the two cases?

src/pubd/repository.rs
```diff
@@ -395,13 +403,39 @@ impl RsyncdStore {
...
+                e,
+            )
+        })?;
 
         if old_dir.exists() {
-            fs::remove_dir_all(&old_dir)?;
+            fs::remove_dir_all(&old_dir).map_err(|e| {
+                KrillIoError::new(
+                    format!(
+                        "Could not clean up old rsync dir '{}' while publishing",
```
ximon18, line 433: Above you say remove, here you say cleanup. Cleanup to me could mean reorganize or remove clutter, it doesn't necessarily mean delete. Perhaps reword this?

src/pubd/repository.rs
```diff
@@ -653,12 +687,37 @@ impl RrdpServer {
...
+                    "Could not read RRDP base directory '{}'",
+                    self.rrdp_base_dir.to_string_lossy()
+                ),
+                e,
+            )
+        })? {
+            let entry = entry.map_err(|e| {
+                KrillIoError::new(
+                    format!(
+                        "Could not read RRDP base directory '{}'",
```
ximon18, line 715: This error message is the same as the one above. Should something be added to make them distinct?

src/pubd/repository.rs
```diff
@@ -673,8 +732,24 @@ impl RrdpServer {
...
+                    "Could not read RRDP session directory '{}'",
+                    session_dir.to_string_lossy()
+                ),
+                e,
+            )
+        })? {
+            let entry = entry.map_err(|e| {
+                KrillIoError::new(
+                    format!(
+                        "Could not read RRDP session directory '{}'",
```
ximon18, line 747: This error message is the same as the one above. Should something be added to make them distinct?

ximon18 (approved):

src/commons/util/file.rs
timbru: I don't think it matters. There is context, it lists the directory and the error. This case is also extremely unlikely. This would be failure to read a parent dir after successfully removing a file under this dir (potentially some levels deeper)

src/commons/util/file.rs
timbru: good point, adding it

src/commons/util/file.rs
timbru: Changed first from 'read' to 'open'.

src/commons/util/file.rs
timbru: done, and renamed `delete` to `delete_file` to be explicit.

src/commons/util/softsigner.rs
timbru: ack

src/daemon/auth/common/crypt.rs
timbru: I don't see what you mean. I think all errors here are subtly different. E.g. "Could not create key file '..'" vs "Could not write to key file '..'"

src/daemon/config.rs
timbru: changed first 'read' to 'open'

src/pubd/repository.rs
timbru: ack

src/pubd/repository.rs
timbru: reworded to make it more clear

src/daemon/auth/common/crypt.rs
ximon18: In `Signer::create_key()` it says:
```
       let mut f = File::create(&path)
            .map_err(|e| KrillIoError::new(format!("Could not create key file '{}'", path.to_string_lossy()), e))?;
```
That is the same message as in this fn. Will the two cases be obviously different in the logs? Is the difference in key path value enough or should the logged message be different?

src/pubd/repository.rs
timbru: ok changed this to "Could not read entry in RRDP base directory '{}'"

src/pubd/repository.rs
timbru: ack added 'entry in'

src/daemon/auth/common/crypt.rs
timbru: Right, I think it's enough to have the context of the filename and the `io::Error`.

Issue #509, opened 2021-04-30
Should re-issue and publish objects during keyroll


timbru:
We have test for this, but despite this when testing a key roll on the RC version ROAs were not re-issued. Probably because of an issue with the 'ca_objects' structure. MUST fix before release obviously.

Issue #480, opened 2021-04-15
Migrate CA repository using a key rollover


timbru:
CAs should be able to migrate from one repository to another, e.g. if a CA had to use a local repository to start with, but is now able to use a third party repository maintained by an RIR or NIR.

The most natural way to achieve this is to use a key rollover where the new key will publish in the new repository. When the old key is revoked at the end of that process the objects in the old repository can be withdrawn.

We have running code to support this that was going to be included in Krill 0.9.0. However, we found that there was a corner case that we have not been able to successfully reproduce and solve where in some cases the migration would not work. Therefore we decided to disable this in 0.9.0 and created this issue to finish and thoroughly test this function. It is our goal to include this in release 0.9.1.

PR #510, opened 2021-05-03
Re-issue ROAs and CA certificates on key activation. (#509)

timbru:


timbru:
Now also featuring a test in `functional.rs` that verifies that a ROA and certificate are in fact re-issued. Before the test was just checking key status - I can't really say why it was not checking this. I assumed that it was I guess - which is why it was missed in the refactoring.

src/daemon/ca/rc.rs
```diff
@@ -450,18 +450,42 @@ impl ResourceClass {
...
 
-        Ok(vec![self.key_state.keyroll_activate(
-            self.name.clone(),
-            self.parent_rc_name.clone(),
-            signer,
-        )?])
+                let mut cert_updates = ChildCertificateUpdates::default();
+                for issued in self.certificates.iter() {
+                    // re-issue
+                    let re_issued = self.re_issue(issued, None, new_key, None, issuance_timing, signer)?;
```
ximon18, line 476: If `self.re_issue()` succeeds on one loop iteration but fails on the next this fn will abort, but in that case the event for the update that happened will not be returned. Is this okay?

tests/migrate_repository.rs
```diff
@@ -403,6 +415,13 @@ async fn migrate_repository() {
                 .await
             );
         }
+
+        // Expect that CA3 publishes nothing in the embedded repo
```
ximon18, line 419: This comment says CA3 but the following code says CA1, is this correct?

ximon18 (approved):
The logic of the test is beyond my ability to confirm but I can see that you are indeed testing more than you were before.

src/daemon/ca/rc.rs
timbru: That is okay, none of this modifies the actual CA. This is creating a number of events which will only be applied if everything was successful - including triggered pre-save effects (create new mft and crl in particular).

tests/migrate_repository.rs
timbru: indeed copy/paste error from previous test. This will be reviewed as part of issue #480

Issue #508, opened 2021-04-30
Objects should be withdrawn when removing a parent


timbru:
Removing a parent seems to leave the objects in the 'ca_objects` structure. They should be withdrawn.

PR #511, opened 2021-05-03
Remove resource classes (and withdraw objects) when a parent is remov…

timbru:
…ed. (#508)


src/daemon/ca/certauth.rs
```diff
@@ -986,7 +986,21 @@ impl CertAuth {
...
         } else {
-            Ok(vec![CaEvtDet::parent_removed(&self.handle, self.version, parent)])
+            let mut event_details = vec![];
+
+            for (rcn, rc) in &self.resources {
+                if rc.parent_handle() == &parent {
+                    event_details.push(CaEvtDet::ResourceClassRemoved {
+                        resource_class_name: rcn.clone(),
+                        parent: parent.clone(),
+                        revoke_requests: vec![], // We will do a best effort revoke request, but not triggered through this event
```
ximon18, line 996: Does this comment agree with the that in `06_ca_manager.md`?
> Note that if the CA actively removes a parent, it will pro-actively send revocation requests for all its keys first, and we do not trigger this task.

tests/functional.rs
```diff
@@ -638,8 +638,15 @@ async fn functional() {
     info("##################################################################");
     info("");
     {
-        delete_ca(&ca4).await;
+        // Remove CA4 from CA3
+        delete_parent(&ca4, &ca3).await;
+
+        // Expect that CA4 withdraws all
+        {
+            assert!(will_publish("CA4 should withdraw objects when parent is removed", &ca4, &[]).await);
```
ximon18, line 646: Stupid question: where is it tested that the objects are withdrawn?

ximon18 (approved):

ximon18 (approved):

src/daemon/ca/certauth.rs
timbru: ah good catch, will update doc. We do, now, remove the resource classes first because this will trigger withdrawing the objects.

tests/functional.rs
timbru: No, you're right.. added another test for this - will now delete CA3 and expect it to withdraw everything, and its parent to revoke and withdraw the .cer issued to it.

Issue #514, opened 2021-05-04
When requesting a new certificate, log the *difference* in resources and/or validity time not after.


timbru:
This is very noisy now at an NIR making it impossible to read what changed.

As this does not change any actual logic, it should be okay to do this after rc2 and before the actual 0.9.0 release. As long as we are 100% sure that this does not add significant overhead, or can result in panics (should not be the case).

PR #516, opened 2021-05-10
Log resource diffs where applicable (#514)

timbru:


src/daemon/ca/keys.rs
```diff
@@ -97,26 +96,42 @@ impl CertifiedKey {
...
 
-        if not_after == new_not_after {
-            trace!("No change in not after time for certificate for key '{}'", self.key_id);
+        if until_new_not_after_millis < 0 {
+            // New not after time is in the past!
+            //
+            // This is rather odd. The parent should just exclude the resource class in the
+            // eligible entitlements instead. So, we will essentially just ignore this until
+            // they do.
+            false
```
ximon18, line 109: If this is odd, should a warning be logged?

ximon18 (approved):

src/daemon/ca/keys.rs
timbru: Yes, you're right. Pushed.

Still not logging in cases where we decide that there is no action required. No resource changes and not_after time unchanged. I could add debug / trace logging for this. What do you think?

timbru: asking the question is answering it I guess.. pushed a 'debug' message in case there are no changes in resources or not after time.

Issue #527, opened 2021-05-18
Revert using separate krillpub* binaries.


timbru:
Separate binaries were introduced in an effort to make things less confusing to end-users. I.e. don't show functions that most users do not need.

However, it seems that things only get more complicated this way. Especially w.r.t. packaging.

So, in short, revert this change before the 0.9.0 release.

- [x] Merge krillpubd
- [x] Merge krillpubc
- [x] Update developer doc
- [x] Update RTD

PR #529, opened 2021-05-19
Merge krillpub* into krill* (#527)

timbru:
well.. and one fix fo the name of the json file containing all the current repo objects.. because I happened to stumble on it again and I think it's good to have something nicer like 0.json rather than just 0 - even if it's just the Krill code looking at this file.


src/pubd/repository.rs
```diff
@@ -931,13 +931,9 @@ impl RepositoryAccessProxy {
...
-        match self.store.get_latest(&self.key) {
-            Ok(repo) => Ok(repo),
-            Err(e) => match e {
-                AggregateStoreError::UnknownAggregate(_) => Err(Error::RepositoryServerNotEnabled),
-                _ => Err(Error::AggregateStoreError(e)),
-            },
-        }
+        self.store
+            .get_latest(&self.key)
+            .map_err(|_| Error::custom("Publication Server data missing"))
```
ximon18, line 936: What would the actual error here contain and should it be logged? Should self.key be logged? Is it correct that this uses a custom error instead of typed error?

ximon18 (approved):

src/pubd/repository.rs
timbru: Yeah, you're right.. not sure why I left it out. I pushed a fix to include the original error message. This case really should never happen, unless there is some data corruption issue.

PR #536, opened 2021-06-01
RPM packaging

ximon18:
Initial workflow, Cargo.toml and supporting files. Will likely need tweaking to fix whatever breaks when run by GitHub Actions. Based heavily on RPM support recently added for Routinator and RTRTR.


ximon18:
Re-targeted at `main` to get GH Actions to run the `pkg` workflow (hopefully).

timbru:
Looks good.

Tested centos7 and centos8 RPMs, and the ubuntu focal pkg on VMs. It all seems to work fine.

timbru (approved):

ximon18:
The self-checks are failing due to the constructed package having the same version as that last released. This is because `Cargo.toml` in `main` contains version v0.9.0 which is the same as the last release. We primarily care about packages built against the `dev` branch or made for new releases so this is not a big concern. As the workflow changes won't be available to future dev branch or release builds unless they are first merged into `main` I am going to disregard this failure and merge the PR.

PR #585, opened 2021-06-22
Update krill.conf

racompton:
Added info that the resource entitlements update is done via the up-down protocol


timbru (approved):
sure.. why not.

PR #586, opened 2021-06-25
Update dependencies: rpki-rs, and therefore bytes, tokio, hyper, reqwest

timbru:


Cargo.toml
```diff
@@ -19,16 +19,16 @@ build = "build.rs"
 [dependencies]
 base64                = "^0.13"
 basic-cookies         = { version = "^0.1", optional = true }
-bcder                 = "0.5.0"
-bytes                 = "^0.5"
+bcder                 = "0.6.0"
```
ximon18, line 22: Even though Cargo will treat "0.6.0" as "^0.6.0", perhaps it is good to insert the caret explicitly for consistency with the other version values used in Cargo.toml.

timbru: pushed

ximon18 (approved):

timbru:
Checks are failing due to slow UI tests again. Merging anyhow..

Issue #522, opened 2021-05-14
The default for 'service_uri' should pick up host and/or port if those values are set


timbru:
The default is https://localhost:3000/ but it should pick up on a non-default host or port if those values are set.

Technically this probably means we should have an Option instead of a simple value and derive it from the other two if it is not set.

PR #594, opened 2021-07-05
Make 'service_uri' default pick up port, insist it's set if 'ip' is changed (#522)

timbru:


src/daemon/config.rs
```diff
@@ -655,19 +661,21 @@ impl Config {
             return Err(ConfigError::other("Port number must be >1024"));
         }
 
-        if !self.service_uri.ends_with('/') {
-            return Err(ConfigError::other("service URI must end with '/'"));
-        } else {
-            uri::Https::from_str(&self.service_uri)
-                .map_err(|_| ConfigError::Other(format!("Invalid service uri: {}", self.service_uri)))?;
+        if self.ip != ConfigDefaults::ip() && self.service_uri.is_none() {
+            return Err(ConfigError::other("Please specify a URI for 'service_uri' if you enable binding to IP addresses other than 127.0.0.1"));
```
ximon18, line 665: Maybe be nice to the user and generate service_uri if the IP is changed to save them specifying the IP twice by hand in the config file?

ximon18 (approved):

src/daemon/config.rs
timbru: changed - will now use the ip address if it is specified, to generate the service_uri if it's not overridden in the config.

Issue #521, opened 2021-05-13
Support semantic versioning for keystore versions


timbru:
We currently have an enum to track all keystore versions of Krill:

```
pub enum KeyStoreVersion {
    Pre0_6,
    V0_6,
    V0_7,
    V0_8_0_RC1,
    V0_8,
    V0_8_1_RC1,
    V0_8_1,
    V0_8_2,
    V0_9_0_RC1,
}
```

This is a bit error prone when it comes to releases. It would be better to use `const::KRILL_VERSION` for this, and implement parsing of ealier (semantic) versions.

PR #596, opened 2021-07-06
Use KrillVersion for keystore based on the Cargo.toml version. (#521)

timbru:


src/commons/util/mod.rs
```diff
@@ -1,16 +1,230 @@
...
+//------------ KrillVersion --------------------------------------------------
+
+/// Defines a Krill version. Will 
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct KrillVersion {
+    major: u64,
+    minor: u64,
+    patch: u64,
+    release: KrillVersionRelease
```
ximon18, line 31: s/KrillVersionRelease/ReleaseType/ ? (version threw me off as we already express the version here in terms of major, minor, patch).

src/commons/util/mod.rs
```diff
@@ -1,16 +1,230 @@
...
+                KrillVersionRelease::Release => Ordering::Equal,
+                _ => Ordering::Greater
+            },
+            KrillVersionRelease::Candidate(nr) => match other {
+                KrillVersionRelease::Release => Ordering::Less,
+                KrillVersionRelease::Candidate(nr_other) => nr.cmp(nr_other),
+                &KrillVersionRelease::Dev(_) => Ordering::Greater
+            },
+            KrillVersionRelease::Dev(text) => match other {
+                KrillVersionRelease::Dev(text_other) => text.cmp(text_other),
```
ximon18, line 203: Maybe dev should always equal dev?

ximon18 (approved):

src/commons/util/mod.rs
timbru: yes, changed now

src/commons/util/mod.rs
timbru: Renamed to KrillVersionReleaseType. A bit wordy but okay and descriptive methinks.

Issue #606, opened 2021-07-14
Krill 0.9.0 publication server locks on requests


timbru:
It was found that the Krill 0.9.0 publication server locks on requests.

This issue seems to be related to unexpectedly slow deserialisation of the repository content state. This issue primarily affects large repositories: more content makes this deserialisation slower and it is triggered more often because the server is frequently contacted by publishers.

Furthermore it seems that, while we do need to have locking in place around updates to the repository, there is an unnecessary lock around serving the latest RRDP files from disk while other work is being done.

Users are advised not to upgrade to 0.9.0 until we have resolved this issue. Clearly, we will work on this with the highest priority and plan to have a bug fix release as soon as we have a tested fix.

timbru:
Fixed long ago (Krill 0.9.1), issue was not closed

PR #607, opened 2021-07-15
Release 0.9.1-rc1

timbru:
This release candidate contains a quick fix for issue #606.

Because we wanted to keep the changes to a minimum the choses solution was two-fold:
1) Remove a superfluous RwLock around KrillServer (it has its own internal locks).
2) Cache the repository content rather than deserializing it all the time.

The latter works for single node setup that we still require.


src/daemon/http/server.rs
```diff
@@ -336,7 +334,7 @@ pub async fn health(req: Request) -> RoutingResult {
 pub async fn metrics(req: Request) -> RoutingResult {
     if req.is_get() && req.path().segment().starts_with("metrics") {
         let server = req.state();
-        let server = server.read().await;
+        let server = server;
```
partim, line 337: Did you keep these assignments on purpose?

src/daemon/http/server.rs
```diff
@@ -597,7 +595,7 @@ pub async fn rfc8181(req: Request) -> RoutingResult {
             Err(e) => return render_error(e),
         };
 
-        let read = state.read().await;
+        let read = state;
```
partim, line 598: Same here; just ditch `read` and keep using `state` below?

src/daemon/krillserver.rs
```diff
@@ -604,7 +604,7 @@ impl KrillServer {
         self.ca_manager.get_ca(handle).await.map(|ca| ca.publisher_request())
     }
 
-    pub async fn ca_init(&mut self, init: CertAuthInit) -> KrillEmptyResult {
+    pub async fn ca_init(&self, init: CertAuthInit) -> KrillEmptyResult {
```
partim, line 607: I don’t think this method needs to be async?

src/pubd/repository.rs
```diff
@@ -151,22 +167,50 @@ impl RepositoryContentProxy {
...
     fn write<F: FnOnce(&mut RepositoryContent) -> KrillResult<()>>(&self, op: F) -> KrillResult<()> {
+        // If there is any existing content, then we can assume that the cache
+        // has it - because it's initialized when we read the content during
+        // initialization.
         let store = self.store.write().unwrap();
-        let mut content: RepositoryContent = store.get(&self.key)?.ok_or(Error::RepositoryServerNotInitialized)?;
+        let mut cache = self.cache.write().unwrap();
+
+        let content_arc: &mut Arc<RepositoryContent> = cache.as_mut().ok_or(Error::RepositoryServerNotInitialized)?;
+        let content = Arc::<RepositoryContent>::get_mut(content_arc).unwrap();
```
partim, line 179: I think this `unwrap` will panic under certain conditions. `Arc::get_mut` only succeeds if there is exactly one clone out there. Yet you clone the arc in `read_content_cache`. So if any of the methods calling `write` are called while the clone returned by `read_content_cache` is still alive, you’ll get a panic.

I think you don’t need the arc here at all? It looks like all the methods that call `read_content` only call a method on the content anyway, so this could probably just return the `RwLockReadGuard` returned by the lock.

partim (changes requested):

src/daemon/http/server.rs
timbru: nope.. now removed. (to quick to search and remove unlocking the now removed RwLock)

src/daemon/http/server.rs
timbru: indeed.. now removed. (to quick to search and remove unlocking the now removed RwLock)

src/daemon/krillserver.rs
timbru: indeed, removed

src/pubd/repository.rs
timbru: Good point. I removed the `Arc`. And then I changed things around a bit and added `read` instead of `read_content` which takes a closure which it can execute within a read 'transaction' without the need to clone.

partim (approved):

PR #609, opened 2021-07-17
Rel 0.9.1 rc2 create snapshot on migrate

timbru:


timbru:
Merging this one without waiting for a review in the weekend - there are no dangerous changes here :) just better logging and saving a snapshot after migration to avoid having to replay events again after a successful migration.

PR #610, opened 2021-07-19
Deal with parent errors or not-performed responses to sign requests.

timbru:
Note: it's very hard to test this in an automated setup - I am not sure how to create these corner cases for testing.

So this needs a good review of the code, but especially of the intent which I will describe here:

Generally speaking we just log any errors or not-performed responses from the parent and we will try synchronise a CA with its parent again in a short while. I.e. this is what we do if we have no idea how to deal with the issue on the CA side. We think it's the parent.

Exception to this is a few not-performed responses where the parent tells the CA that it believes it is wrong to request a certificate. See here for all possible error codes: https://datatracker.ietf.org/doc/html/rfc6492#section-3.6

The CA will drop a resource class under a parent in case it gets one of the following responses:
         1201         request - no such resource class
         1202         request - no resources allocated in resource class
         1204         request - already used key in request

The first two are most likely to be caused by a race condition where the parent told the CA it had an entitlement moments before removing it. In that case the CA may request a certificate and then find it's no longer eligible.

The last case indicates an odd disagreement between the parent and child. Krill CAs will never re-use keys across RCs so this seems very odd indeed. Still.. dropping the RC seems to be the only thing that the child can do to recover from this.

Note: if an RC is dropped by the child, but it's still entitled, then it will simply create a new RC with a new key when it synchronizes again - and then issue ROAs and delegated certificates under it when it receives a certificate. In other words: it should self-heal.


ximon18:
Regarding:

> Note: it's very hard to test this in an automated setup - I am not sure how to create these corner cases for testing.

Could this be done using a mock parent that can configurably fail as desired, similar to what is done with the OpenID Connect Mock?

src/commons/api/history.rs
```diff
@@ -537,6 +538,11 @@ impl WithStorableDetails for StorableCaCommand {
             } => CommandSummary::new("cmd-ca-rcn-receive", &self)
                 .with_rcn(resource_class_name)
                 .with_resources(resources),
+            StorableCaCommand::DropResourceClass {
+                resource_class_name,
+                ..
```
ximon18, line 543: Is ignoring the drop reason here deliberate?

src/daemon/ca/certauth.rs
```diff
@@ -1272,6 +1277,32 @@ impl CertAuth {
 
         Ok(res)
     }
+
+    /// Drop a resource class because it no longer works on this parent for the specified
```
ximon18, line 1281: "works on this parent" - is "on" the correct term, say rather than "with" or "under" or ??

src/daemon/ca/certauth.rs
```diff
@@ -1248,7 +1253,7 @@ impl CertAuth {
     ///
     /// This will also generate appropriate events for changes affecting
     /// issued ROAs and delegated certificates - if because resources were
-    //  lost and ROAs/Certs would be become invalid.
+    ///  lost and ROAs/Certs would be become invalid.
```
ximon18, line 1256: Errant leading whitespace.

src/daemon/ca/manager.rs
```diff
@@ -887,92 +904,236 @@ impl CaManager {
...
+                                            parent.clone(),
+                                            rcn.clone(),
+                                            reason,
+                                        ));
+                                        break;
+                                    }
+                                },
+                                rfc6492::Res::NotPerformed(not_performed) => {
+                                    match not_performed.status() {
+                                        1201 | 1202 => {
```
ximon18, line 1000: Note what 1201 and 1202 mean here, like was done above for 1301 and 1302?

src/daemon/ca/manager.rs
```diff
@@ -887,92 +904,236 @@ impl CaManager {
...
+                                        // cannot fail unless there are bigger issues like this being the wrong
+                                        // response for this resource class. This would be extremely odd because
+                                        // we only just asked the resource class which request to send. Still, in
+                                        // order to handle this the most graceful way we can, we should just drop
+                                        // this resource class and report an error. If there are are still resource
+                                        // entitlements under the parent for this resource class, then a new class
+                                        // will be automatically created when we synchronize the entitlements again.
+
+                                        let reason = format!("received certificate cannot be added, error: {}", e);
+
```
ximon18, line 977: Rather than lots of similar/identical code blocks and comments to push an error and break, maybe factor that out?

ximon18 (approved):
Aside from some minor nits it seems fine within my limited knowledge of RPKI, i.e. it looks similar to what you describe in the PR summary but I can't really say if it does everything it should.

src/commons/api/history.rs
timbru: Yes, well kind of, the reason is a string rather than a type and cannot be translated. We include a map of key-values to help translations. That said.. losing it altogether is not the best choice. So including it now.

src/daemon/ca/certauth.rs
timbru: fixed

src/daemon/ca/certauth.rs
timbru: fixed

src/daemon/ca/certauth.rs
```diff
@@ -1272,6 +1277,32 @@ impl CertAuth {
 
         Ok(res)
     }
+
+    /// Drop a resource class because it no longer works on this parent for the specified
+    /// reason. Note that this will generate revocation requests for the current keys which
```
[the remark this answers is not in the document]
timbru: fixed

timbru:
> Regarding:
> 
> > Note: it's very hard to test this in an automated setup - I am not sure how to create these corner cases for testing.
> 
> Could this be done using a mock parent that can configurably fail as desired, similar to what is done with the OpenID Connect Mock?

Not really.. this goes through too many stacks.

src/daemon/ca/manager.rs
timbru: There are subtle differences and if you look at the actual code it's only 2 or 3 calls per option. The remainder is comments. I find it more readable like this despite the duplication.

src/daemon/ca/manager.rs
timbru: These error codes are defined in RFC 6492. The comments explain what the server told us in more understandable language. An enum could have been used when defining ``NotPerformed`` but we didn't do this at the time. I do not see a strong case for doing it now though because these values will not change.

ximon18: Okay, but anyone reading this is going to have to know what 1201 means and for that they'll have to look at the spec, isn't it easier for future maintainers just to mention what 1201 means like you did above for 1301 and 1302? Or am I missing something?

timbru:
> Aside from some minor nits it seems fine within my limited knowledge of RPKI, i.e. it looks similar to what you describe in the PR summary but I can't really say if it does everything it should.

I think this is the best we can do until we get replies from all CA implementers on the question I posted to the sidrops mailing list: https://mailarchive.ietf.org/arch/msg/sidrops/YdIZszf_h1rh3Jmb2yC4ONcMrjw/

If we know for sure that all CAs always include all current certificates for each key, then we can make further improvements here. I suggest to make a new issue for this at the time because it may take a while.

src/daemon/ca/manager.rs
timbru: Sure, the text below already explained the meaning but I have added an explicit mention of each error code. (Note and please ignore that the code was also reformatted as a result of saving it in my IDE)

Issue #601, opened 2021-07-09
Parent addition should trigger ROA re-issuance


timbru:
This affects users who created ROA (configs) under a parent, then remove the parent, and add it again.

Temporary work-around: create an additional ROA and then remove it. This triggers republication of all ROAs.

PR #611, opened 2021-07-20
Create ROAs if needed, when keys are activated (#601)

timbru:
I.e. ROAs will be re-created when a parent is removed and re-added.

### DIFF
--- a/tests/functional.rs
+++ b/tests/functional.rs
@@ -517,6 +517,21 @@ async fn functional() {
         assert!(will_publish(test_msg, &ca4, &expected_files).await);
     }
 
+    // short hand to expect ROAs under CA4, re-added when parent comes back
+    // i.e. it now has RC 2 and 3, but no more 0 and 1
+    async fn expect_roas_for_ca4_re_added(test_msg: &str, roas: &[RoaDefinition]) {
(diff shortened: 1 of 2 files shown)


tests/functional.rs
```diff
@@ -517,6 +517,21 @@ async fn functional() {
         assert!(will_publish(test_msg, &ca4, &expected_files).await);
     }
 
+    // short hand to expect ROAs under CA4, re-added when parent comes back
+    // i.e. it now has RC 2 and 3, but no more 0 and 1
```
ximon18, line 521: This comment says "no more 0 and 1" but the caller passes in `roas` with value `&[route_rc0_1, route_rc1_1]` (i.e. referencing a 0 and a 1). Is this correct?

ximon18 (approved):

tests/functional.rs
timbru: oh right, well.. I think the rc0 and rc1 in the route definition are slightly misleading. They are under 10_0_0_0 and 10_1_0_0 respectively which happen to map to those RCs (until they become 2 and 3 that is). Changed the names to make this less confusing.

Issue #613, opened 2021-07-28
Patch workflow for cargo-generate-rpm issue #21


ximon18:
Issue https://github.com/cat-in-136/cargo-generate-rpm/issues/21 caused Routinator RPM packaging to fail. We worked around this temporarily in the Routinator project with PR https://github.com/NLnetLabs/routinator/pull/629. If the issue isn't fixed upstream in the cargo-generate-rpm project we will need to apply the same workaround here too otherwise the Krill "Packaging" workflow will fail the next time it runs.

PR #617, opened 2021-07-29
Patch workflow for cargo-generate-rpm issue #21 (#613)

ximon18:
Temporary workaround for https://github.com/cat-in-136/cargo-generate-rpm/issues/21.

### DIFF
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -170,7 +170,9 @@ jobs:
       run: |
         case ${OS_NAME} in
           centos)
-            cargo install cargo-generate-rpm --version ${CARGO_GENERATE_RPM_VER}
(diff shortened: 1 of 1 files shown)


ximon18:
The pkg workflow failure `E: Packages were downgraded and -y was used without --allow-downgrades` is nothing to do with this PR, it's because `Cargo.toml` in the `main` branch has the same version as the last DEB package published to packages.nlnetlabs.nl: 0.9.1.

ximon18:
Note: the `cargo-generate-rpm` project has a fix for this issue but is unable to publish the fix to crates.io as it depends on an unpublished version of the `rpm-rs` crate. See: https://github.com/cat-in-136/cargo-generate-rpm/issues/24

timbru (approved):
Sorry for the delay! Sure.. works for me for now.

Issue #612, opened 2021-07-27
Remove unexpected subdir created by GH Actions packaging workflow artifact upload


ximon18:
When the GH Actions packaging workflow was extended to support RPM packages the upload artifact command was extended with a second wildcard path, `*.rpm` in addition to `*.deb`. However this causes the parent directory that contains the package file to be uploaded to be included in the created zip archive. The packages.nlnetlabs.nl server-side script doesn't expect this subdirectory to exist so ideally we would still just put the `.rpm` or .`deb` file into the root of the created zip archive. For now the server-side script has been modified to handle this unexpected case but it's a bit of a quick hack and it would be good to clean this up, preferably by just behaving as we did before, i.e. no subdirs inside the created zip files.

PR #618, opened 2021-07-29
Remove unexpected subdir created by GH Actions packaging workflow artifact upload (#612)

ximon18:
See #612.


ximon18:
Contains the changes in PR #617 (otherwise it fails on `cargo install cargo-generate-rpm`).

ximon18:
Won't fix, our package server scripts handle this case just fine which is good enough for now.

PR #619, opened 2021-07-29
Mark roas lost resources

timbru:


src/commons/bgp/analyser.rs
```diff
@@ -62,23 +62,44 @@ impl BgpAnalyser {
         }
     }
 
-    pub async fn analyse(&self, roas: &[RoaDefinition], scope: &ResourceSet) -> BgpAnalysisReport {
+    pub async fn analyse(
+        &self,
+        roas: &[RoaDefinition],
+        resources_held: &ResourceSet,
+        limit: Option<ResourceSet>,
```
ximon18, line 69: Question: Why is `limit` a better name here than `scope`? I find the concept of filtering on a "limit" that "contains" something confusing.

ximon18:
Note: https://github.com/NLnetLabs/lagosta/pull/118 is part of this work.

ximon18 (approved):
Code here and in the associated Lagosta PR looks good, but I have not properly tested the Lagosta changes.

src/commons/bgp/analyser.rs
ximon18: Okay. A limit to my mind is a point part way through some spectrum, but in this case it is a set of resources. Would `limited_scope` make sense? I see what you are saying so I suppose this is bikeshedding thus if you're fine with it we can also just leave it as is.

timbru: Sure, if it makes this more clear to you and other readers I am all for it. Pushed.

Issue #620, opened 2021-07-30
Monitor Krill versions of children so that they may be approached in case of issues.


timbru:
Let Krill CAs send a user-agent to parents or publication servers so that they can be monitored. Add this to the stats for the Krill parent/publication server side.

This is extremely useful in case there are known critical issues with a Krill version. This way the parent CA or publication server can reach out to their customers/clients/members and alert them.

Since there is a trust relationship between CAs and parents and publication servers this should not be a concern. The user-agent should only be sent here - and not for example when retrieving the RIS dump files.

timbru:
Edit: do not keep track of this from the publication server side. The use case for this is less clear than for parent CAs which have a more clear authority to keep track of child versions and the authority to speak to them.

Furthermore adding these stats for the publication server is significantly more involved, so I figured that this can be postponed for now. If there is a clear indication of this need in future, then we can address it then. 

Also note that the user agent is sent to the publication server - so presumably they can use a proxy and log this - if they really need to.

PR #622, opened 2021-08-04
Monitor krill versions

timbru:
This PR adds the function to monitor child CA connections. The stats are kept outside of the CA state in the existing structure that we have in order to avoid churn in the parent CA. The stats include the last connection time, result and user agent of the child CA.

Furthermore Krill CAs will now send their version in the user agent when sending RFC 6492 or 8181 protocol messages to their parents and publication server. See issue: #620

### DIFF
--- a/src/cli/options.rs
+++ b/src/cli/options.rs
@@ -9,9 +9,9 @@ use std::{env, fmt};
 use bytes::Bytes;
 use clap::{App, Arg, ArgMatches, SubCommand};
 
-use rpki::uri;
 use rpki::repository::crypto::KeyIdentifier;
 use rpki::repository::x509::Time;
+use rpki::uri;
 
 use crate::commons::crypto::{IdCert, SignSupport};
 use crate::commons::remote::rfc8183;
@@ -469,6 +469,15 @@ impl Options {
         app.subcommand(sub)
     }
 
+    fn make_cas_children_connections_sc<'a, 'b>(app: App<'a, 'b>) -> App<'a, 'b> {
+        let mut sub = SubCommand::with_name("connections").about("Show connections stats for children of a CA");
+
+        sub = Self::add_general_args(sub);
+        sub = Self::add_my_ca_arg(sub);
+
+        app.subcommand(sub)
+    }
+
     fn make_cas_children_sc<'a, 'b>(app: App<'a, 'b>) -> App<'a, 'b> {
         let mut sub = SubCommand::with_name("children").about("Manage children for a CA");
 
@@ -477,6 +486,7 @@ impl Options {
         sub = Self::make_cas_children_info_sc(sub);
         sub = Self::make_cas_children_remove_sc(sub);
         sub = Self::make_cas_children_response_sc(sub);
+        sub = Self::make_cas_children_connections_sc(sub);
 
         app.subcommand(sub)
     }
@@ -1409,6 +1419,14 @@ impl Options {
         Ok(Options::make(general_args, command))
     }
 
+    fn parse_matches_cas_children_connections(matches: &ArgMatches) -> Result<Options, Error> {
+        let general_args = GeneralArgs::from_matches(matches)?;
+        let my_ca = Self::parse_my_ca(matches)?;
+
+        let command = Command::CertAuth(CaCommand::ChildConnections(my_ca));
+        Ok(Options::make(general_args, command))
+    }
+
     fn parse_matches_cas_children(matches: &ArgMatches) -> Result<Options, Error> {
         if let Some(m) = matches.subcommand_matches("add") {
             Self::parse_matches_cas_children_add(m)
@@ -1420,6 +1438,8 @@ impl Options {
             Self::parse_matches_cas_children_update(m)
         } else if let Some(m) = matches.subcommand_matches("remove") {
             Self::parse_matches_cas_children_remove(m)
+        } else if let Some(m) = matches.subcommand_matches("connections") {
+            Self::parse_matches_cas_children_connections(m)
         } else {
             Err(Error::UnrecognizedSubCommand)
         }
@@ -2087,6 +2107,7 @@ pub enum CaCommand {
     ChildAdd(Handle, AddChildRequest),
     ChildUpdate(Handle, ChildHandle, UpdateChildRequest),
     ChildDelete(Handle, ChildHandle),
+    ChildConnections(Handle),
 
     // Key Management
     KeyRollInit(Handle),
--- a/src/commons/api/ca.rs
+++ b/src/commons/api/ca.rs
@@ -1538,6 +1538,30 @@ impl fmt::Display for ExchangeResult {
     }
 }
 
+//------------ ChildrenStats -------------------------------------------------
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct ChildrenStats {
+    children: Vec<ChildStats>,
+}
+
+impl ChildrenStats {
+    pub fn new(children: Vec<ChildStats>) -> Self {
+        ChildrenStats { children }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct ChildStats {
+    handle: ChildHandle,
+    last_exchange: Option<ChildExchange>,
+}
+
+impl ChildStats {
+    pub fn new(handle: ChildHandle, last_exchange: Option<ChildExchange>) -> Self {
+        ChildStats { handle, last_exchange }
+    }
+}
+
 //------------ ChildStatus ---------------------------------------------------
 
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
@@ -1566,6 +1590,12 @@ impl ChildStatus {
 impl Default for ChildStatus {
     fn default() -> Self {
         ChildStatus { last_exchange: None }
+    }
+}
+
+impl From<ChildStatus> for Option<ChildExchange> {
+    fn from(status: ChildStatus) -> Self {
+        status.last_exchange
     }
 }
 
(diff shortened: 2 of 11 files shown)


src/cli/options.rs
```diff
@@ -469,6 +469,15 @@ impl Options {
         app.subcommand(sub)
     }
 
+    fn make_cas_children_connections_sc<'a, 'b>(app: App<'a, 'b>) -> App<'a, 'b> {
+        let mut sub = SubCommand::with_name("connections").about("Show connections stats for children of a CA");
```
ximon18, line 473: Strictly speaking this shows stats for a single CA, right?

src/commons/api/ca.rs
```diff
@@ -1503,41 +1503,139 @@ impl ParentExchange {
...
+}
+
+//------------ ChildStatus ---------------------------------------------------
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct ChildStatus {
+    last_exchange: Option<ChildExchange>,
+}
+
+impl ChildStatus {
```
ximon18, line 1602: As the user-agent comes from the connecting child CA should we consider the user agent string to be untrusted input and store only upto some defined maximum number of characters rather than store whatever it sends to us irrespective of length?

timbru: We could choose an arbitrary max length. I am not too worried about abuse though. This only gets saved if the exchange is actually successful. But sure, let's cap it at.. what? 256 bytes?

ximon18: For Krill child CAs the UA string is quite short, but in principle to cater to any potential child CA implementation then yes I guess 256 bytes is reasonable. Looking at http://useragentstring.com/pages/useragentstring.php?name=All there are some there that are around 200 characters long for example, and I wouldn't know where to split a UA string if we only wanted to capture part of it.

src/daemon/ca/status.rs
```diff
@@ -149,10 +180,36 @@ impl StatusStore {
...
+
+        Ok(())
+    }
+
+    async fn update_ca_child_status<F>(&self, ca: &Handle, child: &ChildHandle, op: F) -> KrillResult<()>
+    where
+        F: FnOnce(&mut ChildStatus),
+    {
+        self.update_ca_status(ca, |status| {
+            if !status.children.contains_key(child) {
```
ximon18, line 205: Could this potentially suffer from a race condition whereby two thread both try to check for the presence of the key at the same time and then both try to insert it, or is that prevented by the acquisition of the `self.lock` in `update_ca_status()` ?

Perhaps it would make sense to use [`Entry::or_insert_with()`](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert_with) anyway?

src/daemon/http/server.rs
```diff
@@ -1241,6 +1250,7 @@ async fn api_ca_children(req: Request, path: &mut RequestPath, ca: Handle) -> Ro
         },
         None => match *req.method() {
             Method::POST => api_ca_add_child(req, ca).await,
+            Method::GET => api_ca_children_stats(req, ca).await,
```
ximon18, line 1253: It feels a bit strange to get statistics about the connections made by a child CA to us by fetching the "contact" for the child, this is not at all the inverse of the POST operation. Should stats perhaps have their own endpoint?

src/daemon/ca/status.rs
```diff
@@ -109,37 +120,57 @@ impl StatusStore {
...
+    ) -> KrillResult<()> {
         let error_response = Self::error_to_error_res(&error);
 
-        status.repo.set_failure(uri, error_response);
-        self.set_ca_status(ca, &status)
+        self.update_ca_child_status(ca, child, |status| status.set_failure(user_agent, error_response))
+            .await
+    }
+
+    pub async fn get_children_status(&self, ca: &Handle) -> KrillResult<ChildrenConnectionStats> {
```
ximon18, line 158: The code mixes and matches the terms status and stats which are not strictly the same thing. Perhaps review the  places where these terms are used for consistency and correctness?

src/commons/api/ca.rs
```diff
@@ -1503,41 +1503,139 @@ impl ParentExchange {
...
+}
+
+impl fmt::Display for ChildrenConnectionStats {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if !self.children.is_empty() {
+            writeln!(f, "handle, user_agent, last_exchange, result")?;
+            for child in &self.children {
+                match &child.last_exchange {
+                    None => {
+                        writeln!(f, "{}, n/a, never, n/a", child.handle)?;
```
ximon18, line 1561: This output has spaces between the commas and the values...

src/commons/api/ca.rs
```diff
@@ -1503,41 +1503,139 @@ impl ParentExchange {
...
+                    }
+                    Some(exchange) => {
+                        let agent = exchange.user_agent.as_deref().unwrap_or("");
+                        let time = Time::new(Utc.timestamp(exchange.timestamp, 0));
+
+                        writeln!(
+                            f,
+                            "{},{},{},{}",
```
ximon18, line 1569: While this output does not have spaces between the commas and the values.

src/commons/api/ca.rs
```diff
@@ -1503,41 +1503,139 @@ impl ParentExchange {
...
+    children: Vec<ChildConnectionStats>,
+}
+
+impl ChildrenConnectionStats {
+    pub fn new(children: Vec<ChildConnectionStats>) -> Self {
+        ChildrenConnectionStats { children }
+    }
+}
+
+impl fmt::Display for ChildrenConnectionStats {
```
ximon18, line 1554: I haven't been able to test this locally yet as I don't have any children of my CA but looking at the code I _think_ the response body to the API GET request for the `../{child}/contact` is this comma-separated string, is that right or is this only for display by `krillc` to command line clients? If we wanted to extend this later with more information, is this format choice a good idea? (maybe I'm just not reading this correctly...)

src/commons/api/ca.rs
```diff
@@ -1503,41 +1503,139 @@ impl ParentExchange {
...
+}
+
+impl From<ChildStatus> for Option<ChildExchange> {
+    fn from(status: ChildStatus) -> Self {
+        status.last_exchange
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct ChildExchange {
```
ximon18, line 1633: Thinking out loud... would it be useful to store both the last success and the last failure rather than just the last result? If I see that it last failed at time XXX, I don't know if it's always been failing or if it ever succeeded and if so when. I suppose that could be an enhancement later assuming the response format for the child status/stats can be extended in a backward compatible way?

ximon18:
Was it the intent of this work to include Prometheus metrics as well?

timbru:
> This work "monitors" RFC 6492 connections from some child to this server, but the description also mentions RFC 8181 - should RFC 8181 connections from some child to this server also be tracked by this work?

https://github.com/NLnetLabs/krill/issues/620#issuecomment-892510429

src/commons/api/ca.rs
timbru: pushed

src/daemon/ca/status.rs
timbru: This is safely inside a closure in a lock. The downside of `entry()` is that we always need to clone the key. But, we were getting the value twice. I have pushed code which will only insert and clone if they value could not be retrieved in the first place.

src/daemon/http/mod.rs
```diff
@@ -357,7 +358,13 @@ impl Request {
     pub fn user_agent(&self) -> Option<String> {
         match self.headers().get(&USER_AGENT) {
             None => None,
-            Some(value) => value.to_str().ok().map(|s| s.to_string()),
+            Some(value) => value.to_str().ok().map(|s| {
+                if s.len() > HTTP_USER_AGENT_TRUNCATE {
+                    s[..HTTP_USER_AGENT_TRUNCATE].to_string()
```
ximon18, line 363: Is it safe to index into/slice a string, aren't there issues without byte boundaries being different to "character" boundaries?

timbru: Yes, because of `HeaderValue::to_str`. I added a comment to make this more clear.

src/commons/api/ca.rs
timbru: fixed formatting to not use whitespace.

src/commons/api/ca.rs
timbru: The API returns json. It's the CLI that will call `to_string()` which will result in a CSV - unless `--format json` is used of course. The idea was indeed to make sure this is extensible.

src/daemon/http/server.rs
timbru: Moved to dedicated endpoint. You're right.. it wasn't a good match. There was no more namespace under 'api/v1/cas/<ca>/children', because the next level maps to a specific child. But I went for `api/v1/cas/<ca>/stats/children/connections` instead. This also leaves room for extension with other stuff.

src/daemon/ca/status.rs
timbru: Using 'stats' now for the most part. Right now this is essentially the last connection status, but it could be extended to include more information.

src/commons/api/ca.rs
timbru: If and when we need it we can extend `ChildConnectionStats` to include an (optional) last failure, or some other history.

timbru:
> Was it the intent of this work to include Prometheus metrics as well?

I did not think of that.

I propose to postpone this, mainly because I am not entirely sure yet how to capture this nicely in prometheus. We would need to show the connections, success and possibly user agent for all CAs. So this is multi-dimensional. I am not sure how to name (or nest) this.

PR #633, opened 2021-08-11
Rsync in tal

timbru:


ximon18:
The e2e test was working around the lack of rsync in the TAL so I was going to say that without having looked yet at this PR, from the subject line I can imagine that this might break the e2e test - but I see now that it indeed failed :-) I hope to take a look at this PR later today.

ximon18:
Note the packaging workflow fails because of the unmerged PRs relating to fixing a known current issue with `cargo-generate-rpm`.

src/commons/api/ca.rs
```diff
@@ -406,20 +406,27 @@ impl Eq for RcvdCert {}
 
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct TrustAnchorLocator {
-    uris: Vec<uri::Https>, // We won't create TALs with rsync, this is not for parsing.
+    uris: Vec<uri::Https>,
```
ximon18, line 409: Might be good to rename references to `uris` to `http_uris` now that there is also `rsync_uri`?

ximon18 (approved):
Code changes look good. I'll investigate what's wrong with the e2e test.

src/commons/api/ca.rs
timbru: I didn't because it will break deserialization. I could have an alias for serde, but I figured it wasn't important enough. Can be persuaded.. if you prefer the alias..

PR #635, opened 2021-08-16
Suspend inactive children

timbru:


defaults/krill.conf
```diff
@@ -249,4 +249,36 @@
...
-# timing_roa_reissue_weeks_before = 4
\ No newline at end of file
+# timing_roa_reissue_weeks_before = 4
+
+#
+#                 Suspend inactive child CAs
+#
+# NOTE: YOU DO NOT NEED THIS UNLESS YOU DELEGATE RESOURCES TO CHILD CAS
+#
+# If a child CA is deactivated, then it will no longer update its publication
```
ximon18, line 259: Deactivated by whom? The operator of this Krill instance or the operator of the child CA? I think the latter. Perhaps just instead say if the child CA no longer updates its publication point?

defaults/krill.conf
```diff
@@ -249,4 +249,36 @@
...
+# point: i.e. its manifest(s) and CRL(s) will expire and it will not maintain
+# its ROAs. While this is a responsibility of the child CA itself, RPKI operators
+# may expect that RPKI parent CAs assume the responsibility to actively remove
+# such deactivated child CAs.
```
ximon18, line 263: s/deactivated/inactive/ ?

defaults/krill.conf
```diff
@@ -249,4 +249,36 @@
...
+#
+# The trouble is that there is no clear cut way for a parent to know that
+# a child CA had been deactivated. The best indication we have is by looking
+# at the last attempt by the child to send an RFC 6492 request to a parent. If
+# we haven't seen any attempt for longer than X hours we may be tempted to
+# suspend the child CA.
+#
+# By default Krill CAs will send RFC 6492 list every 10 minutes, and they will
```
ximon18, line 271: Is there a verb missing after "RFC 6492 list" ?

defaults/krill.conf
```diff
@@ -249,4 +249,36 @@
...
+# republish manifests and CRLs 8 hours before they would expire. This means that
+# for Krill based child CAs we could use a value of 8 hours if we want to prevent
```
ximon18, line 273: I read "for Krill based child CAs" as meaning something should be done on the child CA system, but here I assume we are configuring the behaviour of the parent?

defaults/krill.conf
```diff
@@ -249,4 +249,36 @@
...
+# that RPKI validators will see expired publication points for inactive children
+# under CAs in this Krill instance. You can enable this by using the following
+# setting:
+#
+# suspend_child_after_inactive_hours = 8
```
ximon18, line 278: What does it actually mean to "suspend" a child CA?

src/commons/api/admin.rs
```diff
@@ -565,28 +565,45 @@ impl AddChildRequest {
 pub struct UpdateChildRequest {
     id_cert: Option<IdCert>,
     resources: Option<ResourceSet>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    suspend: Option<bool>,
```
ximon18, line 570: Why is suspend an Option?

src/commons/api/admin.rs
```diff
@@ -598,6 +615,9 @@ impl fmt::Display for UpdateChildRequest {
         if let Some(resources) = &self.resources {
             write!(f, "new resources: {} ", resources)?;
         }
+        if let Some(suspend) = self.suspend {
+            write!(f, "change suspend status to: {}", suspend)?;
```
ximon18, line 619: Is it definitely a change? If it was Some before and is Some now that's now a change right?

src/commons/api/ca.rs
```diff
@@ -1590,6 +1600,13 @@ impl ChildConnectionStats {
     pub fn new(handle: ChildHandle, last_exchange: Option<ChildExchange>) -> Self {
         ChildConnectionStats { handle, last_exchange }
     }
+
+    pub fn inactive(&self, threshold_hours: i64) -> bool {
+        match &self.last_exchange {
+            None => false, // if there has been no exchange at all, the child is not yet active, rather than inactive
+            Some(exchange) => exchange.timestamp < (Time::now() - Duration::hours(threshold_hours)).timestamp(),
```
ximon18, line 1607: This logic doesn't seem consistent with the documentation above that says that child CAs "republish manifests and CRLs 8 hours before they would expire" as the latter talks about how near to the moment of expiration we are while the code talks about how long it has been since there was an exchange with the client, perhaps the code better matches the documentation where it says "Krill CAs will send RFC 6492 list every 10 minutes"? (and is that a reference to Krill _child_ CAs?)

src/commons/api/history.rs
```diff
@@ -428,6 +428,12 @@ pub enum StorableCaCommand {
     ChildRemove {
         child: ChildHandle,
     },
+    ChildSuspendInactive {
+        child: ChildHandle,
+    },
+    ChildUnsuspend {
```
ximon18, line 434: I think I like Unsuspend as it is clearly the opposite of suspend, but I can't help thinking of "resume" as another term that could be used... perhaps unsuspend is best...

src/commons/api/ca.rs
```diff
@@ -1590,6 +1600,13 @@ impl ChildConnectionStats {
     pub fn new(handle: ChildHandle, last_exchange: Option<ChildExchange>) -> Self {
         ChildConnectionStats { handle, last_exchange }
     }
+
+    pub fn inactive(&self, threshold_hours: i64) -> bool {
```
ximon18, line 1604: There are probably places like this all over the code that use an integer for some unit of time but I was just reminded by https://rust-lang.github.io/api-guidelines/type-safety.html#c-newtype that an alternate approach could be to use an "hours" type rather than i64 with a name including the term "_hours".

src/commons/api/history.rs
```diff
@@ -514,6 +520,12 @@ impl WithStorableDetails for StorableCaCommand {
             StorableCaCommand::ChildRemove { child } => {
                 CommandSummary::new("cmd-ca-child-remove", &self).with_child(child)
             }
+            StorableCaCommand::ChildSuspendInactive { child } => {
+                CommandSummary::new("cmd-ca-child-suspend-inactive", &self).with_child(child)
+            }
+            StorableCaCommand::ChildUnsuspend { child } => {
+                CommandSummary::new("cmd-ca-unsuspend", &self).with_child(child)
```
ximon18, line 527: s/cmd-ca-unsuspend/cmd-ca-child-unsuspend/ ?

src/daemon/ca/certauth.rs
```diff
@@ -207,15 +207,15 @@ impl Aggregate for CertAuth {
...
+                let (issued, removed, suspended) = updates.unpack();
                 for iss in issued {
                     rc.certificate_issued(iss)
                 }
                 for rem in removed {
                     rc.key_revoked(&rem);
 
                     // This loop is inefficient, but certificate revocations are not that common, so it's
-                    // not a big deal. Tracking this better would require that track the child handle somehow.
+                    // not a big deal. Tracking this better would require to track the child handle somehow.
```
ximon18, line 218: "to track" feels strange to me, perhaps s/that track/that we track/ or s/that track/tracking/ instead?

src/daemon/ca/certauth.rs
```diff
@@ -224,6 +224,9 @@ impl Aggregate for CertAuth {
                         }
                     }
                 }
+                for suspended in suspended {
```
ximon18, line 227: "for suspended in suspended" looks confusing, would this work and be better as "suspended.for_each()" instead?

src/daemon/ca/certauth.rs
```diff
@@ -238,6 +241,10 @@ impl Aggregate for CertAuth {
                 self.children.remove(&child);
             }
 
+            CaEvtDet::ChildSuspended { child } => self.children.get_mut(&child).unwrap().suspend(),
```
ximon18, line 244: Are these `unwrap()` calls safe?

src/daemon/ca/certauth.rs
```diff
@@ -964,6 +973,123 @@ impl CertAuth {
...
     }
 
+    // Suspend a child. The intention is that this is called when it is discovered
+    // that the child has been inactive, i.e. not contacting this parent for a pro-longed
+    // period of time (hours).
+    //
+    // When a child is suspended we need to:
+    // - mark it as suspended
+    // - withdraw all certificates issued to it (suspend them)
+    fn child_suspend_inactive(&self, child_handle: &ChildHandle) -> KrillResult<Vec<CaEvt>> {
```
ximon18, line 983: This fn doesn't care about whether the child is inactive or not right, it just suspends the given child, so perhaps remove inactive from the fn name?

src/daemon/ca/certauth.rs
```diff
@@ -964,6 +973,123 @@ impl CertAuth {
...
+
+        Ok(res)
+    }
+
+    // Unsuspend a child. The intention is that this is called automatically when
+    // a suspended (inactive) child CA is seen to contact this parent again.
+    //
+    // When a child is unsuspended we need to:
+    // - mark it as unsuspended
+    // - republish suspended certificates for it, which
```
ximon18, line 1031: I would replace "which" with "if" here as I initially read this as meaning that republished certificates would be given a modified expiration timestamp to ensure that they will not expire for another day, but I suspect it actually means only republish certificates that will not expire for another day, right?

src/daemon/ca/certauth.rs
```diff
@@ -964,6 +973,123 @@ impl CertAuth {
...
+            if certified_keys.is_empty() {
+                continue;
+            }
+
+            let mut cert_updates = ChildCertificateUpdates::default();
+
+            for key in certified_keys {
+                if let Some(issued) = rc.suspended(&key) {
+                    // check that the cert is actually not expired or about to expire and not overclaiming
+                    if issued.validity().not_after() > Time::now() + Duration::days(1)
```
ximon18, line 1064: Should `Time::now() + Duration::days(1)` be grouped together in parentheses?

src/daemon/ca/certauth.rs
```diff
@@ -964,6 +973,123 @@ impl CertAuth {
...
+
+            if certified_keys.is_empty() {
+                continue;
+            }
+
+            let mut cert_updates = ChildCertificateUpdates::default();
+
+            for key in certified_keys {
+                if let Some(issued) = rc.suspended(&key) {
+                    // check that the cert is actually not expired or about to expire and not overclaiming
```
ximon18, line 1063: So, some certs will be republished but others that are valid but nearing (within a day of) expiration will not be. Is it okay to have this "grey" situation where some valid certs are published but others are not? Will this not be confusing?

src/daemon/ca/certauth.rs
```diff
@@ -964,6 +973,123 @@ impl CertAuth {
...
+            res.push(CaEvtDet::child_certificates_updated(
+                handle,
+                version,
+                rcn.clone(),
+                cert_updates,
+            ));
+            version += 1;
+        }
+
+        info!("CA '{}' suspended inactive child '{}'", handle, child_handle);
```
ximon18, line 1020: There is no logic in this function to determine "inactivity" - perhaps remove "inactive" from this text?

src/daemon/ca/child.rs
```diff
@@ -29,20 +45,35 @@ impl LastResponse {}
...
             used_keys: HashMap::new(),
         }
     }
 
+    pub fn is_suspended(&self) -> bool {
+        self.state == ChildState::Suspended
+    }
+
+    pub fn suspend(&mut self) {
+        self.state = ChildState::Suspended;
```
ximon18, line 70: This doesn't actually suspend the child but only mark it as suspended right? Would it be better then not to use the verb but to call this "mark_suspended" or such?

src/daemon/ca/child.rs
```diff
@@ -120,21 +151,37 @@ pub struct Children {
...
+        self.suspended.insert(ki, suspended);
     }
 
     pub fn key_revoked(&mut self, key: &KeyIdentifier) {
         self.inner.remove(key);
+        self.suspended.remove(key);
     }
 
-    pub fn get(&self, ki: &KeyIdentifier) -> Option<&IssuedCert> {
+    pub fn get_issued(&self, ki: &KeyIdentifier) -> Option<&IssuedCert> {
```
ximon18, line 177: Should inner also be likewise renamed to issued?

src/daemon/ca/commands.rs
```diff
@@ -63,6 +63,16 @@ pub enum CmdDet {
...
     ChildRemove(ChildHandle),
 
+    // Suspend a child (done by a background process which checks for inactive children)
+    // When a child is inactive it is assumed that they no longer maintain their repository.
+    // The certificate(s) issued to the child will be removed (and revoked) until
+    // the child is seen again and unsuspended (see below).
+    ChildSuspendInactive(ChildHandle),
+
+    // Unsuspend a child (when it contacts the server again). I.e. re-issue
+    // any and all certificates for it based on the current state.
```
ximon18, line 73: Unsuspend doesn't re-issue any and all certificates, only those that are not about to expire and do not overclaim, right?

src/daemon/ca/commands.rs
```diff
@@ -348,6 +358,14 @@ impl CmdDet {
         eventsourcing::SentCommand::new(handle, None, CmdDet::ChildRemove(child_handle), actor)
     }
 
+    pub fn child_suspend_inactive(handle: &Handle, child_handle: ChildHandle, actor: &Actor) -> Cmd {
```
ximon18, line 361: Same as above, is there really any notion of inactiveness here or is this just a command to suspend and the caller determines whether it is suspending due to inactivity or for some other reason?

src/daemon/ca/events.rs
```diff
@@ -376,31 +376,58 @@ impl fmt::Display for RoaUpdates {
...
+            removed,
+            suspended,
+        }
     }
 
     pub fn is_empty(&self) -> bool {
         self.issued.is_empty() && self.removed.is_empty()
     }
+
+    /// Add an issued certificate to the current set of issued certificates.
```
ximon18, line 395: Doesn't it say above that the child is suspended which means its certificates are unpublished and revoked, i.e. is it right to talk about suspension and unsuspension of the _certificate_ here?

src/daemon/ca/events.rs
```diff
@@ -421,21 +448,30 @@ pub enum CaEvtDet {
...
+    /// A child key was revoked.
     ChildKeyRevoked {
         child: ChildHandle,
         resource_class_name: ResourceClassName,
         ki: KeyIdentifier,
     },
+
+    /// Child certificates (for potentially multiple children) were updated
+    /// under a CA resource class. I.e. child certificates were issued,
+    /// removed, or suspended.
```
ximon18, line 474: Again, and perhaps I'm just losing my way here, are we talking about suspension of child certificates or of child CAs and certificates are not actually suspended?

src/daemon/ca/manager.rs
```diff
@@ -544,24 +552,38 @@ impl CaManager {
 
         let content = ca.verify_rfc6492(msg)?;
 
-        let (child, recipient, content) = content.unpack();
+        let (child_handle, recipient, content) = content.unpack();
+
+        // If the child was suspended, because it was inactive, then we can now conclude
```
ximon18, line 557: Could there ever be a situation where a child was suspended for a reason other than inactivity and then that it might not be right to unsuspend it based solely on fresh activity?

src/daemon/ca/manager.rs
```diff
@@ -743,6 +765,21 @@ impl CaManager {
...
+                    if let Some(threshold_hours) = self.config.suspend_child_after_inactive_hours {
+                        if let Ok(child_stats) = self.ca_stats_child_connections(&ca_handle).await {
+                            for child in child_stats.inactive_children(threshold_hours) {
+                                info!(
+                                    "Child '{}' under CA '{}' was inactive for more than {} hours. Will suspend it.",
+                                    child, ca_handle, threshold_hours
+                                );
+                                let req = UpdateChildRequest::suspend(true);
+                                if let Err(e) = self.ca_child_update(&ca_handle, child, req, actor).await {
+                                    error!("Could not suspend inactive child, error: {}", e);
```
ximon18, line 778: If the cause of this error will mean that suspension will fail on every attempt, is it not the case that this logic will keep trying to suspend the child? Would that be a problem?

src/daemon/config.rs
```diff
@@ -486,7 +488,7 @@ impl Config {
         let auth_users = None;
         #[cfg(feature = "multi-user")]
         let auth_openidconnect = None;
-        let ca_refresh = 1;
+        let ca_refresh = if enable_ca_refresh { 1 } else { 86400 };
```
ximon18, line 491: Should `ca_refresh` be renamed `ca_refresh_seconds` or something here to express the unit of the value?

src/commons/api/admin.rs
```diff
@@ -565,28 +565,45 @@ impl AddChildRequest {
...
 
     pub fn resources(resources: ResourceSet) -> Self {
         UpdateChildRequest {
             id_cert: None,
             resources: Some(resources),
+            suspend: None,
+        }
+    }
+
+    pub fn suspend(suspend: bool) -> Self {
```
ximon18, line 597: Elsewhere you have unsuspend, would it be more consistent to add another method here called unsuspend rather than take a boolean flag?

src/test.rs
```diff
@@ -439,6 +461,7 @@ pub async fn ca_contains_resources(handle: &Handle, resources: &ResourceSet) ->
         if ca_current_resources(handle).await.contains(resources) {
             return true;
         }
+        refresh_all().await;
```
ximon18, line 464: Was something broken here before or is this now needed because of the extension to the tests regarding suspension and unsuspension?

tests/functional.rs
```diff
@@ -622,6 +624,8 @@ async fn functional() {
         ca_equals_resources(&ca3, &ca3_res_reduced).await;
         ca_equals_resources(&ca4, &ca4_res_reduced).await;
         refresh_all().await; // if we skip this, then CA4 will not find out that it's resources were reduced
+        refresh_all().await; // if we skip this, then CA4 will not find out that it's resources were reduced
```
ximon18, line 627: Errm, why call `refresh_all()` multiple times in a row?

ximon18 (approved):

Issue #533, opened 2021-05-21
Perform a session reset whenever the Publication Server restarts


timbru:
Just to be on the safe side it would be good to perform an RRDP session reset whenever the Publication Server restarts.

Reason is that Krill has no way of knowing really if it was started using an old back up of consistent data. In which case RPs would have seen more recent serials compared to its restored state.

The price is that all RPs will be forced to get a snapshot, but as restarts should be infrequent this should not be an issue.

PR #637, opened 2021-08-17
Perform RRDP session reset on startup (if repo is initialized) (#533)

timbru:


timbru:
The code already existed, except.. it was only called in tests.

src/pubd/repository.rs
```diff
@@ -179,9 +176,14 @@ impl RepositoryContentProxy {
         self.read(|content| content.write_repository(config))
     }
 
-    /// Reset the RRDP session
+    /// Reset the RRDP session if is initialized. Otherwise do nothing.
```
ximon18, line 179: s/if is/if it is/

src/pubd/repository.rs
```diff
@@ -179,9 +176,14 @@ impl RepositoryContentProxy {
...
     pub fn session_reset(&self, config: &RepositoryRetentionConfig) -> KrillResult<()> {
-        self.write(|content| content.session_reset(config))
+        if self.cache.read().unwrap().is_some() {
```
ximon18, line 181: Is `unwrap()` correct here or should that be the `?` operator?

src/pubd/repository.rs
```diff
@@ -324,6 +326,8 @@ impl RepositoryContent {
     }
 
     pub fn session_reset(&mut self, config: &RepositoryRetentionConfig) -> KrillResult<()> {
+        info!("Perform RRDP session reset - ensures a consistent view for RPs in case we restarted from a backup.");
```
ximon18, line 329: Performing instead of perform? And perhaps also "rest - ensures" would be nicer as "reset. This ensures" ?

ximon18 (approved):

src/pubd/repository.rs
timbru: This can only fail because of poison errors. And we can only have those in case there was a panic while a lock was hold. Krill is set up to exit on panic. So, this is safe and commonly done. Furthermore.. even if Krill did not crash earlier it should crash now if a poison error would occur because we cannot recover from it.

src/pubd/repository.rs
timbru: done

src/pubd/repository.rs
timbru: done

PR #646, opened 2021-09-01
Use async reqwest for openid connect checks

timbru:


src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -151,19 +153,19 @@ impl OpenIDConnectAuthProvider {
         })
     }
 
-    fn initialize_connection_if_needed(&self) -> KrillResult<()> {
-        let mut conn_guard = self.conn.write().unwrap(); // should never fail, better to panic and crash out if it does
+    async fn initialize_connection_if_needed(&self) -> KrillResult<()> {
+        let mut conn_guard = self.conn.write().await; // should never fail, better to panic and crash out if it does
```
ximon18, line 157: This comment seems to be outdated now as it no longer unwraps so I guess it can be removed.

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -810,8 +818,8 @@ impl OpenIDConnectAuthProvider {
         None
     }
 
-    fn get_connection(&self) -> KrillResult<RwLockReadGuard<Option<ProviderConnectionProperties>>> {
-        let conn_guard = self.conn.read().unwrap(); // should never fail, better to panic and crash out if it does
+    async fn get_connection<'a>(&'_ self) -> KrillResult<RwLockReadGuard<'_, Option<ProviderConnectionProperties>>> {
+        let conn_guard = self.conn.read().await; // should never fail, better to panic and crash out if it does
```
ximon18, line 822: Ditto, no longer unwraps, outdated comment?

src/daemon/auth/providers/openid_connect/provider.rs
```diff
@@ -1087,17 +1098,17 @@ impl AuthProvider for OpenIDConnectAuthProvider {
...
-        self.initialize_connection_if_needed().map_err(|err| {
+        self.initialize_connection_if_needed().await.map_err(|err| {
             OpenIDConnectAuthProvider::internal_error(
                 "OpenID Connect: Cannot authenticate request: Failed to connect to provider",
                 Some(&stringify_cause_chain(err)),
             )
         })?;
 
-        let res = match self.get_bearer_token(request) {
+        let res = match AdminTokenAuthProvider::get_bearer_token(request) {
```
ximon18, line 1111: Maybe ideally this would be pulled out as it's not specific to the AdminToken provider.

src/daemon/auth/providers/openid_connect/provider.rs
timbru: indeed, done

src/daemon/auth/providers/openid_connect/provider.rs
timbru: indeed, done

src/daemon/auth/providers/openid_connect/provider.rs
timbru: moved to `commons::util::httpclient`.

src/commons/util/httpclient.rs
```diff
@@ -58,6 +58,24 @@ fn report_delete(uri: &str, content_type: Option<&str>, token: Option<&Token>) {
     }
 }
 
+/// Gets the Bearer token from the request header, if present.
+pub fn get_bearer_token(request: &hyper::Request<hyper::Body>) -> Option<Token> {
+    if let Some(header) = request.headers().get("Authorization") {
```
ximon18, line 63: This was not caused by you but I don't like the hard-coded string header name which could easily be broken by a typo, we should instead use https://docs.rs/hyper/0.14.12/hyper/header/constant.AUTHORIZATION.html. I'm not sure if it's really better but this could also be written [in more functional style like this](https://gist.github.com/rust-play/e2013acd78fd6b3e355febe387b9ecc1).

ximon18 (approved):

src/commons/util/httpclient.rs
timbru: I agree.. now fixed.

Issue #648, opened 2021-09-04
Track last successful connection to parent, and add to metrics


timbru:
We keep track of the last connection and the result, but not _yet_ of the last successful connection time. This would be useful to do.

Furthermore add to prometheus metrics:
- krill_cas_parent_state {ca = , parent = } 0/1
- krill_cas_parent_last_success {ca = , parent = } timestamp

Or something like this..

PR #649, opened 2021-09-06
CA status and metrics (#648)

timbru:


src/daemon/http/server.rs
```diff
@@ -508,6 +515,49 @@ pub async fn metrics(req: Request) -> RoutingResult {
                 res.push_str(&format!("krill_cas_bgp_announcements_valid{{ca=\"{}\"}} {}\n", ca, nr));
             }
 
+            res.push('\n');
+            res.push_str("# HELP krill_ca_parent_success status of last ca-parent connection (0=issue, 1=success)\n");
+            res.push_str("# TYPE krill_ca_parent_success gauge\n");
```
ximon18, line 520: Should this be a counter rather than a gauge as a counter is for values that "can only increase or be reset to zero on restart" while a gauge is for a value that "can arbitrarily go up and down" ? (based on https://prometheus.io/docs/concepts/metric_types/)

src/daemon/http/server.rs
```diff
@@ -508,6 +515,49 @@ pub async fn metrics(req: Request) -> RoutingResult {
...
+                                ca, parent, value
+                            ));
+                        }
+                    }
+                }
+            }
+
+            res.push('\n');
+            res.push_str(
+                "# HELP krill_ca_parent_last_success_time timestamp of last successful ca-parent connection\n",
```
ximon18, line 539: What format is this timestamp in? Do we need to say something about the format or about the time zone?

ximon18 (approved):

src/daemon/http/server.rs
ximon18: I meant to also say that I can imagine scenarios where time doesn't just increase as expected so maybe that's another reason for a gauge?

ties: Typically you would not reset a counter to zero except on program restart. 

If you want this to be a counter you could count the number of successful events and monitor how long it did not increase

timbru: I suppose this is opinion and not gospel, but still well phrased:
https://www.robustperception.io/are-increasing-timestamps-counters-or-gauges

In short: use gauge for timestamp. Time can even go backwards due to NTP.

Also, we have been using gauge for existing timestamp metrics (krill_server_start).

src/daemon/http/server.rs
timbru: I believe that 'timestamp', unless specified otherwise, is generally understood to be seconds since 1/1/1970. But.. I will clarify the help texts to be clear on this - no harm in that!

ximon18: I've never come across that definition of timestamp before. I'm used to seeing seconds since 1/1/1970 as "epoch time" though.

src/daemon/http/server.rs
ximon18: > Time can even go backwards due to NTP.

Yup, I alluded to that above. Gauge it is then.

src/daemon/http/server.rs
timbru: Changed it to "unix timestamp in seconds". (epoch time is wrong according to wikipedia because other epochs exist)

defaults/krill.conf
```diff
@@ -106,6 +106,102 @@
...
 ### service_uri = "https://localhost:3000/"
 
+
+######################################################################################
+#                                                                                    #
+#                                     METRICS                                        #
+#                                                                                    #
+######################################################################################
+
+# Krill exposes metrics for use with prometheus under: <service_uri>/metrics
```
ximon18, line 116: Capital P for Prometheus?

defaults/krill.conf
```diff
@@ -106,6 +106,102 @@
...
+#
+# By default a lot of details are included in these metrics. Generating these metrics
+# is not particularly hard on Krill, but in case you have many CAs, children or publishers
+# under your Krill instance you may still want to disable certain metrics to avoid
```
ximon18, line 120: As we don't know that it can overwhelm Prometheus maybe instead of "to avoid overwhelming your prometheus instance with too much data" say "to reduce the amount of data fetched and stored by Prometheus" ?

ximon18: Should we recommend a fetch frequency? The Prometheus default is once a minute according to https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config. Does Krill cache the metric data and thus fetching more often than the cache period wouldn't make sense?

defaults/krill.conf
```diff
@@ -106,6 +106,102 @@
...
+# are other ways than using Prometheus to monitor details.
+
+# Always enabled:
+#################
+#
+# krill_server_start              unix timestamp in seconds of last krill server start
+# krill_version_major             krill server major version number
+# krill_version_minor             krill server minor version number
+# krill_version_patch             krill server patch version number
+# krill_auth_session_cache_size   total number of cached login session tokens
```
ximon18, line 133: It's not strictly true that this is always enabled as it is behind a feature flag, I can guess it's simpler to stick to the default feature set for these docs otherwise they will become complicated for the majority case for the sake of the few if any customers who change the Rust feature set used to build Krill.

ximon18 (approved):

defaults/krill.conf
timbru: ack

defaults/krill.conf
timbru: sure wfm

timbru: regarding fetch frequency: krill caches status for reporting so doesn't mind. You don't know when something might change.. as soon as something does the cache is updated. So, local policy.. no recommendation I think.

defaults/krill.conf
timbru: true, fixed

defaults/krill.conf
```diff
@@ -152,6 +152,8 @@
 #  CA -> Child metrics
 #  -------------------
 # 
+# NOTE: These metrics are only shown if have any child CAs under your CA(s) in Krill.
```
ximon18, line 155: s/if have/if you have/

src/commons/api/ca.rs
```diff
@@ -141,23 +141,55 @@ impl From<&IdCert> for IdCertPem {
     }
 }
 
+//------------ ChildState ----------------------------------------------------
+
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[allow(clippy::large_enum_variant)]
```
ximon18, line 147: This clippy directiver is redundant here, right?

src/daemon/http/server.rs
```diff
@@ -531,11 +531,16 @@ pub async fn metrics(req: Request) -> RoutingResult {
                     }
                 }
 
-                if !server.config.metrics.metrics_hide_child_details {
+                // Do not show child metrics if none of the CAs has any children..
+                // Many users do not delegate so, showing these metrics would just be confusing.
+                let any_children = cas_stats.values().any(|ca| ca.child_count() > 0);
```
ximon18, line 536: A minor thing, but would `!is_empty()` here be more idiomatic and potentially less computation?

src/daemon/http/server.rs
```diff
@@ -572,8 +577,24 @@ pub async fn metrics(req: Request) -> RoutingResult {
 
                     res.push('\n');
                     res.push_str(
-                    "# HELP krill_ca_child_last_connection unix timestamp in seconds of last child to CA connection\n",
-                );
+                        "# HELP krill_ca_child_state child state (see 'suspend_child_after_inactive_hours' config) (0=suspended, 1=active)\n",
+                    );
+                    res.push_str("# TYPE krill_ca_child_state gauge\n");
+                    for (ca, status) in ca_status_map.iter() {
+                        // skip the ones for which we have no status yet, i.e it was really only just added
```
ximon18, line 584: This loop doesn't seem to skip anything.

ximon18 (approved):

src/daemon/http/server.rs
timbru: no, child_count is just a number on `CertAuthStats`.

defaults/krill.conf

src/commons/api/ca.rs
timbru: indeed, removed

src/daemon/http/server.rs
timbru: removed comment

PR #652, opened 2021-09-06
Monitor ca to publication server connections with prometheus. (#650)

timbru:
Builds on the other PR for monitoring ca-parent connections. Propose to merge into that, and then both to dev.

### DIFF
--- a/src/commons/api/ca.rs
+++ b/src/commons/api/ca.rs
@@ -1437,6 +1437,7 @@ impl Default for ParentStatus {
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct RepoStatus {
     last_exchange: Option<ParentExchange>,
+    last_success: Option<i64>,
     next_exchange_before: i64,
     published: Vec<PublishElement>,
 }
@@ -1445,6 +1446,7 @@ impl Default for RepoStatus {
     fn default() -> Self {
         RepoStatus {
             last_exchange: None,
+            last_success: None,
             next_exchange_before: Self::now_plus_hours(1),
             published: vec![],
         }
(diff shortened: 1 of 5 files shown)


src/commons/api/ca.rs
```diff
@@ -1445,21 +1446,27 @@ impl Default for RepoStatus {
...
             next_exchange_before: Self::now_plus_hours(1),
             published: vec![],
         }
     }
 }
 
 impl RepoStatus {
-    fn next_exchange_before(&self) -> Time {
+    pub fn next_exchange_before(&self) -> Time {
         Time::new(Utc.timestamp(self.next_exchange_before, 0))
```
ximon18, line 1458: Since `TIme::new(Utc.timestamp(..., 0))` is invoked in multiple places maybe it's worth factoring out to ensure consistency?

timbru: yeah this is done across the code base.. and in a lot of places (at least for the metrics) we then call .timestamp() again. So perhaps it should just return the timestamps instead.

src/daemon/http/server.rs
```diff
@@ -498,13 +498,18 @@ pub async fn metrics(req: Request) -> RoutingResult {
                 roas_total: HashMap::new(),
             };
 
-            let mut parent_stats: HashMap<Handle, ParentStatuses> = HashMap::new();
+            let mut ca_parent_statuses: HashMap<Handle, ParentStatuses> = HashMap::new();
```
ximon18, line 501: If both hash maps are always updated together and walked one after the other does it make sense to combine them?

src/daemon/http/server.rs
```diff
@@ -558,10 +563,52 @@ pub async fn metrics(req: Request) -> RoutingResult {
...
+                        res.push_str(&format!(
+                            "krill_ca_repo_last_success_time{{ca=\"{}\"}} {}\n",
+                            ca,
+                            last_success.timestamp()
+                        ));
+                    }
+                }
+
+                res.push('\n');
+                res.push_str("# HELP krill_ca_repo_next_before_time timestamp of next planned ca to publication server connection\n");
```
ximon18, line 597: Consistency: Sometimes "ca" is mentioned in metric help texts and other times "CA" is mentioned.

ximon18 (approved):

src/commons/api/ca.rs
timbru: now using new type `Timestamp` and avoiding overhead.

src/daemon/http/server.rs
timbru: separate sets

src/daemon/http/server.rs
timbru: fixed (on the branch this was merged into)

PR #653, opened 2021-09-07
Monitor child cas

timbru:


src/commons/api/ca.rs
```diff
@@ -1574,6 +1571,15 @@ pub enum ExchangeResult {
     Failure(ErrorResponse),
 }
 
+impl ExchangeResult {
+    pub fn was_success(&self) -> bool {
+        match self {
```
ximon18, line 1576: I'm not sure it's better but I think you could shrink this to something like `matches!(self, ExchangeResult::Success)`.

src/daemon/http/server.rs
```diff
@@ -511,53 +512,59 @@ pub async fn metrics(req: Request) -> RoutingResult {
...
-                    for (parent, status) in parent_statuses.iter() {
-                        // skip the ones for which we have no status yet, i.e it was really only just added
-                        // and no attempt to connect has yet been made.
-                        if let Some(exchange) = status.last_exchange() {
-                            let value = if exchange.was_success() { 1 } else { 0 };
-                            res.push_str(&format!(
-                                "krill_ca_parent_success{{ca=\"{}\", parent=\"{}\"}} {}\n",
-                                ca, parent, value
-                            ));
+            {
```
ximon18, line 521: It's a bit hard to see from the GH diff but does each last success time metric have a corresponding last connection time metric?

src/daemon/http/server.rs
```diff
@@ -607,62 +618,163 @@ pub async fn metrics(req: Request) -> RoutingResult {
...
+                }
 
-            res.push('\n');
-            res.push_str("# HELP krill_cas_bgp_roas_stale number of ROAs for this CA for which no announcements are seen (0 may also indicate that no BGP info is available)\n");
-            res.push_str("# TYPE krill_cas_bgp_roas_stale gauge\n");
-            for (ca, nr) in all_bgp_stats.roas_stale.iter() {
-                res.push_str(&format!("krill_cas_bgp_roas_stale{{ca=\"{}\"}} {}\n", ca, nr));
+                res.push('\n');
+                res.push_str(
+                    "# HELP krill_ca_child_agent_total total children per user agent on their last connection\n",
```
ximon18, line 687: Should this perhaps be called `krill_ca_child_agent_last_total` ?

ximon18 (approved):

PR #663, opened 2021-09-15
Ca status cache fixes

timbru:
Would like to merge (after review) so we can test a new rc2 on testbed - even if we do not release it publicly.

### DIFF
--- a/src/daemon/ca/manager.rs
+++ b/src/daemon/ca/manager.rs
@@ -12,7 +12,6 @@ use rpki::{repository::crypto::KeyIdentifier, uri};
 use crate::{
     commons::{
         actor::Actor,
-        api::rrdp::PublishElement,
         api::{
             self, AddChildRequest, Base64, CaCommandDetails, CaCommandResult, CertAuthList, CertAuthSummary,
             ChildCaInfo, ChildHandle, CommandHistory, CommandHistoryCriteria, Entitlements, Handle, IssuanceRequest,
@@ -20,6 +19,7 @@ use crate::{
             RepositoryContact, ResourceClassName, ResourceSet, RevocationRequest, RevocationResponse, RtaName,
             StoredEffect, UpdateChildRequest,
         },
+        api::{rrdp::PublishElement, Timestamp},
         crypto::{IdCert, KrillSigner, ProtocolCms, ProtocolCmsBuilder},
         error::Error,
         eventsourcing::{Aggregate, AggregateStore, Command, CommandKey},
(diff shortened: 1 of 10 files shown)


src/daemon/ca/status.rs
```diff
@@ -177,6 +190,30 @@ impl StatusStore {
...
+    /// a new empty default status will be re-generated when it is accessed for this CA.
+    pub async fn remove_ca(&self, ca: &Handle) -> KrillResult<()> {
+        let mut cache = self.cache.write().await;
+
+        if !cache.contains_key(ca) {
+            Ok(()) // idempotent
+        } else {
+            let scope = ca.as_str();
+
+            self.store.drop_scope(scope)?; // idem potent; won't do anything if the scope does not exist.
```
ximon18, line 209: s/idem potent/idempotent/, but perhaps also redundant to say as "won't do anything" also implies it is idempotent.

Also, you could slightly simplify this fn by doing:
```
if cache.contains_key(ca) {
    ...
}
Ok(())
```

ximon18 (approved):

ximon18 (approved):

ximon18 (approved):

src/daemon/ca/status.rs
timbru: Fixed the comment.

Left the logic as-is, to me it's more explicit this way the removal just won't do anything when there is nothing to remove, rather then throwing an error.

src/daemon/ca/manager.rs
```diff
@@ -807,17 +807,33 @@ impl CaManager {
...
             let mut updates = vec![];
 
+            // Set threshold hours if it was configured AND this server has been started
+            // longer ago than the hours specified. Otherwise we risk that *all* children
+            // without prior recorded status are suspended on upgrade, or that *all* children
+            // are suspended if the server had been down for more than the threshold hours.
+            let threshold_hours = self
+                .config
+                .suspend_child_after_inactive_hours
+                .map(|hours| {
```
ximon18, line 821: Couldn't this be more simply done using [filter](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.filter) ?

ximon18: Should we log that we are delaying the check?

ximon18 (approved):

src/daemon/ca/manager.rs
timbru: Now using filter, thanks for the pointer.. I don't think we need to log. It would be incompatible with filter again, but more importantly I like to keep logging quiet. There is no problem in postponing.

ximon18 (approved):

PR #672, opened 2021-09-22
Suspend krill 092 higher only

timbru:
Now with tests!


src/daemon/config.rs
```diff
@@ -761,7 +773,7 @@ impl Config {
         }
 
         if let Some(threshold) = self.suspend_child_after_inactive_hours {
-            if threshold < 1 {
+            if threshold < CA_SUSPEND_MIN_HOURS {
                 return Err(ConfigError::other(
                     "suspend_child_after_inactive_hours must be 1 or higher (or not set at all)",
```
ximon18, line 778: The constant is set to 2 but the text says 1.

src/daemon/config.rs
```diff
@@ -249,7 +249,9 @@ pub struct Config {
     #[serde(default = "ConfigDefaults::ca_refresh_seconds", alias = "ca_refresh")]
     pub ca_refresh_seconds: u32,
 
-    pub suspend_child_after_inactive_hours: Option<i64>,
+    #[serde(skip)]
+    suspend_child_after_inactive_seconds: Option<i64>,
```
ximon18, line 253: Why both hours and seconds? Is it because hours has already been serialized by some deployments of Krill and thus must continue to be supported?

timbru: no, it's because I need to use seconds internally but wanted let users configure hours. We are not very consistent in time units in the configuration file. I think this is okay, because it depends on the context. In this case you really should use multiple hours and then the math using seconds gets annoying.. (unless you dream DNS I guess)

ximon18: Right, but you can express hours internally as seconds and then not store both? (and use `Duration` to handle the "math"?)

src/daemon/config.rs
```diff
@@ -578,6 +587,8 @@ impl Config {
             None
         };
 
+        let suspend_child_after_inactive_seconds = if enable_suspend { Some(3) } else { None };
```
ximon18, line 590: Should this be a constant or configurable (instead of false use None/Some)?

src/daemon/config.rs
timbru: Then I need custom deserialization when parsing the config file, I just figured this was less complicated.

timbru: I would be nice to let users use notations like: 1s, 5m, 2h etc but it requires more work and we should then also review the other parameters. And probably remember for each attribute what the default unit should be for existing configurations where it is not specified.

ximon18: I wasn't referring to end user configuration, just internal representation (but yes it would be nice - add it to the todo list :-))

timbru: Note that the 'seconds' field is not (de)serialized - so users can't set it

ximon18: Yes I realize that. I am only referring to having a single representation of this value internally in the Krill config struct.

src/daemon/ca/manager.rs
```diff
@@ -811,54 +811,70 @@ impl CaManager {
...
+                            child, ca_handle, threshold_string
+                        );
+                        if let Err(e) = self
+                            .status_store
+                            .lock()
+                            .await
+                            .set_child_suspended(&ca_handle, &child)
+                            .await
+                        {
+                            error!(
```
ximon18, line 862: Under what circumstances could this fail? Is logging an appropriate response? Not being able to update the status seems quite a severe problem but I haven't dived into the called code tree to see what it could really mean.

src/daemon/http/server.rs
```diff
@@ -1168,6 +1169,19 @@ async fn api_ca_stats(req: Request, path: &mut RequestPath, ca: Handle) -> Routi
     }
 }
 
+async fn api_ca_sync(req: Request, path: &mut RequestPath, ca: Handle) -> RoutingResult {
+    if req.is_post() {
+        let actor = req.actor();
```
ximon18, line 1174: This currently only checks at the parent level if the caller has CA_READ permission. Is that a suitable permission for something that has side-effects like sync or should we enforce a higher permission requirement here?

src/test.rs
```diff
@@ -196,10 +204,32 @@ pub async fn krill_admin_expect_error(command: Command) -> Error {
     }
 }
 
-pub async fn refresh_all() {
+pub async fn cas_all_sync_parents() {
```
ximon18, line 207: Above there are separate sync and refresh functions. Here we have a sync function with a refresh command. Is this correct?

tests/functional.rs
```diff
@@ -590,7 +590,7 @@ async fn functional() {
     info("##################################################################");
     info("");
     {
-        refresh_all().await;
+        cas_all_sync_parents().await;
```
ximon18, line 593: Ditto: was refresh, now sync, am confused about the terminology and wondering if this is correct?

tests/suspend.rs
```diff
@@ -0,0 +1,115 @@
...
+    info("#                                                                #");
+    info("#   Uses the following lay-out:                                  #");
+    info("#                                                                #");
+    info("#                  TA                                            #");
+    info("#                   |                                            #");
+    info("#                testbed                                         #");
+    info("#                   |                                            #");
+    info("#                  CA                                            #");
+    info("#                                                                #");
+    info("#  - Disable the normal background jobs for this test            #");
```
ximon18, line 26: Which background job and where does this code disable it?

tests/suspend.rs
```diff
@@ -0,0 +1,115 @@
...
+    info("#  - Set suspension threshold to 5 seconds                       #");
```
ximon18, line 27: The code in `config.rs` says:
```rust
let suspend_child_after_inactive_seconds = if enable_suspend { Some(3) } else { None };
```

Where is this set to 5 seconds?

tests/suspend.rs
```diff
@@ -0,0 +1,115 @@
...
+    info("#  - Test that a certificate is published for CA                 #");
+    info("#  - Wait 6 seconds                                              #");
```
ximon18, line 29: This says wait 6 seconds but below the code waits 5 seconds.

src/daemon/config.rs
timbru: fixed

ximon18 (approved):

ximon18 (approved):

src/daemon/config.rs
timbru: it could be prettier, but it's not really in the way right now and requires refactoring many tests. So I prefer to just leave it. I used a bool because it felt consistent with 'enable_testbed' and 'enable_ca_refresh'.

src/daemon/ca/manager.rs
timbru: fair point.. it can't really fail - unless there is an I/O error updating the disk, in which case a panic and a crash is probably the only sensible thing to do

ximon18 (approved):

src/daemon/ca/manager.rs
timbru: pushed a change to panic here - see also issue #675

src/daemon/http/server.rs
timbru: Changed to require CA_UPDATE. It doesn't really matter because this can't break anything, but still.

tests/functional.rs
timbru: using 'refresh' again

tests/suspend.rs
timbru: changed this to use comments, and fixed the comment

tests/suspend.rs
timbru: using code comments now

tests/suspend.rs
timbru: using code comments now

Issue #539, opened 2021-06-10
Extend KrillSigner to support multiple signers


ximon18:
One signer should be the default which will be used for generating new key pairs. Other signers will be used to work with keys that were previously created with those signers. Extend KrillSigner functions to lookup the signer to use based on the Key Identifier, or in the case of key generation by using the signer marked as the “default”. Only the SoftSigner exists at this point but the collection should support any signer implementation.

ximon18:
Delivered with the [Krill v0.10.0](https://github.com/NLnetLabs/krill/releases/tag/v0.10.0) release.

PR #674, opened 2021-09-23
HSM: Support multiple signers (#539)

ximon18:
Separates out signer dispatch from KrillSigner to a new SignerProvider struct so that SignerProvider can handle the noisy enum based dispatch to a particular Signer implementation and KrillSigner can focus on business logic. Adds a 'hsm' feature gate and puts the a 'DummySigner' behind it.


src/commons/crypto/signing.rs
```diff
@@ -36,84 +40,244 @@ use crate::{
...
+                key,
+                &base_repo.ca_repository(name_space).join(&[]).unwrap(), // force trailing slash
+                &base_repo.rpki_manifest(name_space, &pub_key.key_identifier()),
+                Some(&base_repo.rpki_notify()),
+            )
+            .map_err(crypto::Error::signing)
+        }
+
+        let enc = match self {
+            SignerProvider::OpenSsl(signer) => func(signer.deref(), base_repo, name_space, key),
```
timbru, line 133: I am (generally) not fond of fn names like 'func'. Perhaps a more descriptive thing like 'sing_csr_with_signer'? Or.. is there a way to use the match to get the signer to use without the need for an inner fn?

src/commons/util/dummysigner.rs
```diff
@@ -0,0 +1,48 @@
+use rpki::repository::crypto::{KeyIdentifier, PublicKey, PublicKeyFormat, Signature, SignatureAlgorithm, Signer};
+
+use super::softsigner::SignerError;
+
+/// A dummy signer to prove that compilation with two different Signer implementations works
```
timbru, line 5: That makes sense for now, but will this be removed when real signer implementations are added?

timbru (approved):
Looks like a good architecture. I had to minor remarks - but neither are show stoppers at this time.

src/commons/crypto/signing.rs
ximon18: Oops, that should indeed not be called `func`.

src/commons/util/dummysigner.rs
ximon18: Indeed. The other PR I am working (https://github.com/NLnetLabs/krill/pull/679) on already removes it (well, replaces it with a real KMIP signer).

ximon18:
Discussed with @timbru that we will not merge this until after the coming v0.9.2 release just to keep our lives that tiny bit easier.

Issue #566, opened 2021-06-10
Implement Krill KMIP based Signer implementation


ximon18:
Extend the Signer enum with a value for a KMIP based wrapper struct around the KMIP TLS library.

ximon18:
Delivered with the [Krill v0.10.0](https://github.com/NLnetLabs/krill/releases/tag/v0.10.0) release.

PR #679, opened 2021-09-28
HSM 1: KMIP walking skeleton (#566)

ximon18:
Introduces a KMIP signer, disabled by default.

The scope for this PR is deliberately limited. The KMIP signer itself is fully functional but it is not yet fully wired into Krill. It is not configurable, it has no metrics, it has no persistent mapping of Krill key identifiers to HSM key identifiers _(**update:** added in PR https://github.com/NLnetLabs/krill/pull/686 for that)_, it doesn't support async operation _(**update:** async will not be added at the moment)_, there's no extensions to the API or CLI to see which Signer owns a particular key, etc. This PR is just about showing that the Krill test suite passes when forcing Krill to use a PyKMIP server via the new KMIP signer.

KMIP support can only be enabled currently by using BOTH the `hsm` and `hsm-tests` feature flags. In normal usage we may want to use a local signer for some operations even when using a KMIP signer for others. The `hsm-tests` feature flag forces Krill to use KMIP for ALL operations. This is used by a new GitHub Actions `hsmtest` job in the CI workflow to test Krill against [PyKMIP](https://github.com/OpenKMIP/PyKMIP). If you look at the [test log output](https://github.com/NLnetLabs/krill/pull/679/checks?check_run_id=3802692825) you can see lines like the following:

```
2021-10-05 11:32:09 [DEBUG] [krill::commons::crypto::signers::kmip::internal] Probing server at 127.0.0.1:5696
2021-10-05 11:32:09 [WARN] [krill::commons::crypto::signers::kmip::internal] KMIP server lacks support for one or more required operations: ModifyAttribute
2021-10-05 11:32:09 [INFO] [krill::commons::crypto::signers::kmip::internal] Using KMIP server 'PyKMIP 0.10.0 Software Server' at 127.0.0.1:5696
```

**To be discussed.**
- OpenSSL is used as it was difficult to make RustLS connect to PyKMIP, and while eventually that was solved it may be unhelpful to customers to force the strict "no unsafe features (e.g. self-signed certificates) and no obsolete crpytography" opinions of RustLS on them as it may be too strict for real world HSM deployments.

**Other notes:**
- Uses newly released [`bcder`](https://crates.io/crates/bcder/) crate DER unsigned integer support.
- Lacks tests of its own though is tested by the Krill test suite and the `kmip-protocol` and `kmip-ttlv` crates have tests, not exhaustive but also quite a bit more than nothing.
- The target server is hard-coded to local host with certificates for testing with PyKMIP. In fact there are no changes anywhere in Krill other than adding support for this signer as a replacement for the existing OpenSSL based "soft signer".

Details:

- Adds a dependency on the [`kmip-protocol`](https://crates.io/crates/kmip-protocol/) crate for TCP+TLS high level KMIP client support based on the underlying [`kmip-ttlv`](https://crates.io/crates/kmip-ttlv/) crate, neither of which has been reviewed yet.
- Adds a dependency on the [`backoff`](https://crates.io/crates/backoff/) crate for retry support.
- Adds a dependency on the [`r2d2`](https://crates.io/crates/r2d2/) crate for connection pooling support.
- Refactor signers to `crypto::signers` and replace the Dummy signer with a KMIP signer.
- Added a "hsmtest" job to the GitHub Actions CI workflow that runs all Krill tests using the KMIP signer against PyKMIP.
- Added a "hsm-tests" Cargo feature flag for configuring Krill to use ONLY KMIP as a signer, not OpenSSL at all.
  Currently building without the "hsm-tests" feature flag set will fail if the "hsm" feature flag is set.
  Krill isn't ready to be used in "hsm" mode yet.
- Introduces `SignerRouter` between `KrillSigner` and `SignerProvider`, to implement the `Signer` trait so that it can be passed to builders so that their invocation of a signer also goes via `SignerRouter` thus dispatching to the correct signer.

This PR doesn't offer any way to see that the keys have actually been created. However, the tests pass, keys are created and signing is done, the KMIP signer is shown connecting to the PyKMIP server in the logs, and the OpenSSL signer is disabled by the feature flags used by the `hsmtest` job. If you watch the PyKMIP log while the test is running you can see that it is connecting and executing KMIP operations.

> **Update:** I've updated the GH Actions CI workflow in this PR branch to always dump the PyKMIP log file, not only on error, which should make it easier for reviewers to see that there is indeed some activity.

Using a custom local KMIP client when I inspected the keys inside my local PyKMIP instance before and after running the Krill test suite, before it was empty, and after it looked like something this:

```
Key: 1787, Name: 'krill-public-key-2a5d9ec24d8b8fe7a17b24d15a381c01c0b47c7b'
Key: 1788, Name: 'krill-private-key-2a5d9ec24d8b8fe7a17b24d15a381c01c0b47c7b'
Key: 1765, Name: 'krill-public-key-d46af8f74d79055a0d0da6ff806f089c8f18eb30'
Key: 1766, Name: 'krill-private-key-d46af8f74d79055a0d0da6ff806f089c8f18eb30'
Key: 1755, Name: 'krill-public-key-cd27dfee6834107489e2eebb54899b8c9038d1a8'
Key: 1756, Name: 'krill-private-key-cd27dfee6834107489e2eebb54899b8c9038d1a8'
...
Key: 3, Name: 'krill-public-key-a3897008c85f8ad7ad2658c89e1dcdbca6301485'
Key: 4, Name: 'krill-private-key-a3897008c85f8ad7ad2658c89e1dcdbca6301485'
Key: 5, Name: 'krill-public-key-7809b44da1d6be664db7cabd7deb9009c94ff5c3'
Key: 6, Name: 'krill-private-key-7809b44da1d6be664db7cabd7deb9009c94ff5c3'
Key: 1, Name: 'krill-public-key-0daebf72d4b7aecf00da86db43ed220a3df78c59'
Key: 2, Name: 'krill-private-key-0daebf72d4b7aecf00da86db43ed220a3df78c59'
```

If you run Krill with trace level logging enabled you will then see a stripped diagnostic representation of the KMIP requests and responses being sent and received, e.g.:

```
2021-10-05 17:10:47 [TRACE] [kmip_protocol::client::client] KMIP TTLV request: 78[77[69[6Ai6Bi]0Di]0F[5Ce2:79[1F[08[0At0Be4:]08[0At0Bi]]65[08[0At0B[55t54e1:]]08[0At0Bi]]6E[08[0At0B[55t54e1:]]08[0At0Bi]]]]]
2021-10-05 17:10:47 [TRACE] [kmip_protocol::client::client] KMIP TTLV response: 7B[7A[69[6Ai6Bi]92d0Di]0F[5Ce2:7Fe0:7C[66t6Ft]]]
```

The `diag_to_txt` example included with the `kmip-protocol` crate can render these diagnostic strings in slightly more readable form, e.g. the first line from the log example above renders like so:

```
 cargo run --example diag_to_txt /tmp/r
   Compiling kmip-protocol v0.3.3-dev (/home/ximon/src/krill-kmip-protocol)
    Finished dev [unoptimized + debuginfo] target(s) in 4.37s
     Running `target/debug/examples/diag_to_txt /tmp/r`
Tag: Request Message (0x420078), Type: Structure (0x01), Data: 
  Tag: Request Header (0x420077), Type: Structure (0x01), Data: 
    Tag: Protocol Version (0x420069), Type: Structure (0x01), Data: 
      Tag: Protocol Version Major (0x42006A), Type: Integer (0x02), Data: <redacted>
      Tag: Protocol Version Minor (0x42006B), Type: Integer (0x02), Data: <redacted>
    Tag: Batch Count (0x42000D), Type: Integer (0x02), Data: <redacted>
  Tag: Batch Item (0x42000F), Type: Structure (0x01), Data: 
    Tag: Operation (0x42005C), Type: Enumeration (0x05), Data: 2
    Tag: Request Payload (0x420079), Type: Structure (0x01), Data: 
      Tag: Common Template-Attribute (0x42001F), Type: Structure (0x01), Data: 
        Tag: Attribute (0x420008), Type: Structure (0x01), Data: 
          Tag: Attribute Name (0x42000A), Type: TextString (0x07), Data: <redacted>
          Tag: Attribute Value (0x42000B), Type: Enumeration (0x05), Data: 4
        Tag: Attribute (0x420008), Type: Structure (0x01), Data: 
          Tag: Attribute Name (0x42000A), Type: TextString (0x07), Data: <redacted>
          Tag: Attribute Value (0x42000B), Type: Integer (0x02), Data: <redacted>
      Tag: Private Key Template-Attribute (0x420065), Type: Structure (0x01), Data: 
        Tag: Attribute (0x420008), Type: Structure (0x01), Data: 
          Tag: Attribute Name (0x42000A), Type: TextString (0x07), Data: <redacted>
          Tag: Attribute Value (0x42000B), Type: Structure (0x01), Data: 
            Tag: Name Value (0x420055), Type: TextString (0x07), Data: <redacted>
            Tag: Name Type (0x420054), Type: Enumeration (0x05), Data: 1
        Tag: Attribute (0x420008), Type: Structure (0x01), Data: 
          Tag: Attribute Name (0x42000A), Type: TextString (0x07), Data: <redacted>
          Tag: Attribute Value (0x42000B), Type: Integer (0x02), Data: <redacted>
      Tag: Public Key Template-Attribute (0x42006E), Type: Structure (0x01), Data: 
        Tag: Attribute (0x420008), Type: Structure (0x01), Data: 
          Tag: Attribute Name (0x42000A), Type: TextString (0x07), Data: <redacted>
          Tag: Attribute Value (0x42000B), Type: Structure (0x01), Data: 
            Tag: Name Value (0x420055), Type: TextString (0x07), Data: <redacted>
            Tag: Name Type (0x420054), Type: Enumeration (0x05), Data: 1
        Tag: Attribute (0x420008), Type: Structure (0x01), Data: 
          Tag: Attribute Name (0x42000A), Type: TextString (0x07), Data: <redacted>
          Tag: Attribute Value (0x42000B), Type: Integer (0x02), Data: <redacted>
```

Operation code 2 is Create Key Pair according to the KMIP specification (see KMIP 1.0 section 9.1.3.2.26 [Operation Enumeration](https://docs.oasis-open.org/kmip/spec/v1.0/os/kmip-spec-1.0-os.html#_Toc236497894)).

The second log line, the response, renders like this:

```
❯ cargo run --example diag_to_txt /tmp/r2
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/examples/diag_to_txt /tmp/r2`
Tag: Response Message (0x42007B), Type: Structure (0x01), Data: 
  Tag: Response Header (0x42007A), Type: Structure (0x01), Data: 
    Tag: Protocol Version (0x420069), Type: Structure (0x01), Data: 
      Tag: Protocol Version Major (0x42006A), Type: Integer (0x02), Data: <redacted>
      Tag: Protocol Version Minor (0x42006B), Type: Integer (0x02), Data: <redacted>
    Tag: Time Stamp (0x420092), Type: DateTime (0x09), Data: <redacted>
    Tag: Batch Count (0x42000D), Type: Integer (0x02), Data: <redacted>
  Tag: Batch Item (0x42000F), Type: Structure (0x01), Data: 
    Tag: Operation (0x42005C), Type: Enumeration (0x05), Data: 2
    Tag: Result Status (0x42007F), Type: Enumeration (0x05), Data: 0
    Tag: Response Payload (0x42007C), Type: Structure (0x01), Data: 
      Tag: Private Key Unique Identifier (0x420066), Type: TextString (0x07), Data: <redacted>
      Tag: Public Key Unique Identifier (0x42006F), Type: TextString (0x07), Data: <redacted>
```

This shows a successful response that includes the internal PyKMIP identifiers for the created keys (which are redacted in the diagnostic output).


src/commons/crypto/signing.rs
```diff
@@ -24,260 +26,321 @@ use rpki::{
...
-    // actually needs to be locked raet
-
-    // The general signer is used for all signing operations except one off signing.
-    general_signer: Arc<RwLock<SignerProvider>>,
-
-    // As the security of a HSM isn't needed for one off keys, and HSMs are slow, by default this should be an instance
-    // of OpenSslSigner. However, if users think the perceived extra security is warranted let them use a different
-    // Signer for one off keys if that's what they want.
-    one_off_signer: Arc<RwLock<SignerProvider>>,
+    dispatcher: Arc<RwLock<SignerRouter>>,
```
ximon18, line 273: A couple of issues with this:
- Firstly, while nameshedding I seem to have become conflicted and used both dispatcher and router. I'm not overly fond of either. The static dispatching could be said to happen in the SignerProvider where it matches on the enum value to "dispatch" to the correct code, not in the SignerRouter. Also the SignerRouter, even if it is considered to do dispatching, doesn't just dispatch, it makes choices about where to dispatch to hence the name "router: rather than dispatcher. However having the name "router" in a routing product for some unrelated purpose makes me uncomfortable as well...

- The RwLock on the whole SignerRouter means that creating a key in a KMIP backend will prevent any action on any other signer at the same time even though they cannot conflict with each other. Ideally there wouldn't even be a RwLock on the individual signers but the use of &mut in the `Signer` trait means we need a lock or else the `mut` leaks out to the owning struct requiring the whole "struct" to be locked at once.

ximon18: I pushed a commit to use consistent naming and to remove the unecessary `Arc<RwLock<..>>` around the `SignerRouter`.

src/commons/crypto/signers/kmip/internal.rs
```diff
@@ -0,0 +1,776 @@
...
+        let kmip_key_ids = KmipKeyPairIds {
+            public_key_id,
+            private_key_id,
+        };
+
+        Ok(kmip_key_ids)
+    }
+
+    /// Make the given KMIP private and public key pair ready for use by Krill.
+    fn prepare_keypair_for_use(&self, private_key_id: &str, public_key_id: &str) -> Result<PublicKey, SignerError> {
```
timbru, line 536: why do we rename things using the key_id? I mean.. ultimately we have another thing mapping key_id to the actual keys and what not, so would it not be easier (more efficient in particular) to just have something unique like a uuid?

src/commons/crypto/signers/kmip/internal.rs
```diff
@@ -0,0 +1,776 @@
...
+        let signed = self
+            .with_conn("sign", |conn| conn.sign(&private_key_id, data))
+            .map_err(|err| SignerError::KmipError(format!("Signing failed: {}", err)))?;
+
+        let sig = Signature::new(SignatureAlgorithm::default(), Bytes::from(signed.signature_data));
+
+        Ok(sig)
+    }
+
+    pub fn destroy_key_pair(&self, kmip_key_pair_ids: &KmipKeyPairIds, mode: KeyStatus) -> Result<bool, SignerError> {
```
timbru, line 680: I wonder if destroy_key should be idempotent. I.e. maybe it should be okay if the key does not exist?

timbru (approved):

src/commons/crypto/signers/kmip/internal.rs
```diff
@@ -0,0 +1,776 @@
...
+        error::SignerError,
+        kmip::{
+            connpool::ConnectionManager,
+            keymap::{KeyMap, KmipKeyPairIds},
+        },
+        util,
+    },
+};
+
+//------------ Types and constants ------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
ximon18: Indeed. One of them, IGNORE_MISSING_CAPABILITIES, is promoted to a server config setting in a later PR, because it needs to be configurable. Given that Krill lets you tweak lots of settings it wouldn't be odd to add more settings to the config file, and it's hard to pick "correct" values for these settings which implies that we or a customer may need to tweak them. But this PR didn't define anything in config because config file support was left for a later PR. However it might be good to create a tracking issue for making these settings configurable.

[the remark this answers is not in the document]
timbru: ok, created #726

ximon18:
@timbru : Am I right in thinking that your review didn't including looking at the `kmip-protocol` crate dependency code or the `kmip-ttlv` crate that it depends on, both of which were written by me and whose code is also in NLnet Labs GitHub repositories?

timbru:
> @timbru : Am I right in thinking that your review didn't including looking at the `kmip-protocol` crate dependency code or the `kmip-ttlv` crate that it depends on, both of which were written by me and whose code is also in NLnet Labs GitHub repositories?

Indeed. I am just focussing on krill itself here - I am happy to look at the kmip libs as a separate effort.

doc/development/hsm/architecture.md
```diff
@@ -1,22 +1,97 @@
...
+Prior to the addition of HSM support there was only ever a single Signer and control flow looked like this:
+
+```
+Krill calling code -> KrillSigner -> OpenSslSigner
+```
+
+With the addition of HSM support there may be multiple concurrently active Signers and the control flow becomes this:
+
+```
+Krill calling code -> KrillSigner -> SignerRouter -> SignerProvider -> One of: OpenSslSigner, KmipSigner or Pkcs11Signer.
```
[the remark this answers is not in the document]
timbru: ok

src/commons/crypto/signers/kmip/internal.rs
timbru: ok, we can look at this later - and see if it impacts performance as well

src/commons/crypto/signers/kmip/internal.rs
timbru: I think that removing a key which is already removed is probably not an issue, but.. key cleanups need work. Created an issue for it: https://github.com/NLnetLabs/krill/issues/728

src/commons/crypto/signers/kmip/internal.rs
ximon18: Speaking of performance we need to cache the results of key lookup.

PR #686, opened 2021-10-12
HSM 2: Persistent signer key mappings

ximon18:
This PR builds on https://github.com/NLnetLabs/krill/pull/679 "KMIP walking skeleton". The purpose of this PR is to add support for remembering which signer backend possesses a given Krill `KeyIdentifier` and which internal HSM key identifier(s) correspond to a given Krill `KeyIdentifier`.

This PR has been marked ready for review because PRs that come later and depend on this PR have been marked ready for review. As it stands as an apparently suitable base for later PRs it is apparently usable to build upon and should thus be reviewed.

As noted in the comment below this PR moves files around which makes it harder to review. With the git defaults it shows files as added and deleted which have actually been moved and edited. Below I try to guide you through the changes to make it easier to review this PR.

# Easy changes to review

- `.github/workflows/ci.yml`: extends the HSM tests to test feature flag `hsm` in isolation, not just when used with the `hsm-tests` flag, just to make sure the build isn't broken.

- `Cargo.lock` and `Cargo.toml`: bump the KMIP dependency to a newer version, and some ease of reading improvements were made to the order of feature flags and with some minor additional comments.

- `doc/development/hsm/xxx.md`: files were updated. However, it might be worth skipping ahead to the versions in https://github.com/NLnetLabs/krill/pull/689 which have had considerably more work done on them.

- `src/commons/api/ca.rs`: minor tweaks to the tests to pass additional arguments into `OpenSslSigner::build()` (a signer name used for logging/error reporting to distinguish reporting by one signer from another when running multiple concurrent signers, and an optional SignerMapper instance which is required when using more than one signer so that the owning signer for a key can be looked up, but in a test that uses only a single OpenSslSigner the SignerMapper can be None as the one signer is being used directly by the test, there is no signer dispatch involved).

- `signers/kmip/connpool.rs`, `signers/kmip/signer.rs`, `signers/util.rs`: Unchanged files moved from `src/commons/crypto/` to `src/commons/crypto/signing/`

- `src/commons/crypto/mod.rs`, `src/commons/error.rs`: Minor updates related to moved modules.

- `signers/kmip/mod.rs` and `signers/kmip/keymap.rs`: `keymap.rs` was deleted and removed from the module (as this PR replaces it with something else)

- `signers/errors.rs`: Adds some new signer error variants. I'm not happy about the added "Other(String)", I think it was added for a case where there is no context to give a better error but it feels ... undesirable. Note that later PR #689 renames the other two variants from `SignerUnavailable` -> `TemporarilyUnavailable` and from `SignerUnusable` -> `PermanentlyUnusable`.

- `src/constants.rs`: Added a named constant for an existing hardcoded path and added a new constant for the new '<work_dir>/signers` directory that the new signer mapper mechanism will store signer and key mapping data in.

# The new signer and key mapping mechanism

`src/commons/crypto/signing/dispatch/signerinfo.rs` adds a new event sourcing aggregate store for storing details about known signers and the keys that they own and the mapping for those keys from Krill `KeyIdentifier` to internal signer identifier(s).

It was based on the test aggregate store defined in `src/commons/eventsourcing/mod.rs`. There are a lot of structs but this is just how the event sourcing code in Krill requires it to be. In essence it's actually quite basic: there is a new high level struct `SignerMapper` that is used by the signing functionality to:

- Record new signers and their identity details.
- Record that a particular Krill key is owned by a given signer and corresponds to a particular signer specific internal key identifer.
- Change the name and/or info of a particular signer.
- Query the set of signers.
- Query which signer owns a given Krill key.

The identity details for a signer consist of three parts:

- A human readable name intended to be set by the operator in the config file which can be changed later if discovered to be incorrect or misleading or outdated.
- Some arbitrary info string intended to be populated with details about the signer backend retrieved from the backend itself, e.g. its make and model and anything other interesting and/or identifying properties.
- A public key.

The public key requires some more explanation.

In the HSM prototype code signers were mapped to KeyIdentifiers and signer backend internal key IDs in a simple CSV file structure. This added new storage engine code to Krill rather than re-using its existing storage abilities, code that would need testing and documenting and wouldn't benefit from any later extension to the core storage engine to for example work with a networked key value store instead of local disk files. The CSV file was indexed by signer name, so if the operator edited the name of the signer in the config file then Krill would no longer be able to locate which signer owned a given key.

This version of the code changes this in several ways:
- Use the existing Krill storage engine instead of inventing a new one.
- Record backend signer info and benefit from the store tracking those changes over time, e.g. you can later see that the signer backend was upgraded or changed in some way.
- Detach the signer records from the operator given name and associate them with a unique handle instead so that the name can be changed.
- Don't require signers to be online and usable when Krill starts up.
- Do common checks on signer backends to verify that they work with Krill.
- Determine which signer backend corresponds to which signer configuration without requiring the operator to keep names or signer order static or to manage an ID per signer.

These last twos point are where the public key comes in. We want anyway to know if a signer backend is actually capable of creating a signing key and signing data with it correctly and to do this we need to ask the backend to create a new key pair which we would then discard, assuming we didn't hit a problem trying to delete this temporary key.

Instead of discarding the created key we can remember the internal ID of the private key and save the public key. When we see the contact the signer backend on a subsequent restart of Krill we attempt to locate the signing key for each known signer. Internal IDs might not be unique, e.g. they could be integers that rise by one, so finding that the signer backend says it has a particular key isn't proof that it owns the key that we created in it earlier. So we ask  the signer to sign a challenge using that private key and we verify the created signature using the stored public key. In this way we can determine if a particular signer backend is the one a subset of our signer store records relate to, i.e. that we think it owns certain keys that we created earlier.

With this the operator doesn't have to correctly manage signer names, IDs or order of definition, and we verify that the signer can create keys and sign with them. It also doesn't matter if the signer backend is upgraded or replaced entirely by a new one or if the means of connecting to it changes from KMIP to PKCS#11 or connection details or credentials change, if it has our keys we will be able to detect that. Admittedly we assume that not just the one identity key was migrated to a new signer or that other keys were not deleted by an operator, but that is always outside of our control.

- `commons/crypto/signing.rs`: KrillSigner was moved and split out into `commons/crypto/signing/dispatch/krillsigner.rs` and other structs such as the SignerRouter and SignerProvider were created in new files in the same module.

  PR https://github.com/NLnetLabs/krill/pull/689 greatly updates the developer `docs/xxx.md` files relating to these changes and is worth reading at this point.

- `commons/crypto/signing/dispatch/signerrouter.rs`: Logic to discover signers and register them in the SignerMapper (defined in `signerrouter.rs`) and to dispatch requests to the signer that owns a key (by looking it up in the SignerMapper).

- `commons/crypto/signing/dispatch/signerprovider.rs`: An enum based dispatch wrapper around multiple different signer implementations.

- `commons/crypto/signers/softsigner.rs`: The OpenSSL soft signer was updated to support recording of keys that it creates in the `SignerMapper` and to implement new functions relating to creating the identity key and signing a challenge using it. Code to create a key was factored out into new function `build_key()` so that it could also be used to create the identity key.

- `commons/crypto/signers/kmip/internal.rs`: Updated to support creation of the new identity key and signing a challenge and to use the new `SignerMapper` instead of the simple `KeyMap` placeholder in-memory mechanism.


ximon18:
The `tests/suspend.rs` test is sometimes failing when the `hsm` feature is active, with failures seen both with and without the `hsm-tests` feature activated.

ximon18:
I increased the time out and the next GH Actions run succeeded. Ideally this would be fixed by checking for the expected conditions rather than sleeping for a period of time but that's outside the scope of this PR.

src/commons/crypto/signing/dispatch/signerrouter.rs
```diff
@@ -0,0 +1,676 @@
...
+                return Ok(IdentifyResult::Unusable);
+            }
+            Ok(res) => res,
+        };
+
+        if public_key.verify(challenge, &signature).is_ok() {
+            debug!("Signer '{}' is ready and known, binding", signer_name);
+            let signer_info = signer_provider.get_info().unwrap_or("No signer info".to_string());
+
+            // Drop the read lock so that we can acquire a write lock
```
timbru, line 549: This does not seem to do anything - clippy tells me that this is dropping a reference only. Furthermore it used again straight after this... so this feels a bit wrong.

Generally speaking I prefer to use scoping when it comes to locks.. if you acquire a lock in a function and want to release it later within the same function then you can wrap part of the function in a scoped block to enforce this.

src/commons/crypto/signing/dispatch/signerrouter.rs
```diff
@@ -0,0 +1,676 @@
...
+}
+
+#[cfg(not(feature = "hsm"))]
+impl SignerRouter {
+    fn bind_ready_signers(&self) {}
+}
+
+#[cfg(feature = "hsm")]
+impl SignerRouter {
+    fn bind_ready_signers(&self) {
```
timbru, line 333: Right, so after digging around the code a bit I think that the is used this way and called at runtime to deal with HSMs that may not be available at startup - and that furthermore this reports an error, but does not return with an error, to prevent that one failing signer makes all other signers unavailable.

Perhaps add a documentation comment to this effect to this fn?

timbru (approved):

src/commons/crypto/signing/dispatch/signerinfo.rs
```diff
@@ -0,0 +1,436 @@
...
+        Ok(())
+    }
+
+    pub fn get_signer_name(&self, signer_handle: &Handle) -> KrillResult<String> {
+        Ok(self.store.get_latest(signer_handle)?.signer_name.clone())
+    }
+
+    pub fn change_signer_name(&self, signer_handle: &Handle, signer_name: &str) -> KrillResult<()> {
+        // TODO: should version be something other than None here?
+        let cmd = SignerInfoCommand::change_signer_name(signer_handle, None, signer_name);
```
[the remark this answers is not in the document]
ximon18: Hmm, then as `fn add_key()` **is** called at runtime your comment seems to indicate that where it uses `Nonw` that is actually not correct, right?

src/commons/crypto/signing/dispatch/signerrouter.rs
ximon18: Without the drop you get a deadlock. I also prefer blocks to scope lock acquisition and release, I'll look at the code again but in this case if I recall correctly it was proving a bit of pain to achieve the desired result in a readable way using block scopes.

src/commons/crypto/signing/dispatch/signerrouter.rs
ximon18: Does this help?

src/commons/crypto/signing/dispatch/signerrouter.rs
ximon18: Ah. This code was changed, it used to use a RwLock but that was removed when &mut was dropped from the Signer trait by a new release of the rpki-rs crate. The drop is indeed now redundant!

src/commons/crypto/signing/dispatch/signerrouter.rs
timbru: created an issue, we can do this later: https://github.com/NLnetLabs/krill/issues/729

src/commons/crypto/signing/dispatch/signerinfo.rs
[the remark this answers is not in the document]
timbru: no, this is still fine - it should just not be called concurrently - ideally - but even that would be fine in this case.

PR #688, opened 2021-10-14
Use new de-mut'd rpki-rs Signer trait.

ximon18:
This PR builds on https://github.com/NLnetLabs/krill/pull/686 "HSM persistent signer key mappings". Uses a fork of the `rpki-rs` crate `reverse-signer` branch as I need it to depend on bcder 0.6.1-dev for my Krill branch to compile. I want to use the de-mut'd Signer trait change from the new rpki-rs crate version, when it gets released at some point, as it removes a lot of boiler plate and unnecessary locking from my multi-signer code.

_**Update**: This PR has been updated to use the new official releases of the `bcder` and `rpki-rs` crates rather than personal forks._


ximon18:
This PR is now redundant as the changes it introduced have been introduced via PR #702 into the upstream branches of this branch.

PR #702, opened 2021-11-01
Aspa poc

timbru:
Still needs the updated ASPA oid in rpki-rs (depends on a branch now).

Propose to squash when merging to reduce the noise. Best to look at the resulting file changes rather than per commit.

### DIFF
--- a/src/cli/client.rs
+++ b/src/cli/client.rs
@@ -331,6 +331,12 @@ impl KrillClient {
                 Ok(ApiResponse::BgpAnalysisSuggestions(suggestions))
             }
 
+            CaCommand::AspasAdd(handle, aspa) => {
+                let uri = format!("api/v1/cas/{}/aspas", handle);
+                post_json(&self.server, &self.token, &uri, aspa).await?;
+                Ok(ApiResponse::Empty)
+            }
+
             CaCommand::Show(handle) => {
                 let uri = format!("api/v1/cas/{}", handle);
                 let ca_info = get_json(&self.server, &self.token, &uri).await?;
--- a/src/cli/client.rs
+++ b/src/cli/client.rs
@@ -332,6 +332,12 @@ impl KrillClient {
                 Ok(ApiResponse::BgpAnalysisSuggestions(suggestions))
             }
 
+            CaCommand::AspasList(handle) => {
+                let uri = format!("api/v1/cas/{}/aspas", handle);
+                let aspas = get_json(&self.server, &self.token, &uri).await?;
+                Ok(ApiResponse::AspaDefinitions(aspas))
+            }
+
             CaCommand::AspasAddOrReplace(handle, aspa) => {
                 let uri = format!("api/v1/cas/{}/aspas", handle);
                 let updates = AspaDefinitionUpdates::new(vec![aspa], vec![]);
--- a/src/cli/client.rs
+++ b/src/cli/client.rs
@@ -339,6 +339,13 @@ impl KrillClient {
                 Ok(ApiResponse::Empty)
             }
 
+            CaCommand::AspasRemove(handle, customer) => {
+                let uri = format!("api/v1/cas/{}/aspas", handle);
+                let updates = AspaDefinitionUpdates::new(vec![], vec![customer]);
+                post_json(&self.server, &self.token, &uri, updates).await?;
+                Ok(ApiResponse::Empty)
+            }
+
             CaCommand::AspasUpdate(handle, customer, update) => {
                 let uri = format!("api/v1/cas/{}/aspas/as/{}", handle, customer);
                 post_json(&self.server, &self.token, &uri, update).await?;
--- a/src/cli/client.rs
+++ b/src/cli/client.rs
@@ -11,8 +11,9 @@ use crate::{
     },
     commons::{
         api::{
-            AllCertAuthIssues, CaRepoDetails, CertAuthIssues, ChildCaInfo, ChildrenConnectionStats, ParentCaContact,
-            ParentStatuses, PublisherDetails, PublisherList, RepoStatus, Token,
+            AllCertAuthIssues, AspaDefinitionUpdates, CaRepoDetails, CertAuthIssues, ChildCaInfo,
+            ChildrenConnectionStats, ParentCaContact, ParentStatuses, PublisherDetails, PublisherList, RepoStatus,
+            Token,
         },
         bgp::BgpAnalysisAdvice,
         error::KrillIoError,
@@ -331,9 +332,10 @@ impl KrillClient {
                 Ok(ApiResponse::BgpAnalysisSuggestions(suggestions))
             }
 
-            CaCommand::AspasAdd(handle, aspa) => {
+            CaCommand::AspasAddOrReplace(handle, aspa) => {
                 let uri = format!("api/v1/cas/{}/aspas", handle);
-                post_json(&self.server, &self.token, &uri, aspa).await?;
+                let updates = AspaDefinitionUpdates::new(vec![aspa], vec![]);
+                post_json(&self.server, &self.token, &uri, updates).await?;
                 Ok(ApiResponse::Empty)
             }
 
--- a/src/cli/client.rs
+++ b/src/cli/client.rs
@@ -337,6 +337,12 @@ impl KrillClient {
                 Ok(ApiResponse::Empty)
             }
 
+            CaCommand::AspasUpdate(handle, customer, update) => {
+                let uri = format!("api/v1/cas/{}/aspas/as/{}", handle, customer);
+                post_json(&self.server, &self.token, &uri, update).await?;
+                Ok(ApiResponse::Empty)
+            }
+
             CaCommand::Show(handle) => {
                 let uri = format!("api/v1/cas/{}", handle);
                 let ca_info = get_json(&self.server, &self.token, &uri).await?;
--- a/src/cli/options.rs
+++ b/src/cli/options.rs
@@ -735,7 +735,7 @@ impl Options {
 
     #[cfg(feature = "aspa")]
     fn make_cas_aspas_add_sc<'a, 'b>(app: App<'a, 'b>) -> App<'a, 'b> {
-        let mut sub = SubCommand::with_name("add").about("Add an ASPA configuration");
+        let mut sub = SubCommand::with_name("add").about("Add or replace an ASPA configuration");
 
         sub = Self::add_general_args(sub);
         sub = Self::add_my_ca_arg(sub);
@@ -1787,7 +1787,7 @@ impl Options {
         let aspa_config_str = matches.value_of("aspa").unwrap(); // required argument
         let aspa = AspaDefinition::from_str(aspa_config_str)?;
 
-        let command = Command::CertAuth(CaCommand::AspasAdd(my_ca, aspa));
+        let command = Command::CertAuth(CaCommand::AspasAddOrReplace(my_ca, aspa));
 
         Ok(Options::make(general_args, command))
     }
@@ -2307,7 +2307,7 @@ pub enum CaCommand {
     BgpAnalysisSuggest(Handle, Option<ResourceSet>),
 
     // ASPAs
-    AspasAdd(Handle, AspaDefinition),
+    AspasAddOrReplace(Handle, AspaDefinition),
     AspasUpdate(Handle, AspaCustomer, AspaConfigurationUpdate),
 
     // Show details for this CA
--- a/src/cli/options.rs
+++ b/src/cli/options.rs
@@ -12,7 +12,7 @@ use bytes::Bytes;
 use clap::{App, Arg, ArgMatches, SubCommand};
 
 use rpki::{
-    repository::{crypto::KeyIdentifier, x509::Time},
+    repository::{aspa::ProviderAs, crypto::KeyIdentifier, x509::Time},
     uri,
 };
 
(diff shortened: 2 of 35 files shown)


src/cli/client.rs
```diff
@@ -331,6 +332,32 @@ impl KrillClient {
                 Ok(ApiResponse::BgpAnalysisSuggestions(suggestions))
             }
 
+            CaCommand::AspasList(handle) => {
```
ximon18, line 335: For ROAs you used the full name e.g. `RouteAuthorizationsUpdate` but here you use the abbreviation ASPA. This is inconsistent. Is this deliberate because names like `AutonomousSystemProviderAuthorizationsUpdate` was felt to be too long or is this inconsistency accidental?

src/cli/client.rs
```diff
@@ -331,6 +332,32 @@ impl KrillClient {
...
+                let uri = format!("api/v1/cas/{}/aspas", handle);
+                let aspas = get_json(&self.server, &self.token, &uri).await?;
+                Ok(ApiResponse::AspaDefinitions(aspas))
```
ximon18, line 338: Same question as previous comment, but this time relating to `Aspa` types themselves rather than variant names of this type.

ximon18: I realize now that of course there is also the question of REST relative URL naming, and again for ROAs that was `/routes` but for ASPAs `/aspas` has been used, so same question here too.

src/cli/client.rs
timbru: Indeed, this is deliberate and writing it out would be too long.

Additionally.. for ROAs I try to talk about "RouteAuthorization" when it comes to configured prefixes which are to be put on one or *more* ROAs, potentially together with other configured prefixes. So there is a many-many mapping going on and using "Roa" there would be more confusing because typically ROA is the *object*.

For ASPA that is slightly different because all providers for a 'customer' ASN MUST (RFC) always appear on a single object. Even if you would get multiple ASPA objects if the customer ASN is held in more than one resource class.. this is much like a one-many mapping.

src/cli/client.rs
timbru: same answer..

Because many (authorized) 'routes' can end up on a single ROA. But an ASPA config maps straight to an ASPA object.

src/cli/options.rs
```diff
@@ -1710,6 +1807,90 @@ impl Options {
...
+                let provider = ProviderAs::from_str(provider_str)
+                    .map_err(|_| Error::invalid_asn(provider_str))?;
+                removed.push(provider);
+            }
+        }
+
+        let update = AspaConfigurationUpdate::new(added, removed);
+        if update.is_empty() {
+            return Err(Error::general("You MUST specify at least one of --add or --remove"));
+        }
```
ximon18, line 1864: Do you also need to check for the same AS being added and removed at the same time?

src/cli/report.rs
```diff
@@ -30,11 +31,16 @@ pub enum ApiResponse {
...
     CertAuthAction(CaCommandDetails),
     CertAuths(CertAuthList),
+
+    // ROA related
     RouteAuthorizations(RoaDefinitions),
     BgpAnalysisAdvice(BgpAnalysisAdvice),
     BgpAnalysisFull(BgpAnalysisReport),
     BgpAnalysisSuggestions(BgpAnalysisSuggestion),
 
+    // ASPA
```
ximon18, line 41: " related"?

timbru: ack, done

src/cli/options.rs
timbru: good idea, added that.. a bit higher up when looping through the removals

src/commons/api/aspa.rs
```diff
@@ -0,0 +1,370 @@
...
+            // Ensure that the providers are sorted, there is at least one, and there are no duplicates
+            providers.sort_by_key(|p| p.provider());
+
+            let mut last_seen = providers
+                .first()
+                .ok_or(AspaConfigurationFormatError::ProvidersMissing)?;
+
+            if providers.len() > 1 {
+                for i in 1..providers.len() {
+                    let next = providers.get(i).unwrap(); // safe i goes until .len()
```
ximon18, line 184: Doesn't i go until `.len()-1` because you used `a..b` syntax rather than `a..=b` syntax?

timbru: i should fix the comment - I meant until but not including - which is correct behaviour - the last index number *is* len()-1. Think of an array with 1 entry as an example.. the index for that item is 0. Since I test that the len()>1 I can do all this (start at 1) with fear of  out of bounds.

src/commons/api/aspa.rs
```diff
@@ -0,0 +1,370 @@
...
+    ExtraParts,
+}
+
+impl fmt::Display for AspaConfigurationFormatError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "ASPA configuration format invalid: ")?;
+        match self {
+            AspaConfigurationFormatError::CustomerMissing => write!(f, "customer AS missing"),
+            AspaConfigurationFormatError::CustomerInvalid(s) => write!(f, "cannot parse customer AS: {}", s),
+            AspaConfigurationFormatError::ProvidersMissing => write!(f, "providers missing"),
```
ximon18, line 213: In the other three messages you don't include AS in the variant name but you do mention AS in the message, but for this message you do not mention AS. For consistency should you also mention AS here?

src/commons/api/aspa.rs
```diff
@@ -0,0 +1,370 @@
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
```
ximon18, line 345: Remove this blank line?

src/commons/api/history.rs
```diff
@@ -720,6 +744,19 @@ impl fmt::Display for StorableCaCommand {
...
             }
 
+            // ------------------------------------------------------------
+            // ASPA Support
+            // ------------------------------------------------------------
+            StorableCaCommand::AspasUpdate { updates } => {
+                write!(f, "{}", updates)
+            }
+            StorableCaCommand::AspasUpdateExisting { customer, update } => {
+                write!(f, "update ASPA for customer ASN: {} {}", customer, update)
```
ximon18, line 754: Here you refer to ASN in the message, while above you use AS in messages. Should they be consistent?

src/commons/api/history.rs
```diff
@@ -566,9 +581,18 @@ impl WithStorableDetails for StorableCaCommand {
...
+
+            // ROA
             StorableCaCommand::RoaDefinitionUpdates { updates } => CommandSummary::new("cmd-ca-roas-updated", &self)
                 .with_added(updates.added().len())
                 .with_removed(updates.removed().len()),
+
+            // ASPA
+            StorableCaCommand::AspasUpdate { .. } => CommandSummary::new("cmd-ca-aspas-update", &self),
+            StorableCaCommand::AspasUpdateExisting { .. } => CommandSummary::new("cmd-ca-aspas-update-existing", &self),
+            StorableCaCommand::AspaRemove { .. } => CommandSummary::new("cmd-ca-aspa-remove", &self),
```
ximon18, line 593: The other two commands above use "aspas-" but this one uses "aspa-". Should they be the same for consistency?

src/commons/error.rs
```diff
@@ -423,7 +440,16 @@ impl fmt::Display for Error {
             Error::CaAuthorizationDuplicate(_ca, roa) => write!(f, "ROA '{}' already present", roa),
             Error::CaAuthorizationInvalidMaxLength(_ca, roa) => write!(f, "Invalid max length in ROA: '{}'", roa),
             Error::CaAuthorizationNotEntitled(_ca, roa) => write!(f, "Prefix in ROA '{}' not held by you", roa),
-            Error::RoaDeltaError(e) => write!(f, "ROA delta rejected:\n\n'{}' ", e),
+            Error::RoaDeltaError(_ca, e) => write!(f, "ROA delta rejected:\n\n'{}' ", e),
+
+            //-----------------------------------------------------------------
+            // Autonomous System Provider Authorization - ASPAs
+            //-----------------------------------------------------------------
+            Error::AspaCustomerAsNotEntitled(_ca, asn) => write!(f, "Customer ASN '{}' is not held by you", asn),
```
ximon18, line 448: Some of the message content above used AS rather than ASN. Should these be made consistent?

src/commons/error.rs
```diff
@@ -787,10 +812,28 @@ impl Error {
...
-                ErrorResponse::new("ca-roa-delta-error", "Delta rejected, see included json")
-                    .with_roa_delta_error(roa_delta_error)
-            }
+            Error::RoaDeltaError(ca, roa_delta_error) => ErrorResponse::new("ca-roa-delta-error", &self)
+                .with_ca(ca)
+                .with_roa_delta_error(roa_delta_error),
+
+            //-----------------------------------------------------------------
+            // Autonomous System Provider Authorization - ASPA
+            //-----------------------------------------------------------------
```
ximon18, line 821: Above there are some "ca-aspas-" style names used, here it is "ca-aspa-", should these be consistent?

src/commons/error.rs
```diff
@@ -1164,20 +1208,19 @@ mod tests {
         let not_held = definition("10.128.0.0/9 => 1");
         let invalid_length = definition("10.0.1.0/25 => 1");
         let unknown = definition("192.168.0.0/16 => 1");
-
+
```
ximon18, line 1211: Shouldn't cargo fmt have removed this whitespace?

src/daemon/auth/common/permissions.rs
```diff
@@ -67,6 +67,9 @@ iterable_enum! {
         ROUTES_READ,
         ROUTES_UPDATE,
         ROUTES_ANALYSIS,
+        ASPAS_READ,
```
ximon18, line 70: It's been a while since I looked at the code but I assume we need to add these new permissions to `defaults/roles.polar`.

src/daemon/ca/aspa.rs
```diff
@@ -0,0 +1,280 @@
...
+        KrillResult,
+    },
+    daemon::{
+        ca::{AspaObjectsUpdates, CertifiedKey},
+        config::{Config, IssuanceTimingConfig},
+    },
+};
+
+pub fn make_aspa_object(
+    aspa_config: AspaDefinition,
```
ximon18, line 32: Given that there is also an `AspaConfigurationUpdate` type I find it confusing here that `aspa_config` is the name of something which is of type `AspaDefinition`. Which of the two names is "incorrect" here? config or definition?

src/daemon/ca/aspa.rs
```diff
@@ -0,0 +1,280 @@
...
+
+    /// Issue new ASPA objects based on configuration, and remove
+    /// object for which the customer AS is no longer held.
+    ///
+    /// Note: we pass in *all* AspaDefinitions for the CA, not all
+    ///   definitions will be relevant for the RC (key) holding
+    ///   this AspaObjects.
+    pub fn update(
+        &self,
+        all_aspas: &AspaDefinitions,
```
ximon18, line 156: As neither `all_aspas` nor `relevant_aspa` are actually `Aspa` instances it would be less confusing to include `def` or `definition` in these variable names as they are instances of the `AspaDefinition` type.

src/daemon/ca/events.rs
```diff
@@ -977,6 +1048,14 @@ impl fmt::Display for CaEvtDet {
                 Ok(())
             }
 
+            // Autonomous System Provider Authorization
+            CaEvtDet::AspaConfigAdded { aspa_config: addition } => write!(f, "{}", addition),
+            CaEvtDet::AspaConfigUpdated { customer, update } => {
+                write!(f, "updated ASPA config for customer ASN: {} {}", customer, update)
+            }
+            CaEvtDet::AspaConfigRemoved { customer } => write!(f, "removed ASPA config for customer ASN: {}", customer),
+            CaEvtDet::AspaObjectsUpdated { .. } => todo!(),
```
ximon18, line 1057: ?

src/daemon/ca/rc.rs
```diff
@@ -683,6 +706,37 @@ impl ResourceClass {
     }
 }
 
+/// # Autonomous System Provider Authorization
+///
+impl ResourceClass {
+    /// Renew all ROAs under the current for which the not-after time closer
```
ximon18, line 712: s/not-after time closer/not-after time is closer/ ?

ximon18: `/// Renew all ROAs` but the function name is `renew_aspas()`, i.e. this is not about ROAs?

tests/functional.rs
```diff
@@ -331,12 +336,15 @@ async fn functional() {
...
+        for aspa in aspas {
+            expected_files.push(ObjectName::aspa(aspa.customer()).to_string());
+        }
         assert!(will_publish_embedded(test_msg, &ca4, &expected_files).await);
     }
 
     // short hand to expect ROAs under CA4, re-added when parent comes back
     // i.e. it now has RC 2 and 3, but no more 0 and 1
-    async fn expect_roas_for_ca4_re_added(test_msg: &str, roas: &[RoaDefinition]) {
+    async fn expect_roas_for_ca4_re_added(test_msg: &str, roas: &[RoaDefinition], aspas: &[AspaDefinition]) {
```
ximon18, line 347: s/expect_roas/expect_aspas/ ?

ximon18 (approved):
The Rust changes seem okay. I can't say if it's doing exactly what should be done for ASPAs though it seems logical. I have **NOT** checked the code base to see if there are other places that ROAs are worked with that should also be updated to work with ASPAs.

src/commons/api/aspa.rs
timbru: fixed

src/commons/api/history.rs
timbru: trying to use AS consistently

src/daemon/ca/aspa.rs
timbru: Using 'definition' now wherever 'configuration' was used.

src/daemon/ca/aspa.rs
timbru: done

src/commons/api/history.rs
timbru: done

src/commons/error.rs
timbru: It's consistent with how things were named for roas/roa. If we make this all consistent across the code base, then we will need to do data migrations (update keys) and it may result in breaking changes in the API with regards to error labels - although I don't think people actually use this.

src/daemon/ca/events.rs
timbru: oops, fixed

src/daemon/ca/rc.rs
timbru: fixed

tests/functional.rs
timbru: fixed

src/daemon/auth/common/permissions.rs
timbru: added

Issue #564, opened 2021-06-10
Manual KMIP testing with PyKMIP


ximon18:

ximon18:
Tested using Krill commit eb97735 (the current head of the `issue-698-record-keys-on-upgrade` branch) and PyKMIP 0.10.0 on Ubuntu 21.10, with PyKMIP configuration and startup script taken from the `test-resources/pykmip/` directory in that Krill commit.

Krill was installed like so:

```
$ cargo install --git https://github.com/NLnetLabs/krill --rev eb97735 --locked --features hsm
```

The Krill config file looked like this:
```toml
admin_token = "abc"
data_dir = "/tmp/krill"
service_uri = "https://localhost:3000/"
log_level = "trace"
log_type = "stderr"

[[signers]]
type = "KMIP"
host = "localhost"
server_ca_cert_path = "<path/to/krill/git/checkout>/test-resources/pykmip/ca.crt"
client_cert_path = "<path/to/krill/git/checkout>/test-resources/pykmip/server.crt"
client_cert_private_key_path = "<path/to/krill/git/checkout>/test-resources/pykmip/server.key"
deficient = true
```

**Note:** `deficient = true` is set because PyKMIP 0.10.0 reports that it is unable to modify attributes but when the `ModifyAttribute` operation is attempted it succeeds so we ignore the claim that it is unsupported, otherwise Krill would refuse to use the PyKMIP server.

Run PyKMIP in one terminal:
```
$ cd <path/to/krill/git/checkout>/test-resources/pykmip
$ ./run-server.py
```

Run Krill in another terminal:
```
$ ~/.cargo/bin/krill -c /tmp/krill.conf
```

Krill configures itself to use the KMIP signer for and a fallback OpenSSL signer:
```
[INFO] ... Configuring signer 'KMIP' (type: KMIP, default: true, oneoff: false, random: false)
[INFO] ... Configuring signer 'Fallback OpenSSL signer' (type: OpenSSL, default: false, oneoff: true, random: true)
```

In a third terminal use `krillc` to create a CA:

```
$ export KRILL_CLI_TOKEN=abc
$ ~/cargo/bin/krillc add --ca some_ca
```

On creation of a CA, Krill attempts to initialize the signers and contact the KMIP server:

```
[WARN] ... [KMIP] Ignoring KMIP server lacking support for one or more required operations: ModifyAttribute
[...]

Issue #553, opened 2021-06-10
Manual PKCS#11 signer testing with SoftHSMv2


ximon18:

ximon18:
> **Note:** In the shell command logs below commands are prefixed by `[N DOTS]>` to indicate which of several shell sessions the command was/should be issued in.

Tested using Krill commit eb97735 (the current head of the `issue-698-record-keys-on-upgrade` branch) and SoftHSM 2.6.1 on Ubuntu 21.10.

Krill was installed like so:

```
[1]> cargo install --git https://github.com/NLnetLabs/krill --rev eb97735 --locked --features hsm
```

The Krill config file looked like this:
```
[1]> cat /tmp/krill.conf
admin_token = "abc"
data_dir = "/tmp/krill"
service_uri = "https://localhost:3000/"
log_level = "trace"
log_type = "stderr"

[[signers]]
type = "PKCS#11"
lib_path = "/usr/lib/softhsm/libsofthsm2.so"
slot_label = "My token 1"
user_pin = "1234"
```

Initialize SoftHSM:
```
[1]> softhsm2-util --init-token --slot 0 --label "My token 1" --so-pin 1234 --pin 1234
The token has been initialized and is reassigned to slot 2096890338
```

Use the [keyls](https://github.com/ximon18/keyls) helper tool to list the keys in the PKCS#11 server before running Krill:
```
[1]> cargo install --git https://github.com/ximon18/keyls --branch main --locked
[1]> keyls "pkcs11:My token 1:1234@/usr/lib/softhsm/libsofthsm2.so"
Using PKCS#11 slot id 2096890338 (0x7cfc01e2)
No keys found
```

Run Krill:
```
[1]> krill -c /tmp/krill.conf
```

Krill configures itself to use the PKCS#11 signer for and a fallback OpenSSL signer:
```
[INFO] ... Configuring signer 'PKCS#11' (type: PKCS#11, default: true, oneoff: false, random: false)
[TRACE] ... Loading PKCS#11 library '"/usr/lib/softhsm/libsofthsm2.so"'
[TRACE] ... Loaded PKCS#11 library '"/usr/lib/softhsm/libsofthsm2.so"'
[INFO] ... Configuring signer 'Fallback OpenSSL signer' (type: OpenSSL, default: false, oneoff: true, random: true)
```

In another terminal use `krillc` to create a CA:

```
[2]> export KRILL_CLI_TOKEN=abc
[2]> krillc add --ca some_ca
```

[...]

PR #703, opened 2021-11-01
HSM 4: Use SoftHSMv2 in the e2e test

ximon18:
Note: needs the `issue-582-use-softhsm-in-the-e2e-test` branch of the rpki-deploy repo (see https://github.com/NLnetLabs/rpki-deploy/pull/32).

Note: I'm not sure if we want to keep this in the end or not because with this change we test the PKCS#11 signer in the e2e test with SoftHSMv2 but we no longer test with the OpenSSL signer in the e2e test.

Perhaps once we can configure signers properly this should be updated to use SoftHSMv2 for signing and OpenSSL for one-time keys (currently this uses feature flag `hsm-tests-pkcs11` as a hacky way to force use of PKCS#11 but that also enables PKCS#11 for all signing including for one-off keys).

However for now it demonstrates that the new signer registration and mapping logic and PKCS#11 with SoftHSMv2 works which is a useful extra test of the new HSM work.


ximon18:
I'll leave this in draft PR status until I get the checks passing in the rpki-deploy repo and get the e2e test to run in this repo against this branch.

timbru (approved):
Good to hear that this works.

We can extend the Krill API to give more information about the keys. We can look into test deployments later or even add manual tests to our release process in the worst case.

ximon18:
> We can look into test deployments later or even add manual tests to our release process in the worst case.

Related to this, see:
- #553
- #564

PR #704, opened 2021-11-02
Force re-issuance of ROAs on upgrade from before 0.9.3-rc2

timbru:


src/daemon/ca/certauth.rs
```diff
@@ -1622,16 +1627,25 @@ impl CertAuth {
...
         for (rcn, rc) in self.resources.iter() {
-            let updates = rc.renew_roas(&config.issuance_timing, signer)?;
+            let updates = rc.renew_roas(force, &config.issuance_timing, signer)?;
             if updates.contains_changes() {
-                info!(
-                    "CA '{}' reissued ROAs under RC '{}' before they would expire: {}",
-                    self.handle, rcn, updates
-                );
+                if force {
+                    info!("CA '{}' reissued all ROAs under RC '{}'", self.handle, rcn);
```
ximon18, line 1642: The strength of the term "force" makes me wonder if "info" is suitable here or should this be "warn" ?

src/upgrades/mod.rs
```diff
@@ -160,7 +158,17 @@ pub trait UpgradeStore {
 
 /// Should be called when Krill starts, before the KrillServer is initiated
 pub fn pre_start_upgrade(config: Arc<Config>) -> Result<(), UpgradeError> {
-    upgrade_0_9_0(config)
+    upgrade_data_to_0_9_0(config)
+}
+
+/// Should be called when Krill starts, before the KrillServer is initiated
```
ximon18, line 164: This comment is the same as the comment on `pre_start_upgrade`, is that intentional?

src/upgrades/mod.rs
```diff
@@ -160,7 +158,17 @@ pub trait UpgradeStore {
...
+pub async fn post_start_upgrade(config: &Config, server: &KrillServer) -> Result<(), UpgradeError> {
+    if needs_upgrade(&config.data_dir, "cas", KrillVersion::candidate(0, 9, 3, 2)) {
+        info!("Reissue ROAs on upgrade to force short EE certificate subjects in the objects");
```
ximon18, line 167: "info", or should that be "warn" ?

src/daemon/ca/certauth.rs
timbru: No "warn" would make users worry and this is by design

src/upgrades/mod.rs
timbru: info, this is by design and no need to worry.

Issue #680, opened 2021-10-01
Bulk refresh of CA/parent leads to timeout


opensourcegal:
The following problem can be encountered:

CA with a lot of parents (about hundred), all the parents reside on the same RIR.

Once the refresh cycle kicks in, I guess from the code that all parents are refreshed simultaneously.
This leads to some rate limiting by the parent and the synchronization fails for most with timeout (not all, the first are fine).

Possible solutions:

- Configurable batch size for the parent refreshs, configurable wait time between batches

- Configurable ca_refresh times if success or failed 
  (if there are f.e. 100 parents and only 10 are allowed by rate limit, try failed ones every 10 minutes, completed ones only every 6 hours)

opensourcegal:
Another proposal related to this problem:

- Make the refresh of a single parent available through CLI, so that important parents (after resource changes) can be refreshed manually without the need of refreshing hundreds of parents which may fail

timbru:
ah yes, there were some assumptions in writing this code that there would not be 100+ parents, and that they would be different parents. So what it does now is fire up all syncs as asynchronous calls and then it waits to join them all (but not failing the whole job of course if any would have failed). Earlier code was sequential, but then this led to issues that the job could take too long.

I will have a think about this. I believe it would be good if the code solved this without requiring user action. Perhaps it should be smart enough to see that multiple parents are are the same entity (i.e. your CA acts as multiple children under a parent CA) and do *those* calls sequentially or in smaller batches.

I can also think of failed first.

Btw you can configure the `ca_refresh` rate - but my guess is that this won't help you in this case as it's the number of parallel calls that seems to be the issue and not so much the frequency.

timbru:
A fix will be available in release 0.9.3 using the following logic:

- limit the batch size to 25 parents per cycle
- process new parents first; i.e. for which no sync has been attempted yet
- then order parents by last exchange time, minute granularity, plan oldest first
- for parents with a last exchange in the same minute -> do failed first

Hopefully this means that rate limits will not be triggered, but even if they are all parents still get scheduled eventually.

Relevant default values can be tweaked in config. Default values are:

```
ca_refresh_parents_batch_size = 25
ca_refresh_seconds = 600
```

Closing this issue for now.. but of course it can be re-opened if this does not solve the issue.

PR #709, opened 2021-11-03
Use max batch size for parent sync (cope with rate limits) (#680)

timbru:


src/commons/api/ca.rs
```diff
@@ -1262,6 +1262,33 @@ impl ParentStatuses {
         self.0.iter()
     }
 
+    // Return the parents sorted by last exchange, i.e. let the parents
+    // without an exchange be first, and then from longest ago to most recent.
+    // Uses minute grade granularity and in cases where the exchanges happened in
+    // the same minute failures take precedence (come before) successful exchanges.
+    pub fn sorted_by_last_exchange(&self) -> Vec<&ParentHandle> {
+        let mut sorted_parents: Vec<(&ParentHandle, &ParentStatus)> = self.iter().collect();
```
ximon18, line 1270: Could this be `self.0.clone()` instead? Not sure.

src/daemon/ca/manager.rs
```diff
@@ -831,10 +831,48 @@ impl CaManager {
...
+                    // more parents than the batch size exist, so sort them:
+                    // - take the ones with no last exchanges first
+                    // - then sort by last exchange, minute grade granularity - oldest first
+                    //    - where failures come before success within the same minute
+                    // - then take the first N parents for this batch
+                    match self.status_store.lock().await.get_ca_status(&ca_handle).await {
+                        Err(e) => {
+                            panic!("System level error encountered while updating ca status: {}", e);
+                        }
+                        Ok(status) => {
```
ximon18, line 851: You don't seem to check the status here. Perhaps I misunderstand but it looks like you'll take some of the parents and sync them then exit this function, so what happens to the remaining parents? If you come back in here again they will have an older refresh time than the ones that were just refreshed, but they will still all be in the total set and so all synced Is there some logic somewhere else that makes that a NO-OP or do you need to check the status returned by `get_ca_status()` here?

ximon18 (commented):
I don't see where the batching is being tested, did I miss it?

src/daemon/ca/manager.rs
timbru: I use the `status` to get the current parents for which a status is known and sort them:
```
let mut parents_by_last_exchange = status.parents().sorted_by_last_exchange();
```

The remaining parents will be first next time.. this is also why I don't sort by last success only, but last exchange.. otherwise continuously failing parents could starve the others.

I am adding the parents for which there is no recorded status yet, i.e. new parents, first so that they take precedence. Since parent status is only created *after* the first sync attempt.

timbru:
> I don't see where the batching is being tested, did I miss it?

This is hard to test at a high level, but you are right that this should have more than eye ball checks.. I will add unit tests for the sorting and truncation logic - i.e. perhaps extract the function for this so it can be tested better.

src/daemon/ca/manager.rs
ximon18: But will not the already handled parents also be included next time?

timbru: By example: Default batch size is 25 parents. If you have 27 then you will do 25 first, and leave 2, and then in t he next run (10 minutes by default) do 25 again: the 2 remaining and 23 that were also processed in the previous batch.

ximon18: Yes. So is right that the additional 23 are also processed?

timbru: yes, sorry.. missing info might be that it is always safe to sync with a parent. This is a no-op in case the entitlements have not changed. Along the same lines.. if you have 1 parent it's perfectly fine to sync it in every run.

Some might say that the default of 10 mins is aggressive, but I have not heard of any issues from parents.

If a CA has 100s of parents, and they use defaults then it will take longer for the parents to be resync-ed. E.g. with 100 parents -> every 40 minutes. But this is not really an issue either. You kind of want to stay below the hour I think - but there are no formal RFC requirements here - every hour would be good because you want to learn about changes in resources soon-ish.

In the unlikely case that a CA has that many parents, they can also tweak the defaults - e.g. refresh 25 parents every 5 minutes instead of every 10 minutes.

ximon18: Thanks.

ximon18 (approved):

PR #712, opened 2021-11-04
Support ASPAs with an explicit empty provider as set.

timbru:
This updates the logic to allow for an empty provider AS set (as per sidrops discussion). Tests were added, and an issue was found in the encoding of empty provider as sets, which was fixed in this PR for rpki-rs:

https://github.com/NLnetLabs/rpki-rs/pull/171

### DIFF
--- a/src/commons/api/aspa.rs
+++ b/src/commons/api/aspa.rs
@@ -133,11 +133,15 @@ impl fmt::Display for AspaDefinition {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         // example: 65000 => 65001, 65002(v4), 65003(v6)
         write!(f, "{} => ", self.customer)?;
-        for i in 0..self.providers.len() {
-            if i > 0 {
-                write!(f, ", ")?;
+        if self.providers.is_empty() {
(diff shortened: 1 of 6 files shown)


src/commons/api/aspa.rs
```diff
@@ -158,26 +163,29 @@ impl FromStr for AspaDefinition {
...
             Err(AspaDefinitionFormatError::ExtraParts)
         } else {
-            // Ensure that the providers are sorted, there is at least one, and there are no duplicates
+            // Ensure that the providers are sorted,  and there are no duplicates
             providers.sort_by_key(|p| p.provider());
 
-            let mut last_seen = providers.first().ok_or(AspaDefinitionFormatError::ProviderAsMissing)?;
-
             if providers.len() > 1 {
+                let mut last_seen = providers.first().unwrap();
```
partim, line 188: This is probably better done using `providers.windows(2).find(...)` or somesuch?

partim (approved):

src/commons/api/aspa.rs
timbru: yes, good point. Much improved now I hope.

Issue #713, opened 2021-11-04
Fix "eligible certificate not after time in the past"


timbru:
As it turns out the eligible not-after time for entitlements is set to "now" in cases where a child did not yet receive any certificates. This is obviously wrong - even if child CAs will typically still request a certificate based on resource eligibility.

PR #714, opened 2021-11-04
Fix eligible not after in case no certificates had been issued yet (#713)

timbru:


src/daemon/ca/certauth.rs
```diff
@@ -753,13 +753,31 @@ impl CertAuth {
...
+        // still trigger them to request a replacement for the first which was about to expire.
         for ki in child_keys {
             if let Some(issued) = my_rc.issued(&ki) {
                 issued_certs.push(issued.clone());
-                let eligible_not_after = Self::child_cert_eligible_not_after(issued, issuance_timing);
-                if eligible_not_after > not_after {
-                    not_after = eligible_not_after
+
+                let expires = issued.validity().not_after();
+                let warn_threshold = Time::now()
```
partim, line 776: Would it be a good idea to create this value outside the for loop?

partim (approved):

Issue #541, opened 2021-06-10
Add support for defining Signers in the Krill config file


ximon18:
KrillSigner should create multiple signers based on config file entries.

In the proof-of-concept Signers were defined in the config file like so:

```
[[signers]]
name = “Some signer name”
type = “OpenSSL or PKCS#11 or KMIP”
<more key value pairs>
```

Where the supported <more key value pairs> depended on the Signer type, e.g.:

_(if there is more than one signer, one of them must have a default = true key/value pair)_

See https://github.com/NLnetLabs/krill/blob/issue-21-hsm-support-for-keys/defaults/krill-hsm.conf for more details.

ximon18:
Some thoughts regarding this:

- For backward compatibility in the absence of any defined signers Krill must default to a local OpenSSL signer. On first use after upgrade Krill will also need to record existing OpenSSL key ids in the new signer store (see #698).

- If an operator intends to use an HSM with Krill but forgets to configure it _before_ creating causing Krill to create a key then that key will be created in the OpenSSL signer store and cannot (yet?) be migrated (see [this related comment](https://github.com/NLnetLabs/krill/issues/570#issuecomment-958752870)).

- If an operator edits the signers in the config file without restarting Krill then the config file may not reflect the actual signers in use by Krill. `krillc` extensions could help with this (se [this related comment](https://github.com/NLnetLabs/krill/issues/570#issuecomment-958752870)).

- If an operator removes a signer from the config file and restarts Krill any keys created in that signer backend will become inaccessible to Krill. Should it even be possible to delete an "in-use" signer? Is this an argument for adding/removing signers via `krillc` commands instead of config file settings and storing the connection details (including secrets?) in the KV store?

- If an operator runs some future version of Krill that supports clustering but HSM configuration is stored in the config file what happens if one cluster node has a different HSM configuration than some other cluster node? Is this an argument for a centralized KV store based config management of signers that would be shared by all nodes in a cluster? **Conversely**, what about the case where the HSM node a Krill node should use has different configuration than for another Krill node, e.g. Krill deployed in two datacenters/zones and a HSM with nodes also in both zones, and Krill should contact the HSM via private IP address which is say 10.0.0.1 in one zone and 10.0.0.2 in the other zone? (however, is it actually the case that HSMs that support clustering typically abstract this away in their PKCS#11 library and its configuration? how does that work for KMIP servers as they use TCP+TLS, is that done via DNS mechanisms and load balancers etc?)

[...]

PR #718, opened 2021-11-09
HSM 5: Add support for defining signers in the config file.

ximon18:
Also see the updated e2e test in PR https://github.com/NLnetLabs/rpki-deploy/pull/33 (which builds against this Krill PR branch).


src/commons/crypto/signing/dispatch/krillsigner.rs
```diff
@@ -141,3 +186,512 @@ impl KrillSigner {
...
+
+            config.default.get_or_insert(true);
+
+            if matches!(config.signer_type, SignerType::OpenSsl(_)) {
+                config.oneoff.get_or_insert(true);
+                config.random.get_or_insert(true);
+            }
+        }
+
+        // One and only one signer should be the default. The default signer is used for operations that don't concern
```
timbru, line 220: right, elsewhere we have been testing config in `Config::verify` - doing so here may keep this code simpler

src/commons/crypto/signing/dispatch/signerprovider.rs
```diff
@@ -15,47 +15,140 @@ use crate::commons::{
...
+pub(crate) struct SignerFlags;
+
+#[cfg(not(feature = "hsm"))]
+impl Default for SignerFlags {
+    fn default() -> Self {
+        Self
+    }
+}
+
+#[cfg(feature = "hsm")]
```
timbru, line 29: I get the feeling that the 'hsm' feature flag could be used less. I.e. let limit the available options for the `SignerProvider` enum, but essentially have the same boilerplate always. This is not urgent, we can  look at this later.. also once all of this is merged and has had more field testing we may even just want to remove the feature as such an leave it to config to decide to use hsms or not.

src/commons/crypto/signing/dispatch/signerrouter.rs
```diff
@@ -159,159 +141,52 @@ impl SignerRouter {
...
+        // One and only one signer should be the default. The default signer is used for operations that don't concern
+        // an existing key, i.e. key creation, one-off signing and random number generation. The latter two may be
+        // delegated below to other signers instead of the default signer.
+        // Create the signers
+        let mut default_signer: Option<Arc<SignerProvider>> = None;
+        let mut one_off_signer: Option<Arc<SignerProvider>> = None;
+        let mut rand_falback_signer: Option<Arc<SignerProvider>> = None;
+        let mut all_signers = Vec::new();
+
+        for signer in signers.drain(..) {
```
timbru, line 158: as mentioned elsewhere I wonder whether we should do this in `Config::verify`. That could keep things simpler and more focussed here.

timbru (approved):

src/commons/crypto/signing/dispatch/krillsigner.rs
ximon18: I put it here to try and keep multi-signer specific stuff out of "core" Krill. The rules are a bit more complicated than this comment alone, as demonstrated by the tests at the bottom of the file.

I'm not very happy about the number of things the test have to check, it makes me feel that the config syntax for multi-signers is wrong and should be revisited. E.g. instead of each signer defining if it is default, oneoff and/or random fallback signer, perhaps those should be top level settings like `auth_type`, e.g. `default_signer = ...` and `random_fallback_signer = `. The problem with the latter approach however is that you then have to name signers so that you can refer to them elsewhere in the config, and you separate something about the signer out from the signer itself. 

Ideas welcome!

timbru: ok, let's keep it here for now

src/commons/crypto/signing/dispatch/signerrouter.rs
timbru: see other comment, let's keep it here for now

ximon18:
@tim: Currently with this approach one sets "role" boolean flags on each signer, with there being some logic for deciding what to do if flags are not supplied.

With no flags specified a signer is the "default" signer (which is perhaps misnamed) which will be used to create new keys and generate random numbers, and an OpenSSL signer will be used to generate, sign with and destroy one off keys and as a fallback random number generator if the "default" signer doesn't support this capability. A signer can explicitly be set as being the one-off signer rather than using OpenSSL, and a signer can be explicitly set as the fallback for random number generation.

We default to one-off signing and random number generation using OpenSSL because that is fast while round trips to and processes within an HSM may be slow, and because the security benefits of using a HSM is not considered to be outweigh the speed concern in the cases of one-off keys (which are never persisted) and random number generation (for which modern servers are assumed to be cryptographically secure enough at this already).

However, there is no way at the moment to force random number generation to be done with OpenSSL. I'm thinking of changing the default behaviour such that no flags/roles means create keys with that signer but don't use it for one-off signing (already the current behaviour) or random number generation, and such that specifying `random` for a signer forces that signer to be used to generate random numbers, and so it becomes a hard fail if it isn't capable of doing so

Do you have any thoughts about this?

ximon18:
> @tim: Currently with this approach one sets "role" boolean flags on each signer, with there being some logic for deciding what to do if flags are not supplied.
> 
> With no flags specified a signer is the "default" signer (which is perhaps misnamed) which will be used to create new keys and generate random numbers, and an OpenSSL signer will be used to generate, sign with and destroy one off keys and will also be used as a fallback random number generator if the "default" signer doesn't support this capability. A signer can explicitly be set as being the one-off signer rather than using OpenSSL, and a signer can be explicitly set as the fallback for random number generation.
> 
> We default to one-off signing and random number generation using OpenSSL because that is fast while round trips to and processes within an HSM may be slow, and because the security benefits of using a HSM is not considered to be outweigh the speed concern in the cases of one-off keys (which are never persisted) and random number generation (for which modern servers are assumed to be cryptographically secure enough at this already).
> 
> However, there is no way at the moment to force random number generation to be done with OpenSSL. I'm thinking of changing the default behaviour such that no flags/roles means create keys with that signer but don't use it for one-off signing (already the current behaviour) or random number generation, and such that specifying `random` for a signer forces that signer to be used to generate random numbers, and so it becomes a hard fail if it isn't capable of doing so.
> 
> Do you have any thoughts about this?

See https://github.com/NLnetLabs/krill/issues/541#issuecomment-977627035.

timbru (approved):

PR #720, opened 2021-11-09
Re-issue all objects on key activation or resource shrink

timbru:


src/commons/api/ca.rs
```diff
@@ -335,6 +335,18 @@ impl IssuedCert {
     pub fn replaces(&self) -> Option<&ReplacedObject> {
         self.replaces.as_ref()
     }
+
+    /// Returns a (possibly empty) set of remaining resources which
+    /// is the intersection of the encompassing resources and this
+    /// certificate's current resources. Returns None if the current
+    /// resource set is not overclaiming.
+    pub fn remaining_resources(&self, encompassing: &ResourceSet) -> Option<ResourceSet> {
```
ximon18, line 343: I have no idea if this is correct or not, I don't know the domain well enough.

src/daemon/ca/child.rs
```diff
@@ -184,6 +187,111 @@ impl ChildCertificates {
...
+                        certified_key,
+                        issuance_timing,
+                        signer,
+                    )?);
+                }
+            }
+        }
+
+        // Also shrink suspended, in case they would come back
+        for suspended in self.issued.values() {
```
ximon18, line 242: This loop is the same as the one above. It use the name `suspended` but suspended is `self.issued.values()` just like above wher `issued` is also `self.issued.values()` - is this iterating over the correct values?

src/daemon/ca/rc.rs
```diff
@@ -301,40 +301,20 @@ impl ResourceClass {
                 rcvd_cert.validity().not_after().to_rfc3339()
             );
 
+            // Prep certified key for updated received certificate
+            let certified_key = CertifiedKey::create(rcvd_cert);
+
             // Check whether child certificates should be shrunk
```
ximon18, line 307: Based on the name of the invoked function and the modification made to the descriptive comment on the function, I assume this comment here is incorrect: this doesn't just check if shrinking is needed, it will also actually do the shrinking, right?

src/daemon/ca/rc.rs
```diff
@@ -301,40 +301,20 @@ impl ResourceClass {
...
-                        None,
-                        &config.issuance_timing,
-                        signer,
-                    )?;
-                    updates.issue(re_issued);
-                }
-            }
+            let updates = self
+                .certificates
+                .shrink_overclaiming(&certified_key, &config.issuance_timing, signer)?;
```
ximon18, line 310: The old code "shrunk" based on `current_key` but the new code moves creation of `certified_key` up and shrinks based on that, is this difference correct?

ximon18 (approved):

src/commons/api/ca.rs
timbru: hopefully more clear now

src/daemon/ca/rc.rs
```diff
@@ -543,24 +523,36 @@ impl ResourceClass {
...
-                    cert_updates.issue(re_issued);
+                if !roa_updates.is_empty() {
+                    let roas_updated = CaEvtDet::RoasUpdated {
+                        resource_class_name: self.name.clone(),
+                        updates: roa_updates,
+                    };
+                    events.push(roas_updated);
+                }
+
+                let aspa_updates = self.aspas.renew(new_key, None, issuance_timing, signer)?;
```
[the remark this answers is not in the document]
timbru: good point - for aspas I used the opportunity to reduce code duplication.. did the same now for ROAs

src/daemon/ca/child.rs
timbru: doh.. yes, good catch it should loop over suspended

src/daemon/ca/rc.rs
timbru: the logic was correct, renamed things and comments to make it more clear.

src/daemon/ca/rc.rs
timbru: logic was correct, but the naming was confusing in this context - hopefully better now

ximon18 (approved):

PR #725, opened 2021-11-10
HSM 7: Record pre-existing OpenSSL keys in the mapper on upgrade.

ximon18:


ximon18:
Hmm, it should probably test with multiple pre-existing keys, not just one.

src/upgrades/mod.rs
```diff
@@ -261,4 +364,55 @@ mod tests {
 
         let _ = fs::remove_dir_all(work_dir);
     }
+
+    #[cfg(all(feature = "hsm", not(any(feature = "hsm-tests-kmip", feature = "hsm-tests-pkcs11"))))]
```
timbru, line 368: Why is this excluded when either of those hsm-test features is enabled?

I seems to me that there would be no conflict in running this test again - and it's a bit friendlier on my IDE with all-features enabled :)

src/commons/crypto/signing/dispatch/signerprovider.rs
```diff
@@ -212,7 +212,7 @@ impl SignerProvider {
 // Implement the functions defined by the `Signer` trait because `SignerRouter` expects to invoke them, but as the
 // dispatching is not trait based we don't actually have to implement the `Signer` trait.
 impl SignerProvider {
-    pub(super) fn create_key(&self, algorithm: PublicKeyFormat) -> Result<KeyIdentifier, SignerError> {
+    pub(crate) fn create_key(&self, algorithm: PublicKeyFormat) -> Result<KeyIdentifier, SignerError> {
```
timbru, line 215: can't this just be `pub` then? This is not (yet anyway) used as a library

src/commons/crypto/signing/dispatch/signerrouter.rs
```diff
@@ -187,6 +187,20 @@ impl SignerRouter {
         })
     }
 
+    pub(crate) fn get_mapper(&self) -> Arc<SignerMapper> {
```
timbru, line 190: Why `pub(crate)` instead of `pub`?

src/commons/crypto/signing/dispatch/krillsigner.rs
```diff
@@ -88,6 +92,16 @@ impl KrillSigner {
         Ok(KrillSigner { router })
     }
 
+    #[cfg(feature = "hsm")]
+    pub fn get_mapper(&self) -> Arc<SignerMapper> {
+        self.router.get_mapper()
+    }
+
+    #[cfg(feature = "hsm")]
+    pub(crate) fn get_active_signers(&self) -> HashMap<Handle, Arc<SignerProvider>> {
```
timbru, line 101: why use  `pub(crate)`?

timbru (approved):

src/upgrades/mod.rs
```diff
@@ -159,13 +166,109 @@ pub trait UpgradeStore {
...
+                            keys_dir.to_string_lossy()
+                        ),
+                        err,
+                    ))
+                })?;
+
+                if entry.path().is_file() {
+                    // Is it a key identifier?
+                    if let Ok(key_id) = KeyIdentifier::from_str(&entry.file_name().to_string_lossy()) {
+                        if krill_signer.is_none() {
```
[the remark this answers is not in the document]
ximon18: The reason for creating it inside the loop is now no longer valid. It was created inside the loop because the creation of the KrillSigner caused Krill to immediately attempt to load the PKCS#11 library. I felt that this could be confusing if it happened during an upgrade step that turned out to be a no-op, why would it then have done anything concerning configured signers? However the PKCS#11 code has since been modified and no longer loads the library until first use. So I'll revisit this code in light of that.

src/upgrades/mod.rs
```diff
@@ -159,13 +166,109 @@ pub trait UpgradeStore {
...
+                            if !krill_signer.get_key_info(&key_id).is_ok() {
+                                // No, record it
+
+                                // Do we already know the handle of the OpenSSL signer? If not, find it. It should be
+                                // possible to find the key via the OpenSSL signer because it shoud look for it in the
+                                // same directory and file name as we are now processing. Finding it confirms for us
+                                // that we have the correct signer handle (unless someone migrated the keys into another
+                                // signer and registered them there and then manually deleted the signer store directory
+                                // ... but then if they go behind Krill's back there are limits to what we can cope with)
+                                if signer_handle.is_none() {
```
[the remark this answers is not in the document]
ximon18: In the current HSM code it is possible to use Krill without any OpenSSL signer at all. By default if you configure a PKCS#11 or KMIP signer then an OpenSSL signer will be used for one-off key generation & signing, and for random number generation but it's possible to configure Krill to use the backend HSM for both of these cases instead of OpenSSL. Aactually I need to look at the random number generation case as there is a subtle difference between using OpenSSL always to generate random numbers and using it only in case the backend HSM doesn't support random number generation...

src/upgrades/mod.rs
ximon18: It was excluded because it conceptually makes no sense (to me at least) to run this test when Krill has either of those special Rust features enabled as they assume a fresh installation that uses solely KMIP or PKCS#11, no OpenSSL at all and no migration from any prior set of OpenSSL keys as there will never be any prior state.

src/upgrades/mod.rs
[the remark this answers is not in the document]
ximon18: Hmm, I guess you are right actually. Krill doens't need an OpenSSL signer for fresh installs, but this code runs in the context of a pre-existing instance of Krill that has at least one OpenSSL key already, therefore it MUST have had an OpenSSL signer to create those keys.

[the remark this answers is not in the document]
ximon18: One caveat however: it's possible that as part of upgrading a user edited their `krill.conf` so that it defines two OpenSSL signers each with their own keys directory, e.g. for testing keyroll between signers. We then can't "find" the signer by assuming it is the one and only OpenSSL signer. That's why this code is looking for the signer that owns the key rather than for the first OpenSSL signer.

Issue #743, opened 2021-12-16
Allow configuring RFC6492/8181 client timeouts


timbru:
Timeouts may occur in case a Krill instance has a huge amount of children or publishers.

This should be alleviated in 0.9.3 wrt publishing because of the jitter added there - which will spread out publishers connecting. We should probably introduce a similar jitter for synchronising with parents.

In any case the default timeout was 2 minutes which was long enough in almost all cases. Now increasing this default to 4 minutes to be even more safe - and making it configurable through the config file:

`post_protocol_msg_timeout_seconds = 240 // Defaults to 4 minutes`

timbru:
in release 0.9.4

PR #744, opened 2021-12-16
Allow configuring RFC6492/8181 client timeouts #743

timbru:


defaults/krill.conf
```diff
@@ -244,6 +244,15 @@
 #
 ### post_limit_rfc6492 = 1048576
 
+# Set the timeout for RFC 6492 and RFC 8181 client http connections to the parent
```
ximon18, line 247: s/http/HTTP/

Also, to be clear this does NOT affect the time permitted for an initial connection to be established, rather once connected it (quoting the [reqwest docs](https://docs.rs/reqwest/0.11.5/reqwest/struct.ClientBuilder.html#method.timeout)) is the maximum time allowed "from when the request starts connecting until the response body has finished".

It's a bit wordy but one alternative could be "Set the timeout for a complete RFC 6492 and RFC 8181 client HTTP request-response round-trip to the parent or publisher, excluding the time required to establish the connection." ?

src/daemon/config.rs
```diff
@@ -285,6 +289,9 @@ pub struct Config {
     #[serde(default = "ConfigDefaults::post_limit_rfc6492")]
     pub post_limit_rfc6492: u64,
 
+    #[serde(default = "ConfigDefaults::post_protocol_msg_timeout_seconds")]
+    pub post_protocol_msg_timeout_seconds: u64,
```
ximon18, line 293: `ca_refresh_seconds` is a u32, so why is this a u64?

timbru: Because of the function signature where this is used.. refresh goes to our background job schedular. This is invisible to users btw

ximon18 (approved):

defaults/krill.conf

PR #753, opened 2022-01-04
Fix compilation on armv7-unknown-linux-gnueabihf by supplying missing Serde impls. (#751)

### LINKED ISSUE #751: Multi-user feature code fails to compile when cross compiling for ARMv7
Krill version tested:
```
$ rustc -V
rustc 1.57.0 (f1edd0429 2021-11-29)

$ git describe --all --long
heads/main-0-g3ab1f38c

$ lsb_release -d
Description:    Ubuntu 21.10

$ uname -a
Linux ximon-laptop 5.15.0-051500-generic #202110312130 SMP Sun Oct 31 21:33:20 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

Prerequisites:

```
$ cargo install cross
$ rustup target add armv7-unknown-linux-gnueabihf
```

Without the `multi-user` feature (which is normally enabled by default):

```
$ cross build --target armv7-unknown-linux-gnueabihf --no-default-features --features static-openssl
    Finished dev [unoptimized + debuginfo] target(s) in 2m 25s
```

With the `multi-user` feature (i.e. the default):

```
$ cross build --target armv7-unknown-linux-gnueabihf --features static-openssl
...
error[E0277]: the trait bound `AtomicU64: Deserialize<'_>` is not satisfied
    --> src/daemon/auth/common/crypt.rs:44:5
     |
44   |     counter: AtomicU64,     // + 64 bits = 96 bits = CHACHA20_NONCE_BIT_LEN
     |     ^^^^^^^ the trait `Deserialize<'_>` is not implemented for `AtomicU64`
     |
note: required by `next_element`
```

However, without cross compiling it works:
```
$ cargo build --features static-openssl
...
    Finished dev [unoptimized + debuginfo] target(s) in 3m 13s
```

ximon18:
Fixes #751.

I'm not sure if this is the _right_ way to fix this, but it seems to work.

Tested on my Raspberry Pi 4b by cross compiling on my x86_64 laptop host then running the produced binary with config-file users on the Pi:

```
$ cargo install cross
$ rustup target add armv7-unknown-linux-gnueabihf
$ cross build --target armv7-unknown-linux-gnueabihf --features static-openssl --release
$ scp target/release/krill pi@192.168.0.18:
$ scp target/release/krillc pi@192.168.0.18:
$ ssh pi@192.168.0.18
$ ./krillc config user --id x18
Enter the password to hash: 
***

[auth_users]
"x18" = { attributes={ role="admin" }, password_hash="********", salt="********" }
$ grep -Ev '^#' krill.conf | sed '/^\s*$/d'
log_level = "trace"
log_type = "stderr"
admin_token = "********"
service_uri = "https://localhost:3000/"
auth_type = "config-file"

[auth_users]
"x18" = { attributes={ role="admin" }, password_hash="********", salt="********" }
$ ./krill -c ./krill.conf
```

I then visited the Krill login page in my browser and checked that an incorrect password was correctly rejected and a valid password was accepted.

Unfortunately GitHub Actions doesn't seem to provide ARMv7 architecture runners, only x86_64 ones, so we can't do automated testing on ARMv7 without using a self-hosted runner or some other service.

If needed I can provide access to my Pi ;-)

### DIFF
--- a/src/daemon/auth/common/crypt.rs
+++ b/src/daemon/auth/common/crypt.rs
@@ -38,61 +38,10 @@ const POLY1305_TAG_BYTE_LEN: usize = POLY1305_TAG_BIT_LEN / 8;
 const CLEARTEXT_PREFIX_LEN: usize = CHACHA20_NONCE_BYTE_LEN + POLY1305_TAG_BYTE_LEN;
 const UNUSED_AAD: [u8; 0] = [0; 0];
 
-//------------ CrossPlatformDeserializableAtomicU64 -------------------------------------------------------------------
-
-/// This type provides Serde (de)serialization irrespective of target platform, as the Serde crate provided derive
-/// impls only appear to produce code when compiled on x86_64. See: https://github.com/NLnetLabs/krill/issues/751
-#[derive(Debug)]
-struct CrossPlatformDeserializableAtomicU64(AtomicU64);
-
-impl std::ops::Deref for CrossPlatformDeserializableAtomicU64 {
-    type Target = AtomicU64;
-
-    fn deref(&self) -> &Self::Target {
-        &self.0
-    }
-}
-
-impl<'de> serde::Deserialize<'de> for CrossPlatformDeserializableAtomicU64 {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        struct DeserializableAtomicU64Visitor;
-
-        impl<'de> serde::de::Visitor<'de> for DeserializableAtomicU64Visitor {
-            type Value = CrossPlatformDeserializableAtomicU64;
-
-            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
-                formatter.write_str("an unsigned integer in the range 0..=2^64-1")
-            }
-
-            fn visit_u64<E>(self, v: u64) -> Result<Self::Value, E>
-            where
-                E: serde::de::Error, {
-                Ok(CrossPlatformDeserializableAtomicU64(AtomicU64::new(v)))
-            }
-        }
-
-        deserializer.deserialize_u64(DeserializableAtomicU64Visitor)
-    }
-}
-
-impl serde::Serialize for CrossPlatformDeserializableAtomicU64 {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: serde::Serializer,
-    {
-        serializer.serialize_u64(self.0.load(Ordering::SeqCst))
-    }
-}
-
-//------------ NonceState ---------------------------------------------------------------------------------------------
-
 #[derive(Debug, Deserialize, Serialize)]
 pub struct NonceState {
-    sender_unique: [u8; 4],           //   32 bits
-    counter: CrossPlatformDeserializableAtomicU64, // + 64 bits = 96 bits = CHACHA20_NONCE_BIT_LEN
+    sender_unique: [u8; 4], //   32 bits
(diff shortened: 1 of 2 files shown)


src/daemon/auth/common/crypt.rs
```diff
@@ -38,10 +38,61 @@ const POLY1305_TAG_BYTE_LEN: usize = POLY1305_TAG_BIT_LEN / 8;
 const CLEARTEXT_PREFIX_LEN: usize = CHACHA20_NONCE_BYTE_LEN + POLY1305_TAG_BYTE_LEN;
 const UNUSED_AAD: [u8; 0] = [0; 0];
 
+//------------ CrossPlatformDeserializableAtomicU64 -------------------------------------------------------------------
+
+/// This type provides Serde (de)serialization irrespective of target platform, as the Serde crate provided derive
+/// impls only appear to produce code when compiled on x86_64. See: https://github.com/NLnetLabs/krill/issues/751
+#[derive(Debug)]
+struct CrossPlatformDeserializableAtomicU64(AtomicU64);
```
timbru, line 46: to me it would be more intuitive to use a notation like `#[serde(deserialize_with = "ext_serde::de_atomic64", serialize_with = "ext_serde::ser_atomicu64")]` instead of a wrapper type. And then copy those functions over to  `commons/utils/ext_serde.rs`.

src/daemon/auth/common/crypt.rs
```diff
@@ -38,10 +38,61 @@ const POLY1305_TAG_BYTE_LEN: usize = POLY1305_TAG_BIT_LEN / 8;
...
+}
+
+//------------ NonceState ---------------------------------------------------------------------------------------------
+
 #[derive(Debug, Deserialize, Serialize)]
 pub struct NonceState {
-    sender_unique: [u8; 4], //   32 bits
-    counter: AtomicU64,     // + 64 bits = 96 bits = CHACHA20_NONCE_BIT_LEN
+    sender_unique: [u8; 4],           //   32 bits
+    counter: CrossPlatformDeserializableAtomicU64, // + 64 bits = 96 bits = CHACHA20_NONCE_BIT_LEN
```
timbru, line 95: I realise this was not introduced now, but I wonder if it would be better to use 64 bits for `sender_unique` and 32 bits (4 billion) for `counter`. Although the chance of a collision for `sender_unique` is small with 32 bits, it's much smaller still with 64 - and 4 billion feels like enough space for the counter. But, just a thought - I am not 100% sure what is best.

src/daemon/auth/common/crypt.rs
ximon18: Let me try that.

src/daemon/auth/common/crypt.rs
ximon18: This design was based on [RFC 8439 Section 2.3](https://datatracker.ietf.org/doc/html/rfc8439#section-2.3) which says:
> In cases where a single key is used by multiple senders, it is important to make sure that they don't use the same nonces.  This can be assured by partitioning the nonce space so that the first 32 bits are unique per sender, while the other 64 bits come from a counter.

ximon18:
Regarding this:
> Unfortunately GitHub Actions doesn't seem to provide ARMv7 architecture runners, only x86_64 ones, so we can't do automated testing on ARMv7 

It might be possible via `cross test` with `cargo-cross` but the docs for that say it is VERY slow and the UI tests are already very slow.

src/daemon/auth/common/crypt.rs
timbru: ok sure.. I think 32 bits is probably enough to not have any collisions (since they all start with counter 0)

PR #759, opened 2022-01-12
Silence multi-user related dead code warnings.

ximon18:
Silence multi-user related dead code warnings about fields that are written to but not read unless the multi-user feature is enabled. 

This typically occurs when disabling default features while developing to speed up compilation and then you get these annoying warnings. We don't exclude the fields entirely from the compilation because that would add more conditional compilation guards which makes the code harder to read.

### DIFF
--- a/src/commons/actor.rs
+++ b/src/commons/actor.rs
@@ -117,7 +117,11 @@ pub struct Actor {
     is_user: bool,
     attributes: Attributes,
     new_auth: Option<Auth>,
+
+    #[cfg_attr(not(feature = "multi-user"), allow(dead_code))]
     policy: Option<AuthPolicy>,
+
+    #[cfg_attr(not(feature = "multi-user"), allow(dead_code))]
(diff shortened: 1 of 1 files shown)


src/commons/actor.rs
```diff
@@ -117,7 +117,11 @@ pub struct Actor {
     is_user: bool,
     attributes: Attributes,
     new_auth: Option<Auth>,
+
+    #[cfg_attr(not(feature = "multi-user"), allow(dead_code))]
```
timbru, line 121: is `allow(dead_code)` still needed when the `feature` notations are there as well?

timbru (approved):

src/commons/actor.rs
ximon18: Yes because you can't use `#[cfg(feature(multi-user))]` here as that would remove the fields entirely and then break lots of other code, and we don't want to mark all those other places as multi-user only as that makes the code less readable, at least that's what we discussed in the past.

PR #762, opened 2022-01-24
Retry xz compression using new cargo deb system xz option

ximon18:
If this works we can remove [this workaround](https://github.com/NLnetLabs/krill/blob/main/.github/workflows/pkg.yml#L158) by using the [new --xz-feature of cargo-deb](https://github.com/kornelski/cargo-deb/issues/12).


ximon18:
Unfortunately this doesn't seem to fix the issue. See https://github.com/kornelski/cargo-deb/issues/12#issuecomment-1019934770.

Issue #764, opened 2022-01-26
ARM Support for Docker Images


ericparton:
Hello,

Would it be possible to produce Krill Docker images that have support for ARM architectures?  With the proliferation of ARM-based MacBook laptops and ARM-based virtual machines offered by cloud providers, ARM architectures are becoming more popular. 

Thanks

ximon18:
Good evening @ericparton,

Apologies for the delayed response, I was busy investigating the various options for doing this.

I have worked out a way to do this which builds on the recent work done to create DEB packages for ARM (specifically for ARMv7/armhf tested on a Raspberry Pi 4b, and ARM64/aarch64 tested on a Pine64). This would result in Docker images that like our current Docker images use an Alpine base image and use [musl libc](https://musl.libc.org/) instead of glibc, and just like the ARM DEB packages would use statically linked OpenSSL.

Would this work for you? Which ARM architecture and operating system are you interested in using this on?

It will take me a little while to work out how best to add this in to our existing GitHub Actions packaging workflows as currently the cross compilation is done in the `pkg` workflow but the Docker image building is done in the `pkg-docker.yml` workflow, and currently they trigger slightly differently so can't easily be merged, and if not merged then one would have to wait for and take binaries built by the other (e.g. using GitHub Action `dawidd6/action-download-artifact@2`), and the Docker workflow needs to be extended to do cross-target building (e.g. using `docker/setup-qemu-action@v1` and `docker/setup-buildx-action@v1`).

If you'd like to test an ARM Krill Docker image I have a prototype setup I can use to build a one-off image, I just need to know which architecture to target, preferably I need to know both of the following:

- QEmu supported target, for ARM that is either `linux/arm/v7` or `linux/arm/v6` (taken from the output of the `docker/setup-qemu-action@v1` GitHub Action)
- Cargo cross supported target name, one of the ARM `target` values listed here: https://github.com/cross-rs/cross#supported-targets

Thanks,

Ximon

ericparton:
Thanks for taking a look at this. Just to give you some more context, our engineers use MacBooks to write software and we use the Krill Docker image during the development and testing process of our RPKI software. 

As of last November, Apple only sells their ARM based MacBooks, so we've been going through all the software that our engineers use and making sure that it works on an ARM based MacBook before we start purchasing more of the new MacBooks for general use throughout the organization.

[...]

PR #773, opened 2022-02-10
Extend Docker publishing to create ARM architecture images and a multi-arch manifest

ximon18:
This PR was already potentially on our roadmap but picked up now due to support being requested in issue #764.

This PR includes a [fix for compilation on ARMv7](https://github.com/NLnetLabs/krill/commit/c60ef30c745c28ff08165a4aef1e1aadefb41186) that was already made to the `dev` branch as without it the workflow fails.

**Note:** This PR uses advanced Docker features such as [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/), [multi-arch manifests](https://docs.docker.com/engine/reference/commandline/manifest/) and Docker [BuildKit](https://docs.docker.com/develop/develop-images/build_enhancements/), and will change the way images are published to [Docker Hub](https://hub.docker.com/). However, no impact or change is expected for users. More on that below.

This change replaces the `pkg-docker.yml` workflow by supporting Docker packaging in the main `pkg.yml` workflow and extends the workflow to support cross-compilation for Docker images as well as for DEB packages (as was already supported).

The changes are quite substantial and also affect the `Dockerfile` (so that it can be built either standalone by compiling current host Krill code as was the case before this PR, or by copying in Krill binaries that were built already and should just be included in the image as is the case with cross-compilation).

I have added extensive comments to attempt to make it clear what is happening and why. This PR is the result of substantial trial and error of different approaches. For example for Docker images it is possible to use the docker-push-build GitHub Action in combination with the setup-qemu-action to compile natively inside a QEmu VM for non-x86-64 architectures, but it is **extremely** slow, on the order of hours, when compiling Krill inside a QEmu VM on a GitHub hosted runner.

Limited testing has been done on a Raspberry Pi 4b (ARMv7 architecture) by running a Docker container produced earlier as part of development and testing work done in my personal Krill fork and Docker Hub repository:

```bash
pi@pi2gb:~ $ sudo docker run -it --rm --pull always ximoneighteen/krill krillc --version
latest: Pulling from ximoneighteen/krill
80ca8924263d: Already exists 
99c4793f5471: Pull complete 
b70bf3b87baa: Pull complete 
3f4004224ece: Pull complete 
d8a7e4a19835: Pull complete 
6741232e22af: Pull complete 
be27dedd1323: Pull complete 
d20d3e15f9b8: Pull complete 
767e31115a67: Pull complete 
Digest: sha256:588544557ca9664df1062b4c35633204ace6f2ff092b592512c444950aba836f
Status: Downloaded newer image for ximoneighteen/krill:latest
Krill Client 0.9.5-dev
```

**Note:** This PR will CHANGE the Docker images available for `nlnetlabs/krill` on Docker Hub. Instead of being a single x86-64 architecture image as is now the case it will become a multi-arch "manifest' image that references three other pushed images:
- nlnetlabs/krill:latest-amd64
- nlnetlabs/krill:latest-arm64
- nlnetlabs/krill:latest-armv7

Users pulling and executing the image will automatically get the right architecture image for their host platform. This change should NOT have any impact on end users.

Additional references:
- https://www.docker.com/blog/multi-platform-docker-builds/


ximon18:
Maybe it would be useful to modify the behavior when not pushing to `main` or to a tag so that it instead DOES push to a tag, e.g. to a tag such as `nlnetlabs/krill:adhoc-<short_commit_id>`.

That would both allow the full behaviour to be tested but also for us to be able to make a Docker image a customer could use if needed to test a specific fix or modification.

Any thoughts on this @timbru?

ximon18:
What I haven't done is Billy anything for Mac specifically, this work just built on the previous ARM work for Raspberry Pi (ARMv7/armhf) and Rock64 (ARM64/aarch64).

ximon18:
**Update:** This PR is now ready for review and merge. The workflow has had some minor changes since the screenshots were posted above but is basically still the same. There is no longer a cleanup step as deletion wasn't always working and anyway it can be useful to have the intermediate cross-compiled binaries and Docker images stored as artifacts along with the GitHub workflow. These intermediate artifacts are named `tmp-...` and are ignored by the packages.nlnetlabs.nl scripts which publish packages based on artifacts attached to a `pkg` workflow.

Example from a `main` branch (in my personal Krill fork) with publication to Docker Hub:
- https://github.com/ximon18/krill/actions/runs/2825179260
- https://hub.docker.com/repository/docker/ximoneighteen/krill/tags

Example from this PR branch in this repo (WITHOUT publication to Docker Hub because this branch is neither `main` nor a tag)
- https://github.com/NLnetLabs/krill/actions/runs/2825186098

A sanity check on my Raspberry Pi 4b of the multi-arch image published as `ximoneighteen/krill:unstable` worked fine, like so:
```
$ sudo docker run -it --rm --pull always ximoneighteen/krill:unstable krillc --version
```

I have marked parts of the comments above with **outdated** where applicable.

Issue #772, opened 2022-02-10
Disable ROA re-issuance on upgrade for instances with many CAs


timbru:
ROAs should be re-issued because they use long subject names. But.. all RPs accept these subject, and re-issuing ALL ROAs for krill instances with many CAs can result in outages because all RPs will then need to re-sync everything at once.

So, we can disable this, and the ROAs will be re-issued automatically when the time comes (4 weeks before expiry - which would be 52 weeks after issuance - by default)

PR #777, opened 2022-02-15
Support prepare-upgrade krill: prepares the new data to migrate.

timbru:
Changed the process to a two-pronged approach:

1. prepare new data (if needed)

This can be called through the command line by calling:

`krill prepare-upgrade -c krill.config`

If this `prepare-upgrade` sub-command is used then preparation is all that will be done. Data will be prepared under `$data_dir/upgrade_data/cas` and/or `$data_dir/upgrade_data/cas` in order to track progress. As a part of this process all events are replayed to ensure that they do in fact work, and a new snapshot is saved.

This step can be repeated any number of times and it will continue from where it got to. The process can also be restarted completely by deleting the `data_dir/upgrade_data` dir.

When krill is (re-)started with the new binary then it will do another prepare-data step, but this time it will clean up those status files as it's expected that no other krill instance is running at this point. Then we go to the next step:

2. finalise:

- update data dirs if applicable
- update data dir version number (always)


timbru:
Note for review: there are lots files added to test the migrations - they are probably less interesting.. the 0.7.3 data was generated by setting up a fresh 0.7.3 installation on a droplet under our testbed. My plan is to do the upgrade there as soon as we have an 0.9.5-rc1 version ready. And then do the following extra manual checks:

- backup data so tests can be repeated
- test CLI prepare
- make some changes
- test CLI prepare again
- upgrade krill
- and.. using the backed up data.. do a krill update without preparing.

src/daemon/http/server.rs
```diff
@@ -116,17 +111,19 @@ pub async fn start_krill_daemon(config: Arc<Config>) -> Result<(), Error> {
     test_data_dirs_or_die(&config);
 
     // Call upgrade, this will only do actual work if needed.
-    pre_start_upgrade(config.clone())?;
+    let upgrade_report = prepare_upgrade_data_migrations(UpgradeMode::PrepareThenFinalise, config.clone()).await?;
+    if let Some(report) = &upgrade_report {
+        finalise_data_migration(report.versions(), config.as_ref())?;
```
ximon18, line 116: ? Didn't we already finalize the migration by passing `UpgradeMode::PrepareThenFinalize' to `prepare_upgrade_data_migration()` ?

src/upgrades/mod.rs
```diff
@@ -40,6 +98,8 @@ pub enum UpgradeError {
 
 impl fmt::Display for UpgradeError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Upgrade failed. Check your data directory. If you find folders named 'arch-cas-[$version]' or 'arch-pubd-[$version]' there, then rename them to 'cas' and 'pubd' respectively and re-install your old $version of krill. Please contact us and let us know about this error. You can reach us at rpki-team@nlnetlabs.nl. Detailed error: ")?;
```
ximon18, line 101: Could/should this rename be handled automatically?

src/upgrades/mod.rs
```diff
@@ -86,30 +146,101 @@ impl From<crate::commons::error::Error> for UpgradeError {
...
+
+    fn version_before(&self, later: KrillVersion) -> Result<bool, UpgradeError> {
+        self.store()
+            .version_is_before(later)
+            .map_err(UpgradeError::KeyStoreError)
     }
 
     fn store(&self) -> &KeyValueStore;
 
+    fn new_store(&self) -> &KeyValueStore;
```
ximon18, line 200: From these names it isn't apparent what the difference is between `store()` and `new_store()`.

src/upgrades/mod.rs
```diff
@@ -86,30 +146,101 @@ impl From<crate::commons::error::Error> for UpgradeError {
...
+    fn remove_data_upgrade_info(&self, scope: &str) -> UpgradeResult<()> {
+        self.new_store()
+            .drop_key(&Self::data_upgrade_info_key(scope))
+            .map_err(UpgradeError::KeyStoreError)
+    }
+
     // Find all command keys and sort them by sequence.
     // Then turn them back into key store keys for further processing.
-    fn command_keys(&self, scope: &str) -> Result<Vec<KeyStoreKey>, UpgradeError> {
+    fn command_keys(&self, scope: &str, from: u64) -> Result<Vec<KeyStoreKey>, UpgradeError> {
```
ximon18, line 233: Is the comment about "find _all_ command keys" still correct if you now filter based on the given "from" ?

src/upgrades/mod.rs
```diff
@@ -129,94 +260,175 @@ pub trait UpgradeStore {
...
-        }
-
-        if self.store().has(&snapshot_bk_key)? {
-            self.archive_to_migration_scope(&snapshot_bk_key)?;
+/// Prepares a Krill upgrade related data migration. If no data migration is needed
+/// then this will simply be a no-op. Returns the [`KrillUpgradeVersions`] if the currently
+/// deployed Krill version differs from the code version. Note that the version may
+/// have increased even if there is no data migration needed.
+///
+/// In case data does not to be migrated, then the this data will be prepared under
```
ximon18, line 270: Should this say "does need to be migrated" instead of "does not to be migrated" ?

src/upgrades/mod.rs
```diff
@@ -129,94 +260,175 @@ pub trait UpgradeStore {
...
+/// the directory returned by `config.data_dir()`. By design, this migration can be
+/// executed while Krill is running as it does not affect any current state. If can
```
ximon18, line 272: s/If can/It can/

src/upgrades/mod.rs
```diff
@@ -129,94 +260,175 @@ pub trait UpgradeStore {
...
+/// be called multiple times and it will resume the migration from the point it got
```
ximon18, line 273: It can be called multiple times but not concurrently, right?

src/upgrades/mod.rs
```diff
@@ -129,94 +260,175 @@ pub trait UpgradeStore {
...
-
-    Ok(())
 }
 
-fn needs_v0_9_0_upgrade(work_dir: &Path, ns: &str) -> bool {
-    needs_upgrade(work_dir, ns, KrillVersion::release(0, 9, 0))
-}
-
-fn needs_upgrade(work_dir: &Path, ns: &str, before: KrillVersion) -> bool {
+fn upgrade_versions_ns(work_dir: &Path, ns: &str) -> Option<UpgradeVersions> {
```
ximon18, line 424: What is "ns" ?

src/upgrades/mod.rs
```diff
@@ -129,94 +260,175 @@ pub trait UpgradeStore {
...
+/// as there can be only one version running.
+fn upgrade_versions(config: &Config) -> Option<UpgradeVersions> {
+    let cas_version = upgrade_versions_ns(&config.data_dir, CASERVER_DIR);
+    let pubd_version = upgrade_versions_ns(&config.data_dir, PUBSERVER_DIR);
+
+    match (cas_version, pubd_version) {
+        (None, None) => None,
+        (Some(upgrade), None) | (None, Some(upgrade)) => Some(upgrade),
+        (Some(upgrade_cas), Some(upgrade_pubd)) => {
+            if upgrade_cas.from() >= upgrade_pubd.from() {
```
ximon18, line 415: Could `std::cmp::max()` be used here?

src/upgrades/mod.rs
```diff
@@ -129,94 +260,175 @@ pub trait UpgradeStore {
...
+    // Move directories - if applicable (servers can have cas, repo server or both)
+
+    let from = upgrade.from();
+    let current = upgrade.to();
+    let data_dir = &config.data_dir;
+    let upgrade_dir = config.upgrade_data_dir();
+
+    // cas -> arch-cas-{old-version}
+    // upgrade-data/cas -> cas
+    // upgrade-data/ca_objects -> ca_objects
```
ximon18, line 336: So is there no `ca_objects` dir prior to the upgrade then?

src/upgrades/mod.rs
```diff
@@ -129,94 +260,175 @@ pub trait UpgradeStore {
...
-}
 
-/// Should be called when Krill starts, before the KrillServer is initiated
-pub fn pre_start_upgrade(config: Arc<Config>) -> Result<(), UpgradeError> {
-    upgrade_data_to_0_9_0(config)
+    Ok(())
 }
 
-/// Should be called when the KrillServer is initiated, before the webserver is started
+/// Should be called when the KrillServer is initiated, before the web server is started
```
ximon18, line 390: s/initiated/initialised/?

ximon18:
A successful packaging workflow that packages the new `krillup` tool can be seen running against this PR branch here: https://github.com/NLnetLabs/krill/actions/runs/1857844878

src/upgrades/mod.rs
```diff
@@ -129,94 +260,175 @@ pub trait UpgradeStore {
...
 /// and operators can make changes.
-pub async fn post_start_upgrade(config: &Config, server: &KrillServer) -> Result<(), UpgradeError> {
-    if needs_upgrade(&config.data_dir, "cas", KrillVersion::candidate(0, 9, 3, 2)) {
+pub async fn post_start_upgrade(upgrade_versions: &UpgradeVersions, server: &KrillServer) -> Result<(), UpgradeError> {
+    if upgrade_versions.from() < &KrillVersion::candidate(0, 9, 3, 2) {
         info!("Reissue ROAs on upgrade to force short EE certificate subjects in the objects");
```
ximon18, line 394: Wasn't there some idea about not doing this?

src/upgrades/mod.rs
```diff
@@ -232,30 +444,62 @@ mod tests {
...
+        let config = Config::test(&work_dir, false, false, false);
         let _ = config.init_logging();
 
-        upgrade_data_to_0_9_0(config).unwrap();
+        let _upgrade = prepare_upgrade_data_migrations(UpgradeMode::PrepareOnly, Arc::new(config.clone()))
+            .await
+            .unwrap()
+            .unwrap();
+
+        // and continue - immediately, but still
```
ximon18, line 459: but still what?

src/upgrades/mod.rs
```diff
@@ -232,30 +444,62 @@ mod tests {
...
+        let report = prepare_upgrade_data_migrations(UpgradeMode::PrepareThenFinalise, Arc::new(config.clone()))
```
ximon18, line 460: If the mode is `PrepareThenFinalize` that sounds like `prepare_upgrade_data_migrations()` is misnamed as it doesn't just prepare but also finalizes?

src/upgrades/mod.rs
```diff
@@ -232,30 +444,62 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn prepare_then_upgrade_0_8_1_pubd_only() {
+        let source = PathBuf::from("test-resources/migrations/v0_8_1_pubd_only/");
+        test_upgrade(source).await;
+    }
+
+    #[tokio::test]
+    async fn test_upgrade_0_6_0() {
```
ximon18, line 489: Why are the other tests called `prepare_then_upgrade` but this one is not, even though this one does prepare then finalize? Ah is it because for the others you do `PrepareOnly` first and with this one you do not?

src/upgrades/v0_9_0/ca_objects_migration.rs
```diff
@@ -30,52 +29,47 @@ use crate::{
...
-                store,
-                ca_store,
-                repo_manager,
-                derived_embedded_ca_info_map,
-            }
-            .migrate()
-        } else {
-            debug!("Krill is up to date, no CA data migration needed.");
-            Ok(())
+        info!("Krill version is older than 0.9.0, will now upgrade CA data structures.");
```
ximon18, line 47: In the old code this statement was checked, now it is not. We could be called without this being the case. Is this the right place for this log statement now?

src/upgrades/v0_9_0/ca_objects_migration.rs
```diff
@@ -145,109 +146,102 @@ impl UpgradeStore for CasStoreMigration {
...
+                let old_init: OldCaIni = self.get(&init_key)?;
+                let (id, _, old_ini_det) = old_init.unpack();
+                let (rfc_8183_id, repo_opt, ta_opt) = old_ini_det.unpack();
+                let ini = IniDet::new(&id, rfc_8183_id.into());
+                self.new_kv_store.store(&init_key, &ini)?;
+
+                // If the CA was initialized as a trust anchor, then we refuse to upgrade.
+                // This can only happen if this is a very old test system. People will need
+                // to set up a new test system instead.
+                if ta_opt.is_some() {
```
ximon18, line 181: If we are going to refuse to upgrade should we have already added the init even then?

src/upgrades/v0_9_0/ca_objects_migration.rs
```diff
@@ -145,109 +146,102 @@ impl UpgradeStore for CasStoreMigration {
...
-                    info.last_command,
-                    StorableCaCommand::RepoUpdate { service_uri },
-                    StoredEffect::Success { events: vec![1] },
-                );
+                    let contact = RepositoryContact::new(repo_response);
+                    let service_uri = contact.service_uri().clone();
 
-                info.last_command += 1;
-                info.last_update = time_for_init_command;
+                    data_upgrade_info.last_event += 1;
```
ximon18, line 208: In the old code you incremented this AFTER creating the stored CA command, now you do it before. Was there an off by one error then?

src/upgrades/v0_9_0/ca_objects_migration.rs
```diff
@@ -267,32 +261,45 @@ impl UpgradeStore for CasStoreMigration {
...
+                // Read and parse the old command.
+                let mut old_cmd: OldStoredCaCommand = self.get(&old_cmd_key)?;
 
+                // If the command was a success, then it will have events. Successful commands
+                // that resulted in no changes are simply not recorded. That said, it may turn
+                // out that the event list to migrate is empty - in that case we skip this command
+                // and continue the command loop.
+                //
+                // Note that if the command resulted in an error we do not not get 'Some' empty
+                // vec of events, but we get a None. So such commands will be migrated.
```
ximon18, line 273: will NOT be migrated? there's no else block handling the None case.

src/upgrades/v0_9_0/pubd_objects_migration.rs
```diff
@@ -46,28 +46,42 @@ impl PubdObjectsMigration {
...
+    /// resuming the migration.
+    ///
+    /// NOTE: this code is not called if the migration to 0.9.x would be complete.
     fn populate_repo_content(config: Arc<Config>) -> UpgradeResult<()> {
+        info!("Populate the repository content based on current state");
         let old_store = AggregateStore::<OldRepository>::disk(&config.data_dir, PUBSERVER_DIR)?;
         if old_store.warm().is_err() {
-            // this is most likely because the info last event is off by one, try deleting the info
+            // This is most likely because the info last event is off by one which affected
+            // some old version. Try archiving the info.json to force that the publication
```
ximon18, line 80: Ah maybe this references the off-by-one comment I made above?

src/upgrades/v0_9_0/pubd_objects_migration.rs
```diff
@@ -173,31 +191,46 @@ impl UpgradeStore for PubdStoreMigration {
...
+            // Read and parse the old command.
             let mut old_cmd: OldStoredRepositoryCommand = self.get(&old_cmd_key)?;
 
+            // If the command was a success, then it will have events. Successful commands
+            // that resulted in no changes are simply not recorded. That said, it may turn
+            // out that the event list to migrate is empty - in that case we skip this command
+            // and continue the command loop.
+            //
+            // Note that if the command resulted in an error we do not not get 'Some' empty
+            // vec of events, but we get a None. So such commands will be migrated.
```
ximon18, line 209: Same comment about None as above, as None isn't handled here is it correct to say "will be migrated" instead of "will NOT be migrated" ?

ximon18 (approved):
LGTM - I did not look at the test-resources dir content.

src/bin/krill.rs
```diff
@@ -2,14 +2,34 @@ extern crate krill;
...
+        )
+        .get_matches();
+
+    let config_file = matches.value_of("config").unwrap_or(KRILL_DEFAULT_CONFIG_FILE);
+
+    match Config::create(config_file, false) {
         Ok(config) => {
             if let Err(e) = server::start_krill_daemon(Arc::new(config)).await {
-                eprintln!("Krill failed to start: {}", e);
+                error!("Krill failed to start: {}", e);
```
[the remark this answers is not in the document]
timbru: yes, this is done as when the `config` is created. So if we have one, then it's initialised.

src/upgrades/v0_9_0/ca_objects_migration.rs
timbru: along the way, I needed it for when prepping the data for the upgrade.

src/upgrades/v0_9_0/ca_objects_migration.rs
timbru: no, but it was all subtly different because I was updating another file in place. Here I am tracking how far the migration got and the `info` is only updated at the end.

src/bin/krill.rs
```diff
@@ -2,14 +2,34 @@ extern crate krill;
...
 async fn main() {
-    match server::parse_config() {
+    let matches = App::new(KRILL_SERVER_APP)
+        .version(KRILL_VERSION)
+        .arg(
+            Arg::with_name("config")
+                .short("c")
+                .long("config")
+                .value_name("FILE")
+                .help("Override the path to the config file (default: './defaults/krill.conf')")
```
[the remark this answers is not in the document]
timbru: wel.. quick fix. done

src/bin/krillup.rs
```diff
@@ -0,0 +1,65 @@
...
+async fn main() {
+    let matches = App::new(KRILL_UP_APP)
+        .version(KRILL_VERSION)
+        .about("\nThis tool can be used to reduce the risk and time needed for Krill upgrades, by preparing and verifying any data migrations that would be needed. The data_dir setting from the provided configuration file is used to find the data to migrate, and prepared data will be saved under 'data_dir/upgrade-data'.")
+        .arg(
+            Arg::with_name("config")
+                .short("c")
+                .long("config")
+                .value_name("FILE")
+                .help("Override the path to the config file (default: './defaults/krill.conf')")
```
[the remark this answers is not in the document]
timbru: quick fix, done

src/bin/krill.rs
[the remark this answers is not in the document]
ximon18: Hmm, I thought when I looked at `Config::create()` that it seemed to me that it could bail out before the logging was initialized if initial parsing failed? It says `Self::read_config(config_file)?` and then only after that `config.init_logging()?` ??

src/daemon/config.rs
```diff
@@ -700,68 +703,52 @@ impl Config {
...
-    pub fn create() -> Result<Self, ConfigError> {
-        let config_file = Self::get_config_filename();
+        if upgrade_only {
+            info!("Prepare upgrade using configuration file: {}", config_file,);
+            info!("Processing data from: {}", config.data_dir.to_string_lossy());
+            info!(
+                "Saving prepared data to: {}",
+                config.upgrade_data_dir().to_string_lossy()
+            );
+            warn!("*** Kill this process if you have doubts about available disk space! ***");
```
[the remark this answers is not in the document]
timbru: I have no good way to guess.. so that's why it's left as an exercise to the operator... I could also leave out the warning altogether but that also seemed wrong. In most cases this really shouldn't be an issue, and people should monitor disk space.. suggestion?

src/commons/util/file.rs
```diff
@@ -28,12 +28,20 @@ pub fn sub_dir(base: &Path, name: &str) -> Result<PathBuf, KrillIoError> {
 
 pub fn create_dir(dir: &Path) -> Result<(), KrillIoError> {
     if !dir.is_dir() {
-        fs::create_dir(dir)
+        fs::create_dir_all(dir)
```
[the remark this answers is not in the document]
timbru: I needed it for this PR because it wants to create dirs that are some levels deep, and it does not impact existing code.

src/daemon/config.rs
```diff
@@ -858,13 +845,20 @@ impl Config {
 
     pub fn read_config(file: &str) -> Result<Self, ConfigError> {
         let mut v = Vec::new();
-        let mut f =
-            File::open(file).map_err(|e| KrillIoError::new(format!("Could not read open file '{}'", file), e))?;
+        let mut f = File::open(file).map_err(|e| {
+            KrillIoError::new(
+                format!(
+                    "Could not read config file '{}'. Please use --config to specify its location.",
```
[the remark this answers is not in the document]
timbru: this is because it's optional and it might be the default file that it can't find - but will rephrase a bit

src/daemon/config.rs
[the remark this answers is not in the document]
ximon18: Why is this different than any other time that Krill might encounter errors writing to disk? Why mention it at all?

src/upgrades/mod.rs
timbru: indeed

src/upgrades/mod.rs
timbru: Issue #772. It's conditional - will skip for instances with many CAs. After this is merged

src/upgrades/mod.rs
timbru: I just wanted to hit both code paths here

ximon18:
Here's a screenshot showing a Docker Debian Bullseye container running latest published Krill 0.9.4 DEB and `krillup` installed from a DEB built by the GitHub Actions packaging workflow changes in this PR:

image

src/upgrades/v0_9_0/pubd_objects_migration.rs
timbru: no - this an off by one in versions I have forgotten about already. The info file is a helper but krill can work it out without it.

src/commons/util/file.rs
```diff
@@ -28,12 +28,20 @@ pub fn sub_dir(base: &Path, name: &str) -> Result<PathBuf, KrillIoError> {
...
             .map_err(|e| KrillIoError::new(format!("could not create dir: {}", dir.to_string_lossy()), e))?;
     }
     Ok(())
 }
 
+pub fn remove_dir_all(dir: &Path) -> Result<(), KrillIoError> {
+    if dir.exists() {
+        fs::remove_dir_all(dir)
+            .map_err(|e| KrillIoError::new(format!("could not remove-all dir: {}", dir.to_string_lossy()), e))?;
```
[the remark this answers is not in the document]
timbru: done

src/upgrades/mod.rs
timbru: Improved the error reporting. It will only talk about this now if the finalize step failed. We cannot recover from that failure so leaving it to the user. It is very unlikely to fail though because essentially we are then just renaming dirs, u updating version files and cleaning up. I/o issues would have come up before this step.

src/daemon/http/server.rs
timbru: renamed it to PrepareToFinalise

ximon18 (approved):

src/upgrades/mod.rs
timbru: indeed, renamed to PrepareToFinalise

ximon18 (approved):

src/upgrades/mod.rs
timbru: renamed

ximon18 (approved):

src/upgrades/mod.rs
timbru: right - the lockfile still needs to be done before closing this. Leaving this open as a reminder

src/upgrades/mod.rs
timbru: name space - used a lot like this in the code for brevity

ximon18 (approved):

src/upgrades/v0_9_0/ca_objects_migration.rs
timbru: should not mention the version again - modified

src/upgrades/v0_9_0/ca_objects_migration.rs
timbru: hopefully more clear now..

                // Note that if the command resulted in an error we do not not get 'Some' empty
                // vec of events, but we get a None. So we skip this block. Error commands will
                // be migrated - we won't call continue as they are expected to have no events.

ximon18:
This is what the Docker experience looks like when using `krillup` which is part of the same image as Krill itself:

image

To actually run `krillup` I assume that a user of Docker has a persistent data volume attached to their Krill Docker container which can be mounted into a new container run from the newer Docker image in which `krillup` can be run. Alternatively an operator could copy the `krillup` binary out of the new image into an existing container and run it like that...

ximon18 (approved):

src/upgrades/mod.rs
timbru: I added lockfile support in the latest commit. We use similar code in krill-sync.

ximon18:
@timbru: Locking looks good.

Issue #775, opened 2022-02-14
Improve 'parent reduced not after time' check when requesting certificates


timbru:
It looks like Krill will sometimes conclude -wrongly- that the not-after time is reduced. This can result in noise, and more importantly in requesting too many certificates.

Looking at the code there is 1 obvious issue, and 1 obvious improvement:
1. We should use 'second' level time precision, not milis.
2. We should probably use a 10% remaining time decrease threshold before requesting a cert, just like we do when requesting certs because the time was extended

PR #778, opened 2022-02-15
Check better if not-after is really reduced, significantly. (#775)

timbru:


src/daemon/ca/keys.rs
```diff
@@ -124,22 +124,20 @@ impl CertifiedKey {
...
-        } else if until_not_after_millis == until_new_not_after_millis {
+        } else if until_not_after == until_new_not_after {
             debug!(
                 "Will not request new certificate for CA '{}' under RC '{}'. Resources and not after time are unchanged.",
                 handle,
                 rcn,
             );
             false
-        } else if until_new_not_after_millis < until_not_after_millis {
+        } else if until_not_after > 0 && (until_new_not_after as f64 / until_not_after as f64) < 0.9_f64 {
```
ximon18, line 134: With these comparisons, as the timestamp value gets further into the future the same number of difference in seconds between `until_not_after` and `until_new_not_after` will change as a percentage, e.g. 200/240=0.83 but 400/440=0.90. If we take the unix epoch timestamp in seconds now as being 1645015173 then even roughly a month of difference won't be 10% in epoch time: (1645015173 - 86400*30) / 1645015173 = 0.998 which is not `< 0.9_f64`. I'm not sure if I've understood this correctly. Is this what is wanted?

src/daemon/ca/keys.rs
```diff
@@ -107,11 +107,11 @@ impl CertifiedKey {
 
         let not_after = self.incoming_cert().cert().validity().not_after();
 
-        let now = Time::now().timestamp_millis();
-        let until_not_after_millis = not_after.timestamp_millis() - now;
-        let until_new_not_after_millis = new_not_after.timestamp_millis() - now;
+        let now = Time::now().timestamp();
```
ximon18, line 110: Regarding the comment above, which wasn't changed as part of this work but with the goal of ensuring everything is correct and clear:

1. What does "we probably should know.." mean? Should know what? Did that mean to say "because we probably should _you_ know" as in we should act because we've been asked to?
2. Is it a) or b), or is it a) and b)?
3. Should the comment be updated to mention the 10% checks?
4. Should the comment, and the comment just below as well, be updated to refer to the new remaining/eligibility based naming?

ximon18 (approved):

src/daemon/ca/keys.rs
ximon18: The rename to "remaining" time made it clear that the numbers compared are not absolute unix epoch timestamps but seconds remaining and thus do not lead to huge variable differences and thus it is now clear that this does what was intended.

One thought: is this 10% behaviour covered somewhere by test cases?

src/daemon/ca/keys.rs
timbru: fixed

PR #779, opened 2022-02-16
Add serde defaults suspended etc

timbru:


src/daemon/ca/certauth.rs
```diff
@@ -102,17 +89,11 @@ pub struct CertAuth {
...
 
-    #[serde(
-        skip_serializing_if = "AspaDefinitions::is_empty",
-        default = "AspaDefinitions::default"
-    )]
+    #[serde(skip_serializing_if = "AspaDefinitions::is_empty", default)]
     aspas: AspaDefinitions,
-
-    #[serde(skip_serializing, default = "CertAuthStatus::default")]
-    status: CertAuthStatus,
```
ximon18, line 115: What was the motivation for removing this field entirely? Is it unused?

src/daemon/ca/events.rs
```diff
@@ -427,9 +427,16 @@ impl AspaObjectsUpdates {
 /// Describes an update to the set of ROAs under a ResourceClass.
 #[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
 pub struct ChildCertificateUpdates {
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
```
ximon18, line 430: These fields were not previously omitted from the serialization. Will the deserializing party handle this difference?

src/daemon/ca/rc.rs
```diff
@@ -52,6 +52,8 @@ pub struct ResourceClass {
     parent_rc_name: ResourceClassName,
 
     roas: Roas,
+
+    #[serde(skip_serializing_if = "AspaObjects::is_empty", default)]
```
ximon18, line 56: This field was not previously omitted from the serialization. Will the deserializing party handle this difference?

src/daemon/ca/certauth.rs
timbru: It is unused. It was there earlier because earlier versions had this and it could be present in snapshots. But snapshots are rebuilt in the migration to 0.9.x so this is now safe to remove completely.

src/daemon/ca/events.rs
timbru: yes, that would be here and 'default' ensures this.

src/daemon/ca/rc.rs
timbru: yes, because of default. I think it's better to remove this if empty to lower the noise. Few users will use ASPA at this point and the presence might confuse people actually.

Issue #783, opened 2022-02-18
Clean up upgrade directory from previous releases


timbru:
If it is found that the 'upgrade-data' directory was created by a version of krill that differs from the current version preparing an update.. then issue a warning and remove the dir to start the new preparation from scratch.

Note that the presence of an 'upgrade-data' dir from a previous version marks an unfinished migration - so it's safe to delete.

PR #785, opened 2022-02-18
Clean up upgrade directory from previous releases #783

timbru:


src/commons/eventsourcing/kv.rs
```diff
@@ -91,11 +91,19 @@ impl KeyValueStore {
...
 
-        if !base.exists() {
-            file::create_dir(&base)?;
+        let store = KeyValueStore::Disk(KeyValueStoreDiskImpl { base });
+
+        match &store {
+            KeyValueStore::Disk(disk_store) => {
+                // If this is a new store then initialise the disk and set the version
+                if !disk_store.base.exists() {
+                    file::create_dir(&disk_store.base)?;
```
ximon18, line 100: Should this be `create_dir_all()`?

src/commons/eventsourcing/kv.rs
```diff
@@ -360,6 +389,14 @@ impl KeyValueStoreDiskImpl {
         Ok(())
     }
 
+    pub fn wipe(&self) -> Result<(), KeyValueError> {
+        if self.base.exists() {
+            file::remove_dir_all(&self.base)?;
+            file::create_dir(&self.base)?;
```
ximon18, line 395: Is there a `create_dir_all()` that should be used here?

ximon18 (approved):

src/commons/eventsourcing/kv.rs
timbru: `file::create_dir` uses `fs::create_dir_all` under the hood. It could be renamed I suppose

PR #789, opened 2022-02-22
0.9.5 rc1 fix migration

timbru:
This PR fixes some issues found in further testing of krillup against known data of large deployments.


Cargo.toml
```diff
@@ -45,8 +45,8 @@ rand                  = "^0.8"
 regex                 = { version = "^1.4", optional = true, default_features = false, features = ["std"] }
 reqwest               = { version = "0.11", features = ["json"] }
 rpassword             = { version = "^5.0", optional = true }
-rpki                  = { version = "0.13.1", features = [ "repository", "rrdp", "serde" ] }
-# rpki                  = { version = "0.13.1-rc1", git = "https://github.com/NLnetLabs/rpki-rs/", features = [ "repository", "rrdp", "serde" ] }
+rpki                  = { version = "0.13.2", features = [ "compat", "repository", "rrdp", "serde" ] }
+# rpki                  = { version = "0.13.2-dev", git = "https://github.com/NLnetLabs/rpki-rs/", branch = "compat-for-prefix-lens", features = [ "compat", "repository", "rrdp", "serde" ] }
```
ximon18, line 49: This line can be removed, right?

timbru: I am keeping it for convenience whenever I need to depend on a development version. Is it in the way?

ximon18:
It just offends my wonky painting sensibility ;-)

Cargo.toml
ximon18: It just offends my wonky painting sensibility ;-)

src/commons/api/ca.rs
```diff
@@ -773,7 +773,15 @@ pub struct Revocations(Vec<Revocation>);
 
 impl Revocations {
     pub fn to_crl_entries(&self) -> Vec<CrlEntry> {
-        self.0.iter().map(|r| CrlEntry::new(r.serial, r.expires)).collect()
+        // Todo: include the revocation time in ['Revocation'] and use that.
```
ximon18, line 776: s/Todo/TODO/ ?

ximon18 (approved):

timbru:
> Doesn't this hack mean that `purge()` below will consider all CRL entries expired next time it is called because (assuming the clock is monotonic) `Time::now() >= r.expires` will hold?

No it doesn't, because `purge()` looks at the contained `Vec<Revocation>` - i.e. our own type rather than the values which are includes on the CRL object - which are produced by `fn to_crl_entries`.

The longer term solution is to extend our internal `Revocation` type and add something like `since`:

```
pub struct Revocation {
    serial: Serial,
    since: Time,
    expires: Time,
}
```

Then we can use `since` for the value on the actual CRL objects and keep using `expires` to purge entries which are no longer relevant.

PR #803, opened 2022-03-15
V0.9.5 rc7 prep

timbru:


src/commons/api/ca.rs
```diff
@@ -1503,16 +1490,11 @@ impl From<&IssuedCert> for ParentStatusCert {
 pub struct ParentStatus {
     last_exchange: Option<ParentExchange>,
     last_success: Option<Timestamp>,
-    next_exchange_before: Timestamp,
```
ximon18, line 1506: Is this a breaking API change?

src/commons/eventsourcing/store.rs
```diff
@@ -422,15 +422,16 @@ where
 
         let stored_command_builder = StoredCommandBuilder::new(&cmd, latest.version(), info.last_command);
 
-        let res = match latest.process_command(cmd) {
+        match latest.process_command(cmd) {
             Err(e) => {
                 let stored_command = stored_command_builder.finish_with_error(&e);
                 self.store_command(stored_command)?;
+                self.save_info(&handle, &info)?; // last_command was updated
```
ximon18, line 429: Which bug does this fix?

src/daemon/mq.rs
```diff
@@ -57,96 +58,135 @@ pub enum QueueTask {
...
+            }
+            Some(existing_priority) => {
+                if existing_priority < priority {
+                    debug!(
+                        "Re-prioritising task: {} from: {} to: {}",
+                        task, existing_priority, priority
+                    );
+                    q.change_priority(&task, priority);
+                } else {
+                    debug!("Keeping existing task: {} with higher priority: {}", task, priority);
```
ximon18, line 146: or equal

ximon18 (commented):
I still need to finish reviewing the `scheduler.rs` changes.

Has anything changed in the Lagosta CSS or JS files? I can't tell from the diff but I suspect they've been recompiled without changes to the underlying Lagosta input files. Is that correct?

src/daemon/scheduler.rs
```diff
@@ -1,145 +1,101 @@
...
+            system_actor,
+            started: Timestamp::now(),
         }
     }
-}
 
-#[allow(clippy::cognitive_complexity)]
-fn make_cas_event_triggers(event_queue: Arc<MessageQueue>, ca_manager: Arc<CaManager>, actor: Actor) -> ScheduleHandle {
-    let started = Timestamp::now();
+    /// Run the schedular in the background. It will sweep the message queue for tasks
```
ximon18, line 59: s/schedular/scheduler/

src/daemon/scheduler.rs
```diff
@@ -169,230 +126,161 @@ fn make_cas_event_triggers(event_queue: Arc<MessageQueue>, ca_manager: Arc<CaMan
...
-            );
-            if let Err(e) = ca_server.renew_objects_all(&actor).await {
-                error!("Background re-issuing of about to expire objects failed: {}", e);
+    /// Queues tasks for background jobs when the server is started
+    async fn queue_start_tasks(&self) -> KrillResult<()> {
+        // If there are only a few CAs in this krill instance, then we
+        // will just want to re-sync them with their parents and repository
+        // on start up.
+        //
+        // If there are many, then will apply some random delays (jitter)
```
ximon18, line 140: s/then will/then/ or s/then will/then we will/

src/daemon/scheduler.rs
```diff
@@ -169,230 +126,161 @@ fn make_cas_event_triggers(event_queue: Arc<MessageQueue>, ca_manager: Arc<CaMan
...
+        let cas = ca_list.cas();
+
+        debug!("Adding tasks at start up");
+
+        let mut use_jitter = cas.len() >= SCHEDULER_USE_JITTER_CAS_THRESHOLD;
+
+        for summary in cas {
+            let ca = self.ca_manager.get_ca(summary.handle()).await?;
+
+            use_jitter = use_jitter || ca.nr_parents() >= self.config.ca_refresh_parents_batch_size;
```
ximon18, line 154: Should this be `let use_jitter = use_jitter...` ? As written I _think_ that a decision made for one CA whether to use jitter or not will affect the decision made for subsequent CAs in the loop but not previous CAs in the loop which seems a bit inconsistent.

src/daemon/scheduler.rs
```diff
@@ -169,230 +126,161 @@ fn make_cas_event_triggers(event_queue: Arc<MessageQueue>, ca_manager: Arc<CaMan
...
+
+            debug!("Adding tasks for CA {}, using jitter: {}", ca.handle(), use_jitter);
+
+            // Plan a regular sync for each parent. Spread these out if there
+            // are too many CAs or parents for a CA. In cases where there are only
+            // a handful of CAs/parents, this 'ca_refresh_start_up' will be 'now'.
+            // Note: users can override using the 'bulk' functions.
```
ximon18, line 161: I'm not sure that "override" is the correct word here, it implies that users can stop this behaviour from occuring which is not the case, they can only cause additional things to happen, right?

src/daemon/scheduler.rs
```diff
@@ -169,230 +126,161 @@ fn make_cas_event_triggers(event_queue: Arc<MessageQueue>, ca_manager: Arc<CaMan
...
+        let mut use_jitter = cas.len() >= SCHEDULER_USE_JITTER_CAS_THRESHOLD;
+
+        for summary in cas {
+            let ca = self.ca_manager.get_ca(summary.handle()).await?;
+
+            use_jitter = use_jitter || ca.nr_parents() >= self.config.ca_refresh_parents_batch_size;
+
+            debug!("Adding tasks for CA {}, using jitter: {}", ca.handle(), use_jitter);
+
+            // Plan a regular sync for each parent. Spread these out if there
```
ximon18, line 158: Technically the won't necessarily be spread (i.e. evenly) out in the range, they will randomly (which could be uneven, or even clumped) adjusted within the range. If the desire is to spread them evenly then the logic should be changed.

README.md
```diff
@@ -54,7 +54,7 @@ In addition to this we added a few other quick fixes in this release:
 - Connect to local parent directly #791
 - Do not sign/validate RFC6492 messages to/from local parent #797
 - Use per CA locking for CA statuses #795
-- Make CA sync strategy configurable #799
```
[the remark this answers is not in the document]
timbru: It was replaced by a different strategy:
+ Decrease CA update frequency and use jitter to spread load #802

src/daemon/config.rs
```diff
@@ -544,6 +545,38 @@ impl Config {
...
+        Self::ca_refresh_next_from(self.ca_refresh_seconds, self.ca_refresh_jitter_seconds)
+    }
+
+    pub fn ca_refresh_start_up(&self, use_jitter: bool) -> Priority {
+        let jitter_seconds = if use_jitter { self.ca_refresh_jitter_seconds } else { 0 };
+
+        Self::ca_refresh_next_from(0, jitter_seconds)
+    }
+
+    fn ca_refresh_next_from(regular_seconds: i64, jitter_seconds: i64) -> Priority {
```
ximon18, line 568: Should `jitter_seconds` be `u64`? The 0..n range expression won't work if it is negative, right?

README.md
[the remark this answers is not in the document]
ximon18: Okay, but aren't they orthogonal and not mutually exclusive? One doesn't have to require the removal of the other, so is it the case that parallelization was implemented but is not actually needed?

src/daemon/scheduler.rs
```diff
@@ -169,230 +126,161 @@ fn make_cas_event_triggers(event_queue: Arc<MessageQueue>, ca_manager: Arc<CaMan
...
+            // Note: if circumstances dictate a sync before it's planned, e.g.
+            // because ROAs are changed, then it will be rescheduled accordingly.
+            // Note: users can override using the 'bulk' functions.
+            self.tasks
+                .sync_repo(ca.handle().clone(), self.config.ca_refresh_start_up(use_jitter));
+
+            // If suspension is enabled then plan a task for it. Since this is
+            // a cheap no-op in most cases, we do not need jitter. If we do not
+            // add this task then it will not be executed (obviously), but more
+            // importantly.. by adding this task we ensure that it will keep being
```
ximon18, line 182: I can't find in the previous code where child suspension was done periodically, only on CA refresh. What's the rationale for adding it as a recurring background task now?

defaults/krill.conf
```diff
@@ -218,12 +218,26 @@
...
 #
-### ca_refresh_seconds = 600
+# This value is specified in seconds for compatability with earlier krill versions.
+#
+### ca_refresh_seconds = 86400
+
+# CA certificate refresh jitter
+#
+# In order to avoid that many child CAs contact their parent at the same time
+# krill adds a random extra 'jitter' time between 0 and the number of seconds
```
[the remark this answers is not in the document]
timbru: fixed - in more places actually

src/daemon/scheduler.rs
```diff
@@ -1,145 +1,101 @@
...
-#[allow(clippy::cognitive_complexity)]
-fn make_cas_event_triggers(event_queue: Arc<MessageQueue>, ca_manager: Arc<CaManager>, actor: Actor) -> ScheduleHandle {
-    let started = Timestamp::now();
+    /// Run the schedular in the background. It will sweep the message queue for tasks
+    /// and re-schedule new tasks as needed.
+    pub async fn run(&self) -> KrillResult<()> {
+        loop {
+            while let Some(evt) = self.tasks.pop(now()) {
+                match evt {
+                    Task::ServerStarted => self.queue_start_tasks().await?, // return error and stop server on failure
```
ximon18, line 65: I've tried to compare the old scheduler setup to the new priority task setup and came up with this summary:

Task::CheckSuspendChildren: NEW: Wasn't periodic, is now?
Task::RefreshAnnouncementsInfo: CHANGED: Was cron every 5 seconds, but is now a recurring task on the queue every 10 minutes?
Task::RenewObjectsIfNeeded: UNCHANGED
Task::RepublishIfNeeded: CHANGED: Was cron every 10 minutes, but is now a recurring task on the queue every 5 minutes?
QueueTask::RescheduleSyncParent: WAS UNUSED SO REMOVED?
QueueTask::RescheduleSyncRepo: WAS UNUSED SO REMOVED?
QueueTask::ResourceClassRemoved: UNCHANGED
QueueTask::ServerStarted: UNCHANGED
Task::SweepLoginCache: UNCHANGED
QueueTask::SyncParent: UNCHANGED
QueueTask::SyncRepo: UNCHANGED
QueueTask::UnexpectedKey: UNCHANGED

src/commons/api/ca.rs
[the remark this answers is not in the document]
timbru: Good point - although I don't think this would be used much - I added a bit to the readme on the changes in ca-parent synchronisation

timbru: yes, well - we never committed to this being stable so I think it's okay with the updated readme.

defaults/krill.conf
```diff
@@ -181,7 +181,7 @@
 # krill_cas_bgp_announcements_invalid_length{ca="ca"}     number of announcements seen for CA resources with RPKI state INVALID (prefix exceeds max length)
 # krill_cas_bgp_announcements_not_found{ca="ca"}          number of announcements seen for CA resources with RPKI state NOT FOUND (none of the CA's ROAs cover this)
 # krill_cas_bgp_roas_too_permissive{ca="ca"}              number of ROAs for this CA which allow excess announcements (0 may also indicate that no BGP info is available)
-# krill_cas_bgp_roas_redundant{ca="ca"}                   number of ROAs for this CA which are redundant (0 may also indicate that no BGP info is available)
+# krill_cas_bgp_roas_redundant{ca="a"}                   number of ROAs for this CA which are redundant (0 may also indicate that no BGP info is available)
```
ximon18, line 184: Why did `ca="ca"` change to `ca="a"` ?

src/commons/eventsourcing/store.rs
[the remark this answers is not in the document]
timbru: No bug in this case. Just moved this logic up into the match because of the other change further down. Now we just let the match return.

src/commons/eventsourcing/store.rs
```diff
@@ -506,6 +507,9 @@ where
 
                     cache.insert(handle.clone(), Arc::new(agg.clone()));
 
+                    // Save the info so that the aggregate can be loaded properly by the listeners.
+                    self.save_info(&handle, &info)?;
```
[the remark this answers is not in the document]
timbru: With the changes in ca-parent sync scheduling it could happen that the event listener saw an update *before* the info file was committed - and it would then try to use an older version of the CA.

I think that this existed before as well except that it was very rarely hit - because the tasks were added and only checked every couple of seconds. So there was normally a delay then.

src/constants.rs
```diff
@@ -68,10 +68,14 @@ pub const KRILL_CLI_API_ENV: &str = "KRILL_CLI_API";
...
-pub const CA_REFRESH_SECONDS_MIN: u32 = 600;
-pub const CA_REFRESH_SECONDS_MAX: u32 = 3600;
-pub const CA_SUSPEND_MIN_HOURS: i64 = 2;
+pub const CA_REFRESH_SECONDS_MIN: i64 = 3600;
+pub const CA_REFRESH_SECONDS_MAX: i64 = 3 * 24 * 3600; // 3 days
+pub const CA_SUSPEND_MIN_HOURS: i64 = 48; // at least 2 days
+pub const SCHEDULER_REQUEUE_DELAY_SECONDS: i64 = 300;
+pub const SCHEDULER_USE_JITTER_CAS_THRESHOLD: usize = 50;
+pub const SCHEDULER_USE_JITTER_CAS_PARENTS_THRESHOLD: usize = 5;
+pub const SCHEDULER_INTERVAL_REPUBLISH_MINS: i64 = 5;
```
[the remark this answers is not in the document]
timbru: No need to mention as this is normally a very quick no-op. It is just a check whether new MFT/CRLs should be issued. As such it's also fine to do it a little more frequently now. This does not involve actual publication unless new objects are generated.

src/daemon/scheduler.rs
```diff
@@ -169,230 +126,161 @@ fn make_cas_event_triggers(event_queue: Arc<MessageQueue>, ca_manager: Arc<CaMan
...
-        self.state.write().unwrap().done();
-    }
+            error!(
+                "Failed to publish for '{}'. Will reschedule to: '{}'. Error: {}",
+                ca, next, e
+            );
 
-    fn is_running(&self) -> bool {
-        self.state.read().unwrap().is_running()
+            self.tasks.sync_repo(ca, next);
```
ximon18, line 209: Should this also be done in an else branch, like in `sync_parent()` below?

src/daemon/ca/manager.rs
```diff
@@ -814,41 +821,26 @@ impl CaManager {
         Ok(())
     }
 
-    /// Refresh all CAs:
-    /// - process all CAs
-    /// - process all parents for each CAs in parallel
-    ///    - send pending requests if present, or
-    ///    - ask parent for updates and process if present
+    /// Schedule refreshing all CAs asap:
```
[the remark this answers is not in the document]
timbru: done

src/daemon/ca/manager.rs
```diff
@@ -814,41 +821,26 @@ impl CaManager {
...
     ///
-    /// Note: this function can be called manually through the API, but is normally
-    ///       triggered in the background, every 10 mins by default, or as configured
-    ///       by 'ca_refresh' in the configuration.
-    pub async fn cas_refresh_all(&self, started: Timestamp, actor: &Actor) {
+    /// Note: this function can be called manually through the API, but normally the
+    ///       CA refresh process is replanned on the taskqueue automatically.
```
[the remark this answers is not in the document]
timbru: done

ximon18 (approved):

src/daemon/config.rs
timbru: yes.. probably.. I created issue #808 for this. This affects more settings that we should review.

test-resources/krill-init-multi-user.conf
```diff
@@ -218,12 +218,26 @@ service_uri = "https://localhost:3001/"
...
+
+# CA certificate refresh jitter
+#
+# In order to avoid that many child CAs contact their parent at the same time
+# krill adds a random extra 'jitter' time between 0 and the number of seconds
+# specified using 'ca_refresh_jitter_seconds'.
+#
+# Defaults to 12 hours: 43200 seconds.
+#
+# Values are trimmed to a maximum of: 50% of ca_refresh_seconds
```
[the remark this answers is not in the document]
timbru: done

src/daemon/mq.rs
```diff
@@ -223,3 +270,63 @@ impl eventsourcing::PostSaveEventListener<CertAuth> for MessageQueue {
...
+/// time value which is soonest has the highest priority. So, in short reverse
+/// order.
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct Priority(i64);
+
+pub fn now() -> Priority {
+    Time::now().into()
+}
+
+pub fn in_seconds(secs: i64) -> Priority {
```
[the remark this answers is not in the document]
timbru: Because it ultimately matches the signature of a function that takes this. But see #808

src/daemon/scheduler.rs
```diff
@@ -169,230 +126,161 @@ fn make_cas_event_triggers(event_queue: Arc<MessageQueue>, ca_manager: Arc<CaMan
...
-    if let Err(e) = ca_manager.cas_repo_sync_single(&ca).await {
-        let requeue_time = if test_mode_enabled() {
-            requeue_time_test()
-        } else {
-            requeue_time()
-        };
-
-        error!("Failed to publish for '{}' will reschedule, error: {}", ca, e);
-        event_queue.reschedule_sync_repo(ca, requeue_time);
+            sleep(Duration::from_millis(500)).await;
```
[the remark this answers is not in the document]
timbru: It doesn't run max 2 a second. It waits 500ms when the queue is empty. And then tries getting things from the queue until it's empty again (see `while let` on line 63).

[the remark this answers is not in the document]
ximon18: Ok, right, but why wait at all? Is this wait new or does it replace equivalent behaviour that existed previously, or?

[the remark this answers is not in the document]
timbru: I am open to another value but this seemed fine, it means that there may be a 500ms delay before things are being processed - but if this value would be too low then we get high CPU load and lock contention on the queue checking.

src/daemon/scheduler.rs
timbru: Previously 'ca refresh' dit two things:
- refresh this CA with its parent
- suspend children (if enabled)

Now these are separate tasks

src/daemon/scheduler.rs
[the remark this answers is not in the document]
ximon18: Ah ok :+1: So it's like a yield but I guess you already know you don't need to process the queue continuously when it's empty so this is harmless and helpful. I wondered briefly about using something like [`yield_now`](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html) instead but I guess you'd potentially still run too often when the system is idle.

src/daemon/scheduler.rs
timbru: Good point - I changed the logic to only force this jitter for cas that have too many parents. And use the normal 'jitter' value for the repo sync.

src/daemon/scheduler.rs
timbru: There is no need for this to be an even spread, it's just an easy way to ensure that the syncs are not all scheduled for the same moment.

This is even more important for the case where we have too many local CAs - because -likely- they will have local parents as well. E.g. 1000+ hosted CAs.

The semantics could be made smarter, but then it would need a lot more work to analyse all CAs and which parents they actually have, and how long ago the last sync (before restart) was.. and I don't think the complication is worth it.

Now - in case of a few CAs and parents per CA we just do a sync at start up, and otherwise we add some random delays. Using the 12 hour default and the mostly flat rather than gaussian random curve that I assume the random number generator will use this should spread things out more than enough.

Normally these syncs are not that time critical. And the user can still force rescheduling of the syncs using `krillc bulk refresh` if they have a reason to assume that there is a need.

src/daemon/scheduler.rs
timbru: Let me go through them one by one:

```Task::CheckSuspendChildren: NEW: Wasn't periodic, is now?```

It was periodic, but done as a side-effect. Now it has its own.

```Task::RefreshAnnouncementsInfo: CHANGED: Was cron every 5 seconds, but is now a recurring task on the queue every 10 minutes?```

This still uses another internal metric to avoid fetching the file every time. It will only do a refetch after 1 hour (and use if-modified-since). So checking every 5 seconds was excessive. I think every 10 minutes is fine given that these files are updated every 8 hours or so.

```Task::RenewObjectsIfNeeded: UNCHANGED```

ack

```Task::RepublishIfNeeded: CHANGED: Was cron every 10 minutes, but is now a recurring task on the queue every 5 minutes?```

It could also be 10 minutes, but every 5 minutes is fine as well. It's a cheap no-op in most case. That said, what I want to do in future is that instead of this a new actual mft/crl re-issuance is planned rather than checking every 5 minutes if work is needed.

```
QueueTask::RescheduleSyncParent: WAS UNUSED SO REMOVED?
QueueTask::RescheduleSyncRepo: WAS UNUSED SO REMOVED?
```

This was used to re-schedule those tasks with a 'due' time in case of failure. Now all tasks have a priority so this is not needed anymore. We just plan a normal sync parent/repo with the appropriate priority: try again in 15 minutes I think. 

```
Task::ResourceClassRemoved: UNCHANGED
Task::ServerStarted: UNCHANGED
Task::SweepLoginCache: UNCHANGED
Task::SyncParent: UNCHANGED
Task::SyncRepo: UNCHANGED
Task::UnexpectedKey: UNCHANGED
```

What may be interesting here is that `ServerStarted` is added when the server starts, and then it triggers that other tasks are planned. But because this is done later - this is asynchronous and does not delay the boot process. We could also just plan all tasks immediately without this. I don't really want to change this now, but I am not 100% sure what is best.

File: src/daemon/mq.rs (hunk 3 above)
Comment (timbru) [reply src/daemon/mq.rs:146]: Yes, but it's an unlikely corner case that does not matter in this context so I opted for brevity.

File: src/daemon/scheduler.rs (hunk 17 above)
Comment (timbru) [reply src/daemon/scheduler.rs:217]: no, in this case we only reschedule if something failed. If another change happens, e.g. MFT/CRL is re-issued or ROAs are changed - then that event will trigger that a new repo sync is scheduled.

For the parent exchange we want to reschedule soon in case of a failure, and later (24-36 hours by default) in case the exchange was successful. Parent syncs are not triggered by other internal events (wel except for repo configuration changes) - so they need to be planned here (and at startup).

File: src/daemon/scheduler.rs [hunk 23]
```diff
@@ -1,145 +1,101 @@
...
-                        try_sync_parent(&event_queue, &ca_manager, ca, parent, &actor).await
-                    }
-                    QueueTask::RescheduleSyncParent { ca, parent, due } => {
-                        if Time::now() > due {
-                            try_sync_parent(&event_queue, &ca_manager, ca, parent, &actor).await
-                        } else {
-                            event_queue.reschedule_sync_parent(ca, parent, due);
-                        }
-                    }
+                    Task::CheckSuspendChildren { ca } => self.suspend_children(ca).await,
```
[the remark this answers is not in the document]
timbru: Good point - renamed tasks for consistency

defaults/krill.conf

src/daemon/scheduler.rs
```diff
@@ -1,145 +1,101 @@
...
                         revocation_requests,
                     } => {
                         info!(
                             "Trigger send revoke requests for removed RC for '{}' under '{}'",
                             ca, parent
                         );
 
-                        if ca_manager
-                            .send_revoke_requests(&ca, &parent, revocation_requests)
+                        let mut requests = HashMap::new();
```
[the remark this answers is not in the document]
timbru: Now using:
```
let requests = HashMap::from([(rcn, revocation_requests)]);
```

This is more concise and the std lib can figure out what is most efficient.

src/daemon/mq.rs
```diff
@@ -223,3 +270,63 @@ impl eventsourcing::PostSaveEventListener<CertAuth> for MessageQueue {
         }
     }
 }
+
+//------------ Priority ------------------------------------------------------
+
+/// Can be used as a priority value for [`PriorityQueue`]. Meaning that the
+/// time value which is soonest has the highest priority. So, in short reverse
```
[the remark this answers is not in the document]
timbru: ack

src/daemon/scheduler.rs
timbru: ack

src/daemon/scheduler.rs
timbru: ack

src/daemon/scheduler.rs
timbru: no, it changes the priority to 'now' for the existing entry if you use 'bulk' - clarified.

timbru:
> I still need to finish reviewing the `scheduler.rs` changes.
> 
> Has anything changed in the Lagosta CSS or JS files? I can't tell from the diff but I suspect they've been recompiled without changes to the underlying Lagosta input files. Is that correct?

The 'next update' time was removed from the parents tab. No other changes. But.. I will do a review of the UI libraries and update dependencies (best effort).

ximon18:
> > Has anything changed in the Lagosta CSS or JS files? I can't tell from the diff but I suspect they've been recompiled without changes to the underlying Lagosta input files. Is that correct?
> 
> The 'next update' time was removed from the parents tab. No other changes. But.. I will do a review of the UI libraries and update dependencies (best effort).

For checking what has changed in the UI it would be useful to quote when making a PR here the relevant commit that was used to compile the changed `lagosta/*` files in the PR, or did you mention that and I missed it?assume

timbru:
> > > Has anything changed in the Lagosta CSS or JS files? I can't tell from the diff but I suspect they've been recompiled without changes to the underlying Lagosta input files. Is that correct?
> > 
> > 
> > The 'next update' time was removed from the parents tab. No other changes. But.. I will do a review of the UI libraries and update dependencies (best effort).
> 
> For checking what has changed in the UI it would be useful to quote when making a PR here the relevant commit in the Lagosta repo that was used to compile the changed `lagosta/*` files in the PR, or did you mention that and I missed it?assume

Indeed,  should have been: https://github.com/NLnetLabs/lagosta/commit/8153cfdf92ccadff73d4b3ebb2a0ae924e37abd0

And.. I just pushed new lagosta files for updated dependencies but this time referencing the commit:
https://github.com/NLnetLabs/lagosta/commit/8986193f33fa5cb145160c1ced1c167678845723

doc/development/06_ca_manager.md
```diff
@@ -118,23 +118,29 @@ To illustrate how the event listening is used here: If a `CaEvtDet::RoasUpdated`
...
+by `KrillServer`. The `Scheduler` runs as a separate async function and is started
+and watches the `TaskQueue` for queued tasks.
+
+The `TaskQueue` use a `PriorityQueue` ordering `Task`-s by `Priority`. The `Priority`
+is based on timestamps - where the lowest timestamp value results in the highest
+priority. The `Task` enum is explained below and contains the various background
+tasks that may be scheduled. They are scheduled or re-scheduled with an appropriate
+priority. Generally speaking:
+- we use 'now' for triggered tasks - such as sync with repo
+- we use 5 minutes for rescheduling failed syncs with parent/repo
```
ximon18, line 138: 5 minutes from when?

ximon18 (approved):

ximon18 (approved):

ximon18 (approved):

ximon18 (approved):

PR #812, opened 2022-04-05
V0.9.5 rc9 prep

timbru:


README.md
```diff
@@ -21,7 +21,15 @@ For more information please refer to the [documentation](https://krill.docs.nlne
 
 # Changelog
 
-## 0.9.5 RC8 'Have You considered these Upgrades?'
+## 0.9.5 RC9 'Have You considered these Upgrades?'
+
+NOTE: We found an issue with a test under Rust 1.59. We believe this issue may be
```
ximon18, line 26: This could be misunderstood to refer to a test of Rust itself. Perhaps instead say "when running one of the Krill tests under Rust 1.59" ?

Also "it was present in Krill 0.9.1" - does that refer to the issue in Tokio or the test in Krill? The latter I assume.

Other minor nits: "Rust" not "rust", and "Tokio" not "tokio".

timbru: yeah, you're right I will think about rephrasing this..

The main point is that that it does not seem to be an issue affecting production tested code - although that code was compiled with an earlier Rust version.

src/commons/eventsourcing/store.rs
```diff
@@ -872,7 +881,7 @@ where
         }
 
         if aggregate_opt.is_none() {
-            debug!("No snapshots found for '{}' will try from initialization event.", id);
+            warn!("No suitable snapshot for '{}' will rebuild state from events.", id);
```
ximon18, line 884: What can or should an operator do with these warnings? Is warn the right level or should these new warnings be info level instead?

timbru: I thought a warning might be appropriate because this can take some time. I could add "This can take some time".

src/daemon/ca/status.rs
```diff
@@ -65,49 +68,130 @@ impl Default for CaStatus {
...
+
+        Ok(())
     }
 
-    /// Returns the stored CaStatus for a CA, or a default (empty) status if it can't be found
-    pub async fn get_ca_status(&self, ca: &Handle) -> Arc<CaStatus> {
-        // Try to get it from disk, if it's missing or corrupt - get a default
-        let lock = self.locks.ca(ca).await;
-        let _ = lock.read().await;
+    /// Load current status from disk, to be used when starting up. Iif there are any
```
ximon18, line 98: Iif? What is this? Mistyped if? Or mistyped IFF?

src/daemon/ca/status.rs
```diff
@@ -65,49 +68,130 @@ impl Default for CaStatus {
...
+    /// issues parsing data then default values are used - this data is not critical
+    /// so any missing, corrupted, or no longer supported data format - can be ignored.
+    /// It will get updated with new status values as Krill is running.
+    fn load_full_status(&self, ca: &Handle) -> KrillResult<()> {
+        let repo: RepoStatus = self.store.get(&Self::repo_status_key(ca))?.unwrap_or_default();
+
+        // parents
+        let mut parents = ParentStatuses::default();
+        for parent_key in self.store.keys(Some(ca.to_string()), "parents-")? {
```
ximon18, line 107: Not urgent but these repeated string constants are a potential source of error if they become out of sync with each other and should instead perhaps be actual constants.

src/daemon/ca/status.rs
```diff
@@ -65,49 +68,130 @@ impl Default for CaStatus {
...
+        // children
+        let mut children = HashMap::new();
+        for child_key in self.store.keys(Some(ca.to_string()), "children-")? {
+            let child = child_key.name().strip_prefix("children-").unwrap();
+            if let Some(child) = child.strip_suffix(".json") {
+                if let Ok(child) = ChildHandle::from_str(child) {
+                    let status: ChildStatus = self.store.get(&Self::child_status_key(ca, &child))?.unwrap_or_default();
 
-        Arc::new(status)
+                    children.insert(child, status);
```
ximon18, line 129: The mismatch of insert here but add above is a bit strange given that the `fn add()` call above actually calls `insert()` ... maybe rename `fn add()` to `fn insert()` ?

src/daemon/ca/status.rs
```diff
@@ -65,49 +68,130 @@ impl Default for CaStatus {
...
+        Ok(())
     }
 
-    /// Returns the stored CaStatus for a CA, or a default (empty) status if it can't be found
-    pub async fn get_ca_status(&self, ca: &Handle) -> Arc<CaStatus> {
-        // Try to get it from disk, if it's missing or corrupt - get a default
-        let lock = self.locks.ca(ca).await;
-        let _ = lock.read().await;
+    /// Load current status from disk, to be used when starting up. Iif there are any
+    /// issues parsing data then default values are used - this data is not critical
```
ximon18, line 99: Is this statement about this functionality not being critical compatible with the use of `unwrap()` calls in the fn below? Could it panic and cause Krill to exit while Krill could actually have happily continued running?

src/daemon/ca/status.rs
```diff
@@ -116,46 +200,43 @@ impl StatusStore {
     ) -> KrillResult<()> {
         let error_response = Self::error_to_error_res(error);
 
-        self.update_ca_status(ca, |status| status.parents.set_failure(parent, uri, error_response))
-            .await
+        // self.update_ca_status(ca, |status| status.parents.set_failure(parent, uri, error_response))
```
ximon18, line 203: Can this commented out code be removed?

src/daemon/ca/status.rs
```diff
@@ -164,107 +245,225 @@ impl StatusStore {
...
-    pub async fn remove_child(&self, ca: &Handle, child: &ChildHandle) -> KrillResult<()> {
-        self.update_ca_status(ca, |status| {
-            status.children.remove(child);
-        })
-        .await
+    pub fn remove_child(&self, ca: &Handle, child: &ChildHandle) -> KrillResult<()> {
+        let mut cache = self.cache.write().unwrap();
+
+        if let Some(mut ca_status) = cache.get_mut(ca) {
+            let ca_status = Arc::make_mut(&mut ca_status);
```
ximon18, line 273: The documentation for `make_mut()` says "If there are other Arc pointers to the same allocation, then make_mut will clone the inner value to a new allocation to ensure unique ownership. This is also referred to as clone-on-write.". You would then be calling `.children.remove(child)` on a clone of the `children` `HashMap`, i.e. not actually removing anything from the _actual_ children HashMap, am I understanding that right and is that really what is wanted?

src/daemon/ca/status.rs
```diff
@@ -164,107 +245,225 @@ impl StatusStore {
...
+        let mut cache = self.cache.write().unwrap();
 
-        let mut status: CaStatus = self.store.get(&Self::status_key(ca)).ok().flatten().unwrap_or_default();
+        if !cache.contains_key(ca) {
+            cache.insert(ca.clone(), Arc::new(CaStatus::default()));
+        }
 
-        op(&mut status);
+        let mut ca_status = cache.get_mut(ca).unwrap(); // safe, we just set it if missing
+        let ca_status = Arc::make_mut(&mut ca_status);
```
ximon18, line 312: Same concern about `make_mut()` as above.

src/daemon/ca/status.rs
```diff
@@ -164,107 +245,225 @@ impl StatusStore {
...
+
+            let mut ca_status = cache.get_mut(ca).unwrap(); // safe, we just set it if missing
+            let ca_status = Arc::make_mut(&mut ca_status);
+
+            let mut parent_status = ca_status.parents.get_mut_status(parent);
+            op(&mut parent_status);
+            parent_status.clone()
+        };
+
+        // let status = ca_status.parents.get(parent).unwrap();
```
ximon18, line 368: Can this commented line and the commented code below be removed?

src/daemon/ca/status.rs
```diff
@@ -164,107 +245,225 @@ impl StatusStore {
...
+
+            let status_testbed_before_migration: CaStatus =
+                serde_json::from_str(status_testbed_before_migration).unwrap();
+
+            let store = StatusStore::new(&d, "status").unwrap();
+            let testbed = Handle::from_str("testbed").unwrap();
+
+            let status_testbed_migrated = store.get_ca_status(&testbed);
+
+            assert_eq!(&status_testbed_before_migration, status_testbed_migrated.as_ref());
```
ximon18, line 466: What is this actually testing? If the migration did nothing this would still succeed wouldn't it?

ximon18 (approved):

src/daemon/ca/status.rs
timbru: if, fixed

src/daemon/ca/status.rs
timbru: can't panic, but pushed a fix to avoid calling `unwrap()` - should be better and less confusing

src/daemon/ca/status.rs
timbru: No it wouldn't. This tests that the CaStatus is the same when parsed before doing any migration, and after it has been migrated. The 'hidden' knowledge in the test is that the original status.json is parsed, split into components, and that these small components are then parsed.

src/daemon/ca/status.rs
timbru: extracted

src/daemon/ca/status.rs
timbru: fair point, done

src/daemon/ca/status.rs
timbru: yes

src/daemon/ca/status.rs
timbru: doh! good point.. I had it reversed somehow..

That said, this triggered me to rethink it all. Read operations here are infrequent - where we need a copy of the data that is -while write operations are frequent. So not using an Arc at all actually simplifies the code quite a bit and makes the updates more efficient.

src/commons/eventsourcing/store.rs
timbru: added

PR #815, opened 2022-04-07
Benchmark test

timbru:
Feel free to have an early look, but this still needs some love as there is lot of code duplication between the testbed and benchmark setup. Extracting some functions is needed for sure.

Note: we need to bypass the normal task queue a bit here. Doing this as a post-start thing is possible but would require orchestration regarding.. only add children (benchmark) when the testbed is fully synced, and only add ROAs after the benchmark CAs are synced. So.. complicated.

And perhaps more importantly.. I think it's better that we does before really starting krill because then we can error out in case of issues during this setup.


timbru:
Merged dev into this, and changed target to dev (0.10.x instead of 0.9.x).

timbru:
@ximon18 could you have a look at this now that it's rebased (merged) to target the 0.10.0 dev branch?

It seems that the e2e test fails because of some strange state:
```
module.docker_deploy.null_resource.setup_docker (local-exec): error[E0761]: file for module `signing` found at both "src/commons/crypto/signing.rs" and "src/commons/crypto/signing/mod.rs"
```

There is no `src/commons/crypto/signing.rr`

PR #817, opened 2022-04-12
Do not resync repo on startup if there are too many CAs.

timbru:


timbru:
Verified that the restart works as expected.

src/daemon/scheduler.rs
```diff
@@ -146,20 +149,21 @@ impl Scheduler {
 
         debug!("Adding tasks at start up");
 
-        let use_jitter = cas.len() >= SCHEDULER_USE_JITTER_CAS_THRESHOLD;
+        let too_many_cas = cas.len() >= SCHEDULER_USE_JITTER_CAS_THRESHOLD;
```
ximon18, line 152: I find these "too_many" variable names confusing and the first one is about jitter and doesn't mention that. Maybe rename these flags to `use_jitter` and `do_repo_resync` or something?

ximon18: Oh it was called `use_jitter` before...

src/daemon/scheduler.rs
```diff
@@ -172,18 +176,19 @@ impl Scheduler {
...
-            self.tasks
-                .sync_repo(ca.handle().clone(), self.config.ca_refresh_start_up(use_jitter));
+            // Plan a sync with the repo. But only in case we only have a handful
+            // of CAs.
+            //
+            // Note: if circumstances dictate a sync e.g. because ROAs are changed,
+            // then it will be scheduled accordingly. Furthermore, users can the 'bulk'
+            // function to explicitly force scheduling a sync.
+            if !too_many_cas_resync_repo {
+                self.tasks.sync_repo(ca.handle().clone(), now());
```
ximon18, line 190: Do I understand correctly that jitter is no longer used here because it's only now done if there are very few CAs and jitter is only needed when there are lots of CAs?

ximon18 (approved):

src/daemon/scheduler.rs
timbru: indeed

src/daemon/scheduler.rs
timbru: improved, I think..

Issue #833, opened 2022-06-10
Schedule "RenewObjectsIfNeeded" at startup.


timbru:
The background jobs were re-architected in release 0.9.5 (13 april 2022). There is a regression that the task to renew signed object (ROAs in particalur) was not scheduled on server (re-)start.

We need to add:
```
self.tasks.renew_if_needed(now());
```

Below this line:
https://github.com/NLnetLabs/krill/blob/7babd7432aaf80737dff34d36e28a8ac918473f3/src/daemon/scheduler.rs#L210

And make a quick 0.9.6 release for this, and port this to our dev branch as well.

Thanks to @ydahhrk for reporting this issue.

PR #834, opened 2022-06-10
Release 0.9.6 'Newer ROAs Please'

timbru:
Fixes issue #833

### DIFF
--- a/src/daemon/scheduler.rs
+++ b/src/daemon/scheduler.rs
@@ -208,6 +208,7 @@ impl Scheduler {
         }
 
         self.tasks.republish_if_needed(now());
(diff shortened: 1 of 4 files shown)


src/daemon/scheduler.rs
```diff
@@ -208,6 +208,7 @@ impl Scheduler {
         }
 
         self.tasks.republish_if_needed(now());
+        self.tasks.renew_if_needed(now());
```
timbru, line 211: Note that when this task is done, it will re-schedule itself - just like the republish_if_needed does.

ximon18 (approved):

PR #839, opened 2022-06-20
Remove bcder

timbru:
Should wait with merging until after https://github.com/NLnetLabs/rpki-rs/pull/211 is merged and krill is updated to use the new rpki-rs error types.


timbru:
Changed the base to the `migrate_0_10_0` branch to make it more clear what would be applied. We will change the target back to `dev` after that has been merged.

PR #846, opened 2022-07-12
Use more descriptive errors in rpki-rs (#845)

timbru:
Targeting unmerged `remove-bcder` for now, needs to target `dev` once PR #839  is merged.


timbru:
retarget to `better-errors` because... that was actually done first and this code was a branch off that branch.

ximon18 (approved):
Seems fine. I haven't checked the history of the better errors branch that is now being depended on, but, assuming that's good, there's very little changed in this PR as far as I can see, mainly some handled errors instead of unwraps and some propagated error messages where there were none before.

PR #860, opened 2022-08-02
Upgrade KMIP dependencies.

ximon18:
No new functionality, only security fixes in upstream dependencies of the crate which were detected by GitHub Dependabot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,16 +21,17 @@ For more information please refer to the [documentation](https://krill.docs.nlne
 
 # Changelog
 
-## 0.10.0 RC2 'Hush'
+## 0.10.0 RC3 'Hush'
 
-This second release candidate fixes the following issues identified
-in RC1:
+This third release candidate fixes the following issues identified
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -21,16 +21,17 @@ For more information please refer to the [documentation](https://krill.docs.nlne
 
 # Changelog
 
-## 0.10.0 RC2 'Hush'
+## 0.10.0 RC3 'Hush'
```
timbru, line 24: Perhaps it's better to do this as a part of a separate release-0.10.0-rc3 issue. It's not critical of course, but the krill version is still 0.10.0-rc2 - so technically this may be premature

timbru (approved):

README.md
ximon18: The idea was to set the Cargo.toml version to rc3 in the rc3 branch that this PR merges into.

timbru: works for me!

PR #867, opened 2022-08-08
Use different GH Actions cache keys for cargo-deb binaries for non-lzma and lzma build variants.

ximon18:
As the Xenial build was slower due to not using LZMA it would always finish last and its version of cargo deb would be cached and then used by all future DEB packaging operations until the cache entry expired and the process repeated. So non-Xenial packages would also have been built without LZMA compression which wasn't the intention. This PR ensures a unique cache key is used for the two different cases.


ximon18:
The fix provided by this PR was already merged to main as part of PR #773.

Issue #868, opened 2022-08-10
CLI should show XML


timbru:
Since implementing #765 the CLI shows a simple text for the RFC8183 files instead of the XML as it should.

PR #869, opened 2022-08-10
Show RFC8183 XML in CLI (#868)

timbru:


src/daemon/http/server.rs
```diff
@@ -1597,8 +1588,7 @@ async fn api_ca_children(req: Request, path: &mut RequestPath, ca: CaHandle) ->
                 Method::DELETE => api_ca_child_remove(req, ca, child).await,
                 _ => render_unknown_method(),
             },
-            Some("contact") => api_ca_parent_contact(req, ca, child).await,
-            Some("parent_response.json") => api_ca_parent_res_json(req, ca, child).await,
+            Some("contact") | Some("parent_response.json") => api_ca_parent_res_json(req, ca, child).await,
```
ximon18, line 1591: This is a change compared to the released Krill 0.9.6 code. Do these not require different responses? Even if correct, is this now then a breaking change to the API? Should one of these be marked obsolete for later removal?

ximon18 (approved):
LGTM. As there are no tests that verify that XML is actually produced I'll test manually against the testbed once you apply this fix to it.

src/daemon/http/server.rs
timbru: The old parent 'contact' vs 'response' is the same as this, except for that it also wraps a possible TA parent (i.e. self). This is not used outside of testing.

The public documentation talks about the RFC8183 XML response for this api point, so technically this was incorrectly documented in 0.9.6 but is now correct once again.

Another place where there is a new difference is here:
https://krill.docs.nlnetlabs.nl/en/stable/cli.html#krillc-parents-contact

This is the parent contact as seen from the child towards a parent instead of vice versa as is done here. I will update the documentation. The 0.10.0 ParentCaContact enum no longer contains the complete response, just the bits that are interesting (service_uri and public key in particular).

PR #883, opened 2022-08-22
Run Krill in a Docker container as user `krill` (fixes #882)

### LINKED ISSUE #882: Run as user 'krill' inside Docker containers, not 'root'
While comparing the current state of the Routinator and Krill build processes I noticed that while the Krill `Dockerfile` defines a user and group for Krill and pre-creates a data dir for Krill and assigns ownership to the `krill` user, the `Dockerfile` never uses the `USER` statement to "become" the `krill` user, thus at runtime the container runs `krill` as user `root`. This was not the intention.

ximon18:


timbru (approved):

ximon18:
Before:
```
$ docker run -it -d --rm nlnetlabs/krill:latest
$ docker exec krill ps ax
PID   USER     TIME  COMMAND
    1 root      0:00 /sbin/tini -- /opt/entrypoint.sh krill -c /var/krill/data/krill.conf
    7 root      0:00 krill -c /var/krill/data/krill.conf
   21 root      0:00 ps ax
```

After: (where the `test-amd64` is that produced by [this GH Actions run](https://github.com/NLnetLabs/krill/actions/runs/2903445912) on this PR branch)
```
$ docker run -it -d --rm nlnetlabs/krill:test-amd64
$ docker exec krill ps ax
PID   USER     TIME  COMMAND
    1 krill     0:00 /sbin/tini -- /opt/entrypoint.sh krill -c /var/krill/data/krill.conf
    7 krill     0:01 krill -c /var/krill/data/krill.conf
   21 krill     0:00 ps ax
```

ximon18:
Note that Krill checks that it has write permission on startup and will fail with an error if not, so anyone who upgrades without fixing ownership on Krill data files will see such an error.

timbru (approved):

PR #884, opened 2022-08-22
Prepare v0.10.0 rc3

timbru:


src/upgrades/pre_0_10_0/cas_migration.rs
```diff
@@ -86,10 +85,7 @@ impl CasMigration {
 
 impl UpgradeStore for CasMigration {
     fn needs_migrate(&self) -> Result<bool, PrepareUpgradeError> {
-        Ok(self.current_kv_store.version_is_after(KrillVersion::release(0, 9, 0))?
-            && self
-                .current_kv_store
-                .version_is_before(KrillVersion::candidate(0, 10, 0, 1))?)
+        unimplemented!("This is checked in upgrades/mod.rs")
```
ximon18, line 88: Should this strictly speaking be `unreachable!()` rather than `unimplemented!()` ?

ximon18 (approved):

src/upgrades/pre_0_10_0/cas_migration.rs
timbru: yes

timbru: this was fixed btw :D

PR #891, opened 2022-08-22
Fix unused ta compilation warning

ximon18:


timbru (commented):
I think I fixed this in PR #896. Not strictly related, but I don't usually bother with separating commits for unused import cleanups.

PR #892, opened 2022-08-22
Fix unused field compilation warnings

ximon18:


timbru (approved):

timbru:
It's in draft but still I would be fine with this change :D

PR #893, opened 2022-08-22
Prevent unused variable warnings when default features are disabled.

ximon18:
Such as [this](https://github.com/NLnetLabs/krill/runs/7955242431?check_suite_focus=true#step:4:459) and [this](https://github.com/NLnetLabs/krill/runs/7955242431?check_suite_focus=true#step:4:466).


timbru (approved):
Looks good for now. We can revisit if/when we revise the use of features.

PR #896, opened 2022-08-23
API to upload repository response should not change #895

timbru:


src/commons/error.rs
```diff
@@ -755,7 +755,7 @@ impl Error {
 
             Error::CaRepoIssue(ca, err) => ErrorResponse::new("ca-repo-issue", &self).with_ca(ca).with_cause(err),
 
-            Error::CaRepoResponseInvalidXml(ca, err) => ErrorResponse::new("ca-repo-response-invalid-xml", &self)
+            Error::CaRepoResponseInvalid(ca, err) => ErrorResponse::new("ca-repo-response-invalid-xml", &self)
```
ximon18, line 758: A minor point perhaps, but I note that you removed the XML suffix from the enum member name but not from the language key. This seems inconsistent and a bit confusing.

test-resources/errors/ca-repo-response-invalid-xml.json
```diff
@@ -1 +1 @@
-{"label":"ca-repo-response-invalid-xml","msg":"CA 'ca' got invalid repository response xml: expected some tag","args":{"ca":"ca","cause":"expected some tag"}}
\ No newline at end of file
+{"label":"ca-repo-response-invalid-xml","msg":"CA 'ca' got invalid repository response: expected some tag","args":{"ca":"ca","cause":"expected some tag"}}
```
ximon18, line 1: Same comment as above, i.e. you're removing XML from the string content but not from the translation key name.

ximon18 (approved):

src/commons/error.rs
timbru: yeah, this is because of the (far?) off chance that someone might parse the error code when using the api so I thought it better not to change it.

PR #905, opened 2022-08-25
Include (optional) comment with configured (ROA) authorisations. #863

timbru:
Ready for review, but only to be (squash) merged after 0.10.0 is released.


timbru:
Note: will document API changes (non-breaking) on a separate branch for the krill documentation.

src/daemon/ca/events.rs
```diff
@@ -970,6 +974,13 @@ impl fmt::Display for CaEvtDet {
 
             // Route Authorizations
             CaEvtDet::RouteAuthorizationAdded { auth } => write!(f, "added ROA: '{}'", auth),
+            CaEvtDet::RouteAuthorizationComment { auth, comment } => {
+                if let Some(comment) = comment {
+                    write!(f, "added comment to ROA: '{}' => {}", auth, comment)
+                } else {
+                    write!(f, "removed comment from ROA: '{}'", auth)
```
ximon18, line 981: Is this a removal if the ROA authorization had no comment to start with? Or is it assumed that we can't get to this point with a None comment if the ROA authorization didn't already have a comment?

test-resources/bgp/expected_full_report.txt
```diff
@@ -26,13 +26,13 @@ Authorizations which may be too permissive:
...
 		10.0.0.0/24 => 64496
 
-AS0 Authorizations disallowing announcements for prefixes
+AS0 ROA configurations disallowing announcements for prefixes
 
-	Definition: 10.0.4.0/24 => 0
+	Configuration: 10.0.4.0/24 => 0
 
-AS0 Authorization which are made redundant by authorizations for the prefix from real ASNs
+AS0 Authorization which are made redundant by configuration(s) for the prefix from real ASNs
```
ximon18, line 33: Two lines up `AS0 Authorizations` was changed to `AS0 ROA configurations`, why then on this line is `AS0 Authorization` left unchanged?

test-resources/bgp/expected_full_report.txt
```diff
@@ -48,17 +48,17 @@ Announcements from an authorized ASN, which are too specific (not allowed by max
 
 	Announcement: 10.0.0.0/24 => 64496
 
-		Disallowed by authorization(s):
+		Disallowed by configuration(s):
```
ximon18, line 51: Should this perhaps say "ROA" configuration as elsewhere you are explicit that this is not some other sort of configuration?

src/commons/api/roas.rs
```diff
@@ -129,29 +129,35 @@ impl RoaAggregateKeyFmtError {
...
 
-//------------ RoaDefinition -----------------------------------------------
+//------------ RoaPayload --------------------------------------------------
 
-/// This type defines the definition of a Route Origin Authorization (ROA), i.e.
-/// the originating asn, IPv4 or IPv6 prefix, and optionally a max length.
+/// This type defines the definition of a Route Origin Authorization (ROA)
+/// payload: ASN, Prefix and optional Max Length
+///
+/// Note that a ROA object may contain multiple payloads (RoaDefinitions)
```
ximon18, line 137: This is confusing as the comment has been updated but refers to the no-longer-existing RoaDefinition type.

src/commons/api/roas.rs
```diff
@@ -280,22 +286,186 @@ impl Ord for RoaDefinition {
...
-impl AsRef<TypedPrefix> for RoaDefinition {
+impl AsRef<TypedPrefix> for RoaPayload {
     fn as_ref(&self) -> &TypedPrefix {
         &self.prefix
     }
 }
 
+//------------ RoaConfiguration --------------------------------------------
+
+/// This type defines a saved RoaConfiguration. It includes the actual ROA
```
ximon18, line 303: > This type defines a saved RoaConfiguration
Perhaps good to avoid a self-referential description of the type, it's a bit strange for the reader. "An apple is an apple" ...

src/commons/api/roas.rs
```diff
@@ -129,29 +129,35 @@ impl RoaAggregateKeyFmtError {
...
+//------------ RoaPayload --------------------------------------------------
 
-/// This type defines the definition of a Route Origin Authorization (ROA), i.e.
-/// the originating asn, IPv4 or IPv6 prefix, and optionally a max length.
+/// This type defines the definition of a Route Origin Authorization (ROA)
+/// payload: ASN, Prefix and optional Max Length
+///
+/// Note that a ROA object may contain multiple payloads (RoaDefinitions)
+/// aggregated by (the same) ASN. This is discouraged and Krill prefers
+/// to generate a dedicated ROA object for each payload, but aggregation
```
ximon18, line 139: What is this "ROA object" that is referred to in multiple places? Is there a suitable RFC link we can make to this or is it a thing in Krill or...?

src/commons/api/roas.rs
```diff
@@ -280,22 +286,186 @@ impl Ord for RoaDefinition {
...
     fn as_ref(&self) -> &TypedPrefix {
         &self.prefix
     }
 }
 
+//------------ RoaConfiguration --------------------------------------------
+
+/// This type defines a saved RoaConfiguration. It includes the actual ROA
+/// payload that needs be authorized in a ROA object as well as other
+/// information that is only visible to users - i.e. an optional comment
```
ximon18, line 305: Users of what? I assume that users of a relying party won't see a ROA comment as a ROA comment is not someting in the RFCs, right? Maybe it's worth saying this is a custom Krill extension to what is covered by RFCs? (assuming I'm right of course!)

src/commons/api/roas.rs
```diff
@@ -280,22 +286,186 @@ impl Ord for RoaDefinition {
...
+/// field, and in future perhaps other things such as tags used for
+/// classification/monitoring/bpp analysis.
+///
+/// Note that this type is backward compatible with the RoaDefinition type
+/// used until Krill 0.10.0.
```
ximon18, line 310: Is "is backward compatible" strong enough or is the case that this MUST remain backward compatible?

src/commons/api/roas.rs
```diff
@@ -129,29 +129,35 @@ impl RoaAggregateKeyFmtError {
...
             asn,
             prefix,
             max_length,
         }
     }
 
-    pub fn explicit_max_length(self) -> Self {
-        RoaDefinition {
+    /// Ensures that the payload uses an explicit max length
+    pub fn into_explicit_max_length(self) -> Self {
```
ximon18, line 159: On `RoaConfigurationUpdates` this is called `into_explicit`. Should these match? Perhaps the longer more descriptive name should be used in both places?

src/commons/api/roas.rs
```diff
@@ -280,22 +286,186 @@ impl Ord for RoaDefinition {
...
+    #[serde(default)] // missing is same as no comment
+    comment: Option<String>,
+}
+
+impl RoaConfiguration {
+    pub fn new(payload: RoaPayload, comment: Option<String>) -> Self {
+        RoaConfiguration { payload, comment }
+    }
+
+    pub fn unpack(self) -> (RoaPayload, Option<String>) {
```
ximon18, line 324: It's a bit late to point this out now as the `unpack` pattern is used in many places in Krill I believe, but isn't this expressable using the `From` trait?

src/commons/api/roas.rs
```diff
@@ -280,22 +286,186 @@ impl Ord for RoaDefinition {
...
+pub struct RoaConfiguration {
+    #[serde(flatten)]
+    payload: RoaPayload,
+    #[serde(default)] // missing is same as no comment
+    comment: Option<String>,
+}
+
+impl RoaConfiguration {
+    pub fn new(payload: RoaPayload, comment: Option<String>) -> Self {
+        RoaConfiguration { payload, comment }
```
ximon18, line 321: Is it handled somewhere that we ensure that comments cannot contain an (unescaped?) bare `#` character, as that character is used to prefix the comment in the `Display` impl?

src/commons/api/roas.rs
```diff
@@ -280,22 +286,186 @@ impl Ord for RoaDefinition {
...
+
+impl From<RoaPayload> for RoaConfiguration {
+    fn from(payload: RoaPayload) -> Self {
+        RoaConfiguration { payload, comment: None }
+    }
+}
+
+//------------ ConfiguredRoa -----------------------------------------------
+
+/// Defines an existing ROA configuration and related information.
```
ximon18, line 391: Above `RoaConfiguration` is documented as "defines a saved RoaConfiguration", how is that different to "Defines an existing ROA configuration"? Perhaps use other wording to make the difference between the purpose of these types clearer.

src/commons/api/roas.rs
```diff
@@ -280,22 +286,186 @@ impl Ord for RoaDefinition {
...
+
+    pub fn effective_max_length(&self) -> u8 {
+        self.roa_configuration.payload.effective_max_length()
+    }
+
+    pub fn nr_of_specific_prefixes(&self) -> u128 {
+        self.roa_configuration.payload.nr_of_specific_prefixes()
+    }
+
+    pub fn as_roa_ip_address(&self) -> RoaIpAddress {
```
ximon18, line 424: Couldn't this be done with a `From`/`Into` impl?

src/commons/api/roas.rs
```diff
@@ -304,59 +474,76 @@ impl fmt::Display for RoaDefinitions {
...
     }
 
-    pub fn new(added: Vec<RoaDefinition>, removed: Vec<RoaDefinition>) -> Self {
-        RoaDefinitionUpdates { added, removed }
+    pub fn new(added: Vec<RoaConfiguration>, removed: Vec<RoaPayload>) -> Self {
+        RoaConfigurationUpdates { added, removed }
+    }
+
+    /// Ensures that an explicit (canonical) max length is used.
+    pub fn into_explicit(self) -> Self {
```
ximon18, line 500: s/into_explicit/into_explicit_max_length/ ?

src/commons/api/roas.rs
```diff
@@ -368,43 +555,35 @@ impl fmt::Display for RoaDefinitionUpdates {
...
 
-impl FromStr for RoaDefinitionUpdates {
+impl FromStr for RoaConfigurationUpdates {
     type Err = AuthorizationFmtError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let mut added = vec![];
         let mut removed = vec![];
 
         for line in s.lines() {
```
ximon18, line 565: So you're removing support for entire comment lines in a block of ROA strings? Is that deliberate? Was that ever officially even supported?

src/commons/bgp/analyser.rs
```diff
@@ -84,8 +84,8 @@ impl BgpAnalyser {
                 .collect(),
         };
 
-        let (roas, roas_not_held): (Vec<RoaDefinition>, _) = roas
-            .iter()
+        let (roas_held, roas_not_held): (Vec<ConfiguredRoa>, _) = roas
+            .into_iter()
             .partition(|roa| resources_held.contains_roa_address(&roa.as_roa_ip_address()));
```
ximon18, line 89: Why wasn't this call to `contains_roa_address()` needed before?

ximon18 (approved):

src/daemon/ca/events.rs
timbru: The update is only accepted if there is a change. So this could only be confusing if the comment was an empty string I suppose and then be removed. But it would still be accurate then, as written.

src/commons/api/roas.rs
timbru: I suppose we could use `impl From<TypeA> for (TypeB, TypeC, ..)` like things instead. But I would make a separate issue for this.. I am using `unpack` rather than `unwrap` as these functions were called in the beginning of krill to avoid at least some confusion :)

src/commons/api/roas.rs
timbru: The implementation currently uses the *first* # on the line to indicate the start of a comment section (which may be empty even btw). Subsequent # are accepted inside comments.

timbru: I don't think this is an issue for json as it's inside a string. I think serde will take care of " characters inside the string..

We could limit the allowed characters, but I am afraid that that will severely limit the usability for languages other than English.

src/commons/bgp/analyser.rs
timbru: that line is actually unchanged.

ximon18: Ignore me ... sigh.

test-resources/bgp/expected_full_report.txt
timbru: ack. fixed

test-resources/bgp/expected_full_report.txt
timbru: ack fixed

src/commons/api/roas.rs
timbru: fixed

src/commons/api/roas.rs
timbru: fixed

src/commons/api/roas.rs
timbru: fixed (I think)

src/commons/api/roas.rs
timbru: yes! But I explained it better now (I think).

Essentially - if we can avoid upgrades and breaking changes we probably should :D

src/commons/api/roas.rs
timbru: using the longer `into_explicit_max_length` now.

src/commons/api/roas.rs
timbru: fixed

src/commons/api/roas.rs
timbru: It could, but because of how references and `Deref` is used here that would result in less readable code. So I prefer to keep it as is.

src/commons/api/roas.rs
timbru: fixed - I did not use a link but I mention the RFC number at least so it can be found easily.

Issue #615, opened 2021-07-28
Replace trace level request/response logging with SSLKEYLOGFILE support?


ximon18:
Currently Krill has code to log detailed (but not complete) requests and responses at TRACE level. This can be particularly helpful when Krill is configured to communicate with an OpenID Connect provider. However, a better solution might be to add support for the [SSLKEYLOGFILE environment variable](https://wiki.wireshark.org/TLS#TLS_Decryption) via https://docs.rs/rustls/0.19.1/rustls/struct.KeyLogFile.html such that packets captured using standard tooling such as `tcpdump` and analyzed using tools such as Wireshark can be decrypted.

ximon18:
Example implementations that I made when writing the `kmip-protocol` crate can be seen here: https://github.com/NLnetLabs/kmip-protocol/search?q=SSLKEYLOGFILE

ximon18:
This was included in the Krill v0.10.0-rc3 release.

PR #910, opened 2022-08-29
Add SSLKEYLOGFILE support (#615)

ximon18:
Per https://github.com/NLnetLabs/krill/issues/615 this PR adds support for SSLKEYLOGFILE env var triggered logging of TLS session keys to disk for use in later decrypting HTTPS sessions for diagnostic purposes.

For more information see https://wiki.wireshark.org/TLS#tls-decryption

### DIFF
--- a/src/daemon/http/tls.rs
+++ b/src/daemon/http/tls.rs
@@ -179,6 +179,7 @@ impl TlsConfigBuilder {
         config.set_single_cert(cert, key).map_err(TlsConfigError::InvalidKey)?;
         config.set_protocols(&["h2".into(), "http/1.1".into()]);
 
+        # See: https://wiki.wireshark.org/TLS#tls-decryption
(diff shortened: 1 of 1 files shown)


ximon18:
Example usage:

1. Run Wireshark (as root if needed to capture from network interfaces) and capture on interface "Loopback: lo".

2. Open Edit -> Preferences -> Protocols -> TLS and set "(Pre)-Master-Secret log filename" to `/tmp/sslkeys.log`.

3. Set the Wireshark display filter to `http2.header.value matches "/api/.*"`.

4. Run Krill with key saving enabled:

```
$ krillc config simple --token blah --data /tmp/krill/ --logfile /tmp/krill.log > /tmp/krill.conf
$ SSLKEYLOGFILE=/tmp/sslkeys.log krill --config /tmp/krill.conf
```

5. Login to the Krill web UI and create a CA.

You should now see something like this in Wireshark:

image

image

timbru (approved):

Issue #808, opened 2022-03-22
Disallow negative numbers in config


timbru:
Negative numbers are now allowed in some places (i64 vs u64) - even if they don't make sense.

Theoretically this allows users kill their own Krill instances by using negative update intervals or some such. The impact should be limited, but it's best to disallow this at the time of parsing/verifying the config file.

PR #912, opened 2022-08-31
Disallow negative numbers in config file. #808

timbru:
[draft] intended to be merged after 0.10.0 is released


src/daemon/ca/manager.rs
```diff
@@ -1685,7 +1685,7 @@ impl CaManager {
             .ca_objects_store
             .ca_objects(ca_handle)?
             .closest_next_update()
-            .unwrap_or_else(|| Timestamp::now_plus_hours(self.config.republish_hours()));
+            .unwrap_or_else(|| self.config.issuance_timing.republish_worst_case().into());
```
ximon18, line 1688: Comparing the old code and new code here is a bit tricky, but it looks to me like the new code introduces the use of jitter, is that correct and if so is that intentional?

timbru: yeah, this was missed earlier. As the fn name now saus this is a worst case guess for the next time, in the remote event that the current of objects is empty.. it can't really happen but I thought it would be better than unwrapping.

ximon18 (approved):

Issue #913, opened 2022-08-31
Allow explicit disabling of https


timbru:
We get requests from time to time from users who want to run krill using plain http rather than HTTPS (with a self-signed certificate).

While we believe it would not be advisable to change the default behaviour for this, there is something to be said for using plain HTTP for testing, or in setups where krill is configured to listen to localhost only (as is the default in fact) and it's only accessible through an HTTPS capable proxy server.

So, in short: add a `disable_https = true|false (default false)` option to the configuration and use it when Krill starts.

timbru:
Implemented using the existing, but undocumented `https_mode` directive.

Documented in the krill.conf template now:

```
# Specify the HTTPS mode. Krill supports three modes:
#
# "generate" (DEFAULT)
#
# Krill will generate a key pair and create a self-signed certificate
# if no previous key pair or certificate is found. File names used
# are data_dir/ssl/key.pem and data_dir/ssl/cert.pem respectively.
#
# "existing"
#
# Krill expects an existing key pair and certificate in the same
# locations where it would otherwise store its generated key pair
# and self-signed certificate.
#
# "disable"
#
# Krill will use plain HTTP.
#
# This mode is not recommended as HTTPS adds little overhead, and
# even with a self-signed certificate provides better security
# out of the box.
#
# That said, if you are running an HTTPS proxy server on the same
# host as Krill and Krill is listening on localhost only, then
# using plain HTTP between Krill and your proxy server may be an
# option you would want to use.
#
# We recommend STRONGLY that you do not use this option if your
# Krill instance is configured to bind to a public IP address.
# 
### https_mode = "generate"
```

PR #915, opened 2022-09-01
Allow plain HTTP for --server in krillc. #913

timbru:
Essentially this is just replacing `uri::Https` with `idexchange::ServiceUri` (both from `rpki-rs) as the latter allows both http and https.

### DIFF
--- a/src/cli/options.rs
+++ b/src/cli/options.rs
@@ -2,7 +2,6 @@
 use std::collections::HashMap;
 
 use std::{
-    convert::TryFrom,
     path::PathBuf,
     str::{from_utf8_unchecked, FromStr},
     {env, fmt},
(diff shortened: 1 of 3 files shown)


src/cli/options.rs
```diff
@@ -54,15 +54,15 @@ impl GeneralArgs {
     fn from_matches(matches: &ArgMatches) -> Result<Self, Error> {
         let server = {
             let mut server = match env::var(KRILL_CLI_SERVER_ENV) {
-                Ok(server_str) => Some(uri::Https::try_from(server_str)?),
+                Ok(server_str) => Some(idexchange::ServiceUri::try_from(server_str)?),
```
ximon18, line 57: Why `try_from` here and `from_str` in the other uses in this file?

ximon18 (approved):

src/cli/options.rs
timbru: no good reason. The old code was doing the same (but then on uri::Https) and this was kept.

That said, I agree that it reads better to just do `from_str` so I modified the code :)

PR #921, opened 2022-09-07
Fix failing Docker publication on release tag push (fixes #919)

### LINKED ISSUE #919: Docker push fails for release tag
GitHub Actions workflow `pkg` run https://github.com/NLnetLabs/krill/actions/runs/2994084685 failed at the final step: `docker-manifest`.

Actually it failed earlier than that at the matrix of `docker` jobs but the steps themselves didn't fail, rather they pushed the created image to the wrong tag in Docker Hub. The manifest push step then failed because the separate component images didn't exist with the expected tags.

The root cause of this is two-fold:
- Firstly, the `docker/metadata-action@v3` returned multiple line-break separated tags for `nlnetlabs/krill:v0.10.0` and `nlnetlabs/krill:latest`. This didn't happen when the worfkflow was tested by forking this repo to the `ximon18/krill` repo because the `ximoneighteen` Docker Hub registry didn't have an existing `ximoneighteen/krill:latest` tag so only one tag resulted from the `docker/metadata-action@v3` lookup. We should either get a single tag only or handle the line breaks.

- Secondly, the old `:latest` tag needs to be replaced by a pushed manifest rather than a pushed tag. Unlike tags, a manifest can't be copied to a new name and pushed again. We therefore need to call `docker manifest create` and `docker manifest push` twice each, once for the `nlnetlabs/krill:v0.10.0` tag and once for the `nlnetlabs/krill:latest` tag.

I've fixed the issue manually for now by downloading the docker image tarballs from the failed workflow saved artifacts, building the manifests locally and pushing them, as the alternative was a delay while working out how to fix the workflow and long delays testing and finally running the modi The workflow however still needs fixing to avoid this failure.

ximon18:
This PR fixes the incorrect handling of the [`docker/metadata-action`](https://github.com/docker/metadata-action) output and publication of Docker Manifests in the case of a push of a "release" tag, for which the action output consists of multiple lines which the workflow logic was not expecting nor was the workflow prepared to make multiple Docker Manifests in this case,

Before going further it is worth saying something about terminology for referring to Docker images. Even the official [`docker tag`](https://docs.docker.com/engine/reference/commandline/tag/) documentation talks about "image name", "name components", "tag name", "names and tags" and "tag" without making it clear which bit is the "tag" and conflating "tag" as a noun and "tag" as a verb.... In the context of the `docker/metadata-action` the "tag(s)" that it outputs are, in Docker Hub UI terminology (i.e. Organizations, Repositories and Tags), one or more line-break separated "&lt;repo&gt;:&lt;tag&gt;" strings, where the full image name (excluding the optional registry prefix) is of the form "&lt;org&gt;/&lt;repo&gt;:&lt;tag&gt;", e.g. "nlnetlabs/krill:v0.10.0" and the `docker/metadata-action` "tag" would be "krill:v0.10.0". 

Adding to the confusion is the fact that we have both Git tags and Docker tags whose value sets overlap but are not equal. A Git tag named "v*" without a dash suffix, e.g. "v0.1.2", is a "release" tag for our Docker purposes, but "0.1.2" is not and neither is "v0.1.2-rc1".

When, and only when, a "release" tag is pushed to GitHub, the `docker/metadata-action` will output two tags: one for the tag itself, e.g. "krill:v0.10.0" and a "krill:latest" tag. This is because when users execute `docker run nlnetlabs/krill` (i.e. without a version suffix) Docker will actually pull the "nlnetlabs/krill:latest" image, and we only want users to get a new "latest" image when that image is considered suitable for release, i.e. not if it is based on a development or main branch or a release candidate, only for final releases.

When previously we were publishing images just for the Docker amd64 architecture we didn't need to use Docker Manifests, we just pushed images with the appropriate full image name and used the `docker tag` command to create a new "latest" tag referencing the just created "nlnetlabs/krill:vX.Y.Z" image. When using Docker Manifests this is no longer possible because there is no command to create a manifest referencing an existing manifest, manifests only reference images. So instead we now have to create two manifests in the release tag case: one for the release tag itself, and one for the "latest" tag. And we have to publish both of those manifests to Docker Hub.

This PR:
- Fixes the workflow to correctly handle single and dual tag output from the `docker/metadata-action` and to correctly create and publish either one or two Docker Manifests.
- Cleans up the workflow to make it easier to understand, normalizing, clarifying and documenting the image/tag terminology used.
- Simplifies the usage of the `docker/metadata-action` (see below).

Some context regarding this last bullet point: Prior to this PR the workflow would run `docker/metadata-action` for each matrix variant of the `docker` job each time producing and handling the same output and the `docker` job was configured to output a value based on the step that ran the `docker/metadata-action`. But a matrix job can only have a single output so which of the N matrix variant runs was the output value based on? Answer: any, it didn't matter as they all produce the same value, but still it was confusing. So now the `docker/metadata-action` step has been moved out into a new `docker_pre` job that is run only once and referenced by both the `docker` and `docker_manifest` jobs in the workflow.

Example runs of the workflow in this PR:
- https://github.com/NLnetLabs/krill/actions/runs/3005666173 shows the development branch case. A single "krill:test" tag is output by `docker/metadata-action` (see [here](https://github.com/NLnetLabs/krill/runs/8222481799?check_suite_focus=true#step:2:46), the step that publishes images to Docker Hub is skipped (see [here](https://github.com/NLnetLabs/krill/runs/8222709637?check_suite_focus=true)) and the job for creating and publishing Docker Manifests is skipped (see [here](https://github.com/NLnetLabs/krill/runs/8223010760?check_suite_focus=true)).
- https://github.com/NLnetLabs/krill/actions/runs/3005868653 shows the development branch with the `docker-manifest` job force enabled (for test purposes, usually it will not run except on push of a Git tag) but with the `docker manifest` commands prefixed by `echo` (see [here](https://github.com/NLnetLabs/krill/runs/8223642292?check_suite_focus=true#step:4:4) and [here](https://github.com/NLnetLabs/krill/runs/8223642292?check_suite_focus=true#step:5:4)) to see their intended effect but to avoid actually altering anything on Docker Hub.
- https://github.com/NLnetLabs/krill/actions/runs/3006612928 forces (for test purposes) `docker/metadata-action` to output a "latest" tag (see [here](https://github.com/NLnetLabs/krill/runs/8225142635?check_suite_focus=true#step:2:46)) so that we can see that the two "tag" case is handled correctly and causes two identical but differently named Docker Manifests to be created and published (except that as above the publication command is prefixed with `echo` to prevent actually altering Docker Hub) (see [here](https://github.com/NLnetLabs/krill/runs/8225748330?check_suite_focus=true#step:4:20) and [here](https://github.com/NLnetLabs/krill/runs/8225748330?check_suite_focus=true#step:5:16)).

### DIFF
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -52,7 +52,11 @@
 #
 # Per architecture images are built and pushed to Docker Hub with xxx-<arch> tags, and then a Docker Manifest is
 # created which groups these images into a single multi-arch image without the -<arch> extension on the tag. The
-# manifest is then also pushed to Docker Hub.
+# manifest is then also pushed to Docker Hub. In the case of a release tag (a "v*" Git tag that does not contain a
+# dash "-" character, i.e. is not a release candidate e.g. v0.1.2-rc3) two manifests are pushed: one for the release
+# tag, e.g. nlnetlabs/<app>:v0.1.2; and one for the "latest" tag, e.g. "nlnetlabs/<app>:latest" (to enable Docker users
+# to just run `docker run nlnetlabs/<app>` and get the latest stable release without having to know/specify the actual
+# version).
 #
 # Building of both x86-64 and non-x86-64 architecture images are handled by a single Dockerfile which supports two
 # modes of operation. In the default 'build' mode our app is compiled within the Docker container and only the final
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -884,6 +884,7 @@ jobs:
     runs-on: ubuntu-20.04
     outputs:
       repo_and_tag_pairs: ${{ steps.encode.outputs.repo_and_tag_pairs }}
+      json: ${{ steps.encode.outputs.json }}
     steps:
       # Based on the Git ref, e.g. a tag or branch, determine the appropriate tag(s) for the Docker image we will
       # create, e.g. <app>:unstable for a main branch commit, or <app>:v1.2.3 _and_ <app>:latest for a release tag, or
@@ -943,8 +944,11 @@ jobs:
       - name: Encode outputs for passing safely to downstream jobs
         id: encode
         run: |
-          ENCODED_REPO_AND_TAG_PAIRS=$(echo ${{ steps.meta.outputs.tags }} | base64)
+          ENCODED_REPO_AND_TAG_PAIRS=$(echo ${{ steps.meta.outputs.tags) }} | base64)
           echo "::set-output name=repo_and_tag_pairs::${ENCODED_REPO_AND_TAG_PAIRS}"
+
+          ENCODED_JSON_TAGS=$(echo ${{ steps.meta.outputs.json }} | base64)
+          echo "::set-output name=json::${ENCODED_JSON}"
 
 
   # -------------------------------------------------------------------------------------------------------------------
@@ -1032,7 +1036,7 @@ jobs:
           # Don't use the [1]th tag (if present) which would be '<repo>:latest', as we won't build or publish an
           # image for 'latest' but instead push a multi-arch manifest for 'latest' which refers to the same
           # set of architecture specific images as the version specific multi-arch manifest we will publish.
-          DOCKER_REPO_AND_TAG=$(echo "${{ needs.docker_pre.outputs.repo_and_tag_pairs[0] }}" | base64 -d)
+          DOCKER_REPO_AND_TAG=$(echo "${{ fromJSON(needs.docker_pre.outputs.json).tags[0] }}" | base64 -d)
           echo "::set-output name=image_name::${DOCKER_ORG}/${DOCKER_REPO_AND_TAG}-${{ matrix.shortname }}"
 
       # Build a single architecture specific Docker image with an explicit architecture extension in the Docker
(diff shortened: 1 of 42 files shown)


ximon18:
I've marked this PR as a DRAFT because it currently contains some deliberate changes to show behaviour for test purposes without actually impacting Docker Hub. I'll revert those changes when the last of my GH Actions runs completes and will then make this PR READY.

timbru:
Question. Can this target `prep-0.11.0-rc1` instead of main? I just want to avoid merge issues with that. Also, the issue could then be moved into the 0.11 project board.

ximon18:
> Question. Can this target `prep-0.11.0-rc1` instead of main? I just want to avoid merge issues with that. Also, the issue could then be moved into the 0.11 project board.

It can but (a) the changes are not related to any particular version of Krill, and (b) just know that when merged to `main` I don't recall if GH will then run the action based on the previous committed version of the workflow rather than the one contained in this PR. Not necessarily a problem as you only need these changes when making a Git release tag.

timbru:
I think targeting `prep-0.11.0-rc1` would be a nicer work flow. The tag for v0.11.0-rc1 will only be done once that is merged.

ximon18:
This PR is obsoleted by PR #931. PR #931 switches to using a reusable workflow defined in the organization wide NLnetLabs/.github repository which includes the work done for this PR. PR #931 should be preferred over this PR.

Issue #929, opened 2022-09-12
Simplify Krill upgrade version checks


timbru:
The logic is too complicated right now resulting in issue #928. We should simplify this and do all the version checks in one place so it's easy to see what is going on, and it will be easy to test that the right upgrade code is called.

timbru:
Done as part of #1094

PR #930, opened 2022-09-12
Do not skip migration of pubd 0.9.0 (fix: 928)

timbru:
Note: testing this is quite awkward because of the complicated way that is used to determine what migrations are needed.

A better fix would be to simplify that logic. I believe that the proper change would be too involved for a quick bug fix release, so this current change is just a quick fix. I created issue #929 so that we can do this properly for release 0.11.0.


ximon18 (approved):
I haven't tested it but it seems sound in principle.

timbru:
I have tested it manually actually, but I don't have an automated test for it.

PR #918, opened 2022-09-05
Silence "unable to initialize frontend: Dialog" debconf warnings

ximon18:
Set DEBIAN_FRONTEND env var at the workflow root level so that we don't have to ensure we always set it where needed, e.g. from https://github.com/NLnetLabs/krill/runs/8188128421?check_suite_focus=true#step:12:29:

```
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
debconf: unable to initialize frontend: Readline
debconf: (Can't locate Term/ReadLine.pm in @INC (you may need to install the Term::ReadLine module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.34.0 /usr/local/share/perl/5.34.0 /usr/lib/x86_64-linux-gnu/perl5/5.34 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl-base /usr/lib/x86_64-linux-gnu/perl/5.34 /usr/share/perl/5.34 /usr/local/lib/site_perl) at /usr/share/perl5/Debconf/FrontEnd/Readline.pm line 7, <> line 1.)
debconf: falling back to frontend: Teletype
dpkg-preconfigure: unable to re-open stdin: 
```

A run of the workflow changes can be seen here: https://github.com/NLnetLabs/krill/actions/runs/2993179220

PR #931, opened 2022-09-13
Migrate to NLnetLabs reusable packaging workflow

ximon18:
This PR obsoletes the following currently open PRs:
- #918
- #921

This PR replaces the current Packaging `pkg.yml` workflow with a much simpler version that outsources the behaviour to a new equivalent reusable workflow defined in the NLnet Labs organisation wide https://github.com/NLnetLabs/.github repository, which includes the functionality of the above obsoleted PRs.

A successful test run against this PR branch can be seen here: https://github.com/NLnetLabs/krill/actions/runs/3036529911

The test run doesn't exercise the behaviour of pushing Docker images and manifests to Docker Hub as that only happens when a release Git tag (i.e. a tag named `v*` without a `-XXX` suffix) is pushed to this Git repository, but was tested elsewhere as part of the development of the centralized reusable workflow (e.g. [here](https://github.com/NLnetLabs/.github/pull/6#issuecomment-1244335732)).

The centralized reusable packaging workflow (see https://github.com/NLnetLabs/.github/blob/master/.github/workflows/pkg-rust.yml) needs more documentation, but in short it is the result of combining and normalizing various fixes and improvements that were added to the Krill and Routinator workflows over time. While similar to the `pkg.yml` workflow defined in this repository prior to this PR, it is sufficiently changed so as to make a direct comparison via diff quite difficult. Changes have been made to make the workflow more generic and reusable and thus for example the set of architectures and operating systems to build for is no longer hard-coded but now passed in as a set of rules, and the scripts to run to test installed packages post install and upgrade are now stored as separate files in the repo and passed by filepath reference to the reusable workflow.

Note: This PR also modifies the way RPM scriptlets are defined because changes were needed to fix issues seen in the Krill and Routinator projects due to the limited way that the [`cargo-generate-rpm`](https://github.com/cat-in-136/cargo-generate-rpm) cargo helper command permits RPM scriptlets to be defined (i.e. they must be defined inline in TOML, they cannot be defined as references to Bash files stored separately).


.github/workflows/pkg.yml
```diff
@@ -1,1079 +1,39 @@
...
+  package:
+    uses: NLnetLabs/.github/.github/workflows/pkg-rust.yml@v1
+    secrets:
+      DOCKER_HUB_ID: ximoneighteen
+      DOCKER_HUB_TOKEN: ${{ secrets.TEST_DOCKER_HUB_TOKEN }}
+    with:
+      #docker_org: ${{ github.repository_owner }}
+      #docker_repo: krill
+      docker_org: ximoneighteen
+      docker_repo: ximontest
```
timbru, line 25: was this for testing? should the commented lines be used again?

timbru (approved):

PR #932, opened 2022-09-13
Show repo init settings

timbru:
Added to the repo stats endpoint (/stats/repo). Example output:

```
{
  "publishers": {
    "testbed": {
      "objects": 3,
      "size": 3762,
      "manifests": [
        {
          "uri": "rsync://localhost/repo/testbed/0/217876AA966B965A6EDFDED43C469B90ED11AC6D.mft",
          "this_update": "2022-09-13T15:10:12Z",
          "next_update": "2022-09-14T18:18:12Z"
        }
      ]
    },
    "ca": {
      "objects": 5,
      "size": 8247,
      "manifests": [
        {
          "uri": "rsync://localhost/repo/ca/0/CC2487CF3A9C774BFAE2DCE4DD8368441C75C720.mft",
          "this_update": "2022-09-13T15:10:12Z",
          "next_update": "2022-09-14T16:01:12Z"
        }
      ]
    },
    "ta": {
      "objects": 3,
      "size": 3663,
      "manifests": [
        {
          "uri": "rsync://localhost/repo/ta/0/0A6EA673F04AAA345D605A38E7710A2D8413B56C.mft",
          "this_update": "2022-09-13T15:10:12Z",
          "next_update": "2022-09-14T17:02:12Z"
        }
      ]
    }
  },
  "session": "0102f89f-3639-40cb-a967-68789c7da891",
  "serial": 25,
  "last_update": "2022-09-13T15:15:13.427709Z",
  "sia_base": "rsync://localhost/repo/",
  "rrdp_notification_uri": "https://localhost:3000/rrdp/notification.xml"
}
```

### DIFF
--- a/src/pubd/repository.rs
+++ b/src/pubd/repository.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::{HashMap, VecDeque},
+    convert::TryFrom,
     fmt, fs, mem,
     path::{Path, PathBuf},
     str::FromStr,
@@ -14,7 +15,7 @@ use rpki::{
         publication::{ListReply, PublicationCms, PublishDelta},
     },
     crypto::KeyIdentifier,
-    repository::x509::Time,
+    repository::{x509::Time, Manifest},
     rrdp::Hash,
     uri,
 };
@@ -98,7 +99,6 @@ impl RepositoryContentProxy {
                 let publishers = HashMap::new();
 
                 let session = RrdpSession::default();
-                let stats = RepoStats::new(session);
 
                 let mut repo_dir = work_dir.to_path_buf();
                 repo_dir.push(REPOSITORY_DIR);
@@ -106,7 +106,7 @@ impl RepositoryContentProxy {
                 let rrdp = RrdpServer::create(rrdp_base_uri, &repo_dir, session);
                 let rsync = RsyncdStore::new(rsync_jail, &repo_dir);
 
-                RepositoryContent::new(publishers, rrdp, rsync, stats)
+                RepositoryContent::new(publishers, rrdp, rsync)
             };
 
             // Store newly initialized repo content on disk
@@ -139,7 +139,7 @@ impl RepositoryContentProxy {
 
     /// Return the repository content stats
     pub fn stats(&self) -> KrillResult<RepoStats> {
-        self.read(|content| Ok(content.stats().clone()))
+        self.read(|content| Ok(content.stats()))
     }
 
     /// Add a publisher with an empty set of published objects.
@@ -240,25 +240,16 @@ impl RepositoryContentProxy {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct RepositoryContent {
     publishers: HashMap<PublisherHandle, CurrentObjects>,
-
     rrdp: RrdpServer,
     rsync: RsyncdStore,
-
-    stats: RepoStats,
 }
 
 impl RepositoryContent {
-    pub fn new(
-        publishers: HashMap<PublisherHandle, CurrentObjects>,
-        rrdp: RrdpServer,
-        rsync: RsyncdStore,
-        stats: RepoStats,
-    ) -> Self {
+    pub fn new(publishers: HashMap<PublisherHandle, CurrentObjects>, rrdp: RrdpServer, rsync: RsyncdStore) -> Self {
         RepositoryContent {
             publishers,
             rrdp,
             rsync,
-            stats,
         }
     }
 
@@ -266,13 +257,11 @@ impl RepositoryContent {
         let publishers = HashMap::new();
         let rrdp = RrdpServer::create(rrdp_base_uri, repo_base_dir, session);
         let rsync = RsyncdStore::new(rsync_jail, repo_base_dir);
-        let stats = RepoStats::new(session);
 
         RepositoryContent {
             publishers,
             rrdp,
             rsync,
-            stats,
         }
     }
 
@@ -280,10 +269,6 @@ impl RepositoryContent {
     pub fn clear(&self) {
         self.rrdp.clear();
         self.rsync.clear();
-    }
-
-    pub fn stats(&self) -> &RepoStats {
-        &self.stats
     }
 }
 
@@ -317,16 +302,12 @@ impl RepositoryContent {
         // to update outside of its jail.
         let objects = self.objects_for_publisher_mut(name)?;
         objects.apply_delta(delta.clone(), jail)?;
-        let publisher_stats = PublisherStats::new(objects, Time::now());
 
         // update the RRDP server
         self.rrdp.publish(delta, jail, config)?;
 
         // write repo (note rsync is based on updated rrdp snapshot)
         self.write_repository(config)?;
-
-        // Update publisher stats
-        self.stats.publish(name, publisher_stats, self.rrdp.notification());
 
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


ximon18:
Is this about `sia_base` and `rrdp_notification_uri`?

What's the reason for this PR?

src/pubd/repository.rs
```diff
@@ -364,18 +343,45 @@ impl RepositoryContent {
...
             self.rrdp.publish(delta, jail, config)?;
-            self.stats.remove_publisher(name, self.rrdp.notification());
-
-            self.write_repository(config)
-        } else {
-            // nothing to remove
-            Ok(())
+            self.write_repository(config)?;
         }
+        // remove publisher is present
```
ximon18, line 354: s/is/if/ present?

src/pubd/repository.rs
```diff
@@ -364,18 +343,45 @@ impl RepositoryContent {
...
+        self.publishers.remove(name);
```
ximon18, line 355: This looks like a logic fix unrelated to the PR? Is that correct?

src/pubd/repository.rs
```diff
@@ -1285,7 +1302,7 @@ impl RepoStats {
     pub fn stale_publishers(&self, seconds: i64) -> Vec<PublisherHandle> {
         let mut res = vec![];
         for (publisher, stats) in self.publishers.iter() {
-            if let Some(update_time) = stats.last_update {
+            if let Some(update_time) = stats.last_update() {
```
ximon18, line 1305: This looks like a logic fix unrelated to the PR? Is that correct?

src/pubd/repository.rs
```diff
@@ -1311,11 +1328,14 @@ impl RepoStats {
 
 impl fmt::Display for RepoStats {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        writeln!(f, "RSYNC BASE:        {}", self.sia_base)?;
+        writeln!(f, "RRDP NOTIFICATION: {}", self.rrdp_notification_uri)?;
```
ximon18, line 1332: Why not "URI" ?

src/pubd/repository.rs
```diff
@@ -1311,11 +1328,14 @@ impl RepoStats {
 
 impl fmt::Display for RepoStats {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        writeln!(f, "RSYNC BASE:        {}", self.sia_base)?;
```
ximon18, line 1331: Why "RSYNC" and not "SIA" at in the stats endpoint?

src/pubd/repository.rs
```diff
@@ -1363,17 +1377,81 @@ impl PublisherStats {
...
     }
 
+    /// Returns the most recent "this_update" time
+    /// from all manifest(s) published by this publisher,
+    /// if any.. i.e. there may be 0, 1 or many manifests
     pub fn last_update(&self) -> Option<Time> {
-        self.last_update
+        let mut last_update = None;
+        for mft in self.manifests() {
+            if let Some(seen_update) = last_update {
```
ximon18, line 1386: Introducing a new `seen_update` is confusing, why not just shape the existing `last_update`? Ah I see, because of the assignment... still I found it confusing.

src/pubd/repository.rs
```diff
@@ -1363,17 +1377,81 @@ impl PublisherStats {
...
+    pub fn next_update(&self) -> Time {
+        self.next_update
+    }
+}
+
+impl TryFrom<&Manifest> for PublisherManifestStats {
+    type Error = ();
+
+    fn try_from(mft: &Manifest) -> Result<Self, Self::Error> {
+        let uri = mft.cert().signed_object().cloned().ok_or(())?;
```
ximon18, line 1450: What will `uri` be here if this fails?

src/upgrades/pre_0_9_0/pubd_objects_migration.rs
```diff
@@ -331,9 +326,6 @@ struct OldRepository {
 
     rrdp: OldRrdpServer,
     rsync: RsyncdStore,
-
-    #[serde(default = "RepoStats::default")]
-    stats: RepoStats,
```
ximon18, line 336: Why is the serialization of `OldRepository` being changed?

ximon18 (commented):
Quick review done on my phone, but ideally I'd review it on a bigger screen, it's a bit hard to follow on a phone. I haven't tested it, would also be good to do that. I don't see any tests updated in this PR to follow the changes... is that correct? Also, am I right in thinking a manual update PR doesn't exist yet for these changes to the API (non-breaking I think) and CLI (breaking, if we even give any guarantees about the text format output)?

src/pubd/repository.rs
timbru: Yeah, that is correct. It was leaving the publisher in the map for the RepositoryContent, with no actual published objects. It doesn't really hurt to do so, but I noticed when touching the code, so figured I would just also remove the map entry.

src/pubd/repository.rs
timbru: fixed

timbru:
> Is this about `sia_base` and `rrdp_notification_uri`?
> 
> What's the reason for this PR?

It was requested by people who use Krill as a Publication Server. It is, was, not possible to query which values were used for `sia_base` and `rrdp_notification_uri` to initialise the server. With this PR we extend the existing `RepoStats` structure - which is currently only being used by the CLI (not the UI) to include this.

I can see how some of the changes to `RepoStats` can be confusing. We were keeping the stats in memory and updating them on access by a publisher. But this does not include any info on the current server settings. So, we would need to get that information dynamically anyway, and could not have sensible defaults for existing values.

In the end I decided to just remove the object from the struct - i.e. it will be ignored when deserializing existing. It is now always calculated dynamically. This adds only minimal overhead as this is queried infrequently.

 But, as a result I made some changes to how the 'last_update' is determined. Instead of setting this on access by a publisher, we now derive this to be the latest 'this_update' time from all _syntactically correct_ manifests found for the publisher.

src/upgrades/pre_0_9_0/pubd_objects_migration.rs
timbru: It was just removed, so it will be ignored, because it no longer needs to be included in the new structure for the migration.

src/pubd/repository.rs
timbru: Because of the inclusion of server information I needed to change to a dynamically calculated structure. This change is a result of that.

src/pubd/repository.rs
timbru: renamed things more explicitly - hopefully improving readability

src/pubd/repository.rs
timbru: An `Err(())`. In this case treat this manifest as syntactically incorrect (which it is - I added a comment as well now) - and we do not include it in our stats about this publisher.

Note that other API end-points let operators review the complete set of objects for publishers. This is just a quick overview. Also, incorrect objects should not be found in practice - as Krill and other conformant CAs will include this.

src/pubd/repository.rs
timbru: I wanted it to be similar to the CLI, but thanks for raising this, because it actually wasn't..

Changed it now to report the rsync and rrdp base URIs as they are also used when initialising the server and renamed the fields appropriately.

src/pubd/repository.rs
timbru: see other comment

timbru:
re testing: I don't think we need to test the text representation, but the last_updated per publisher stat is actually used and tested wrt child suspension/unsuspension.

Issue #935, opened 2022-09-14
Local publishing without HTTPS/CMS


timbru:
This should help increase performance in cases where there are many CAs publishing at a local publication server (i.e. inside the same Krill instance).

PR #936, opened 2022-09-15
Local publishing without HTTPS/CMS #935

timbru:
With these changes CAs with a local publication server should bypass HTTPS and CMS for RFC 8181 (publication) and this should result in a hopefully big performance improvement.

Automated tests are okay. Will now build a package and try it in our test VMs.


timbru:
This will need to target prep-0.12.0-rc1 after 0.11.0 is released

ximon18 (approved):

PR #937, opened 2022-09-15
Temporarily use unreleased bug fix branch of the reusable packaging workflow.

ximon18:
To work around this error: https://github.com/NLnetLabs/krill/actions/runs/3060211032/jobs/4938451367#step:9:40


ximon18:
This PR is no longer required as the issue has been fixed in a new release of the reusable packaging workflow.

Issue #939, opened 2022-09-16
Improve speed of key value store reads


timbru:
As it turns out we can deserialize 100 times faster if we read a file into memory first and then call `serde_json::from_reader` rather than streaming the file.

This can have an impact on memory usage, but.. it's minimal for small installations and it's a price worth paying for large installations.

timbru:
Gist, change the following in `KeyValueStoreDiskImpl::get`
```
let v = serde_json::from_reader(f)?;
```

To:
```
let mut bytes = Vec::new();
f.read_to_end(&mut bytes).unwrap();
let v = serde_json::from_slice(&bytes)?;
```

timbru:
done

PR #942, opened 2022-09-20
Improve speed of key value store reads #939

timbru:


src/commons/eventsourcing/kv.rs
```diff
@@ -341,14 +341,39 @@ impl KeyValueStoreDiskImpl {
...
+            // from a benchmark server with 5000 CAs and 10 ROAs per CA.
+            //
+            // So, we do this in memory. This should not be an issue for our application because
+            // we only read large files (snapshots) during startup. After this data is kept in
+            // memory. I.e. we do not expect that large files are read concurrently resulting
+            // in high memory consumption.
+            //
+            // Furthermore, this has a negligible impact on installations with a single small
+            // CA instance. And large operations - i.e. a bug Publication Server or Parent CA
+            // to lots of children, can afford a server with a decent amount of memory.
```
ximon18, line 360: With a minimal bump to MSRV from 1.56 to 1.57 you could use [Vec::try_reserve_exact](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.try_reserve_exact) to attempt to pre-allocate the required amount of memory and fall back to the old method if insufficient memory is available. Is this worth considering or would you then consider that panic'ing is the right thing to do anyway as the server will need that memory to operate normally anyway?

ximon18 (approved):

src/commons/eventsourcing/kv.rs
timbru: I think that the memory usage is minimal for small setups so it does not matter. And organisations who run a big parent CA, or more likely Publication Server, can allocate enough memory. I think panicking is probably right in this case.

Issue #693, opened 2021-10-19
Optionally stage publication by publishers to prevent excessive RRDP deltas


timbru:
Stage deltas received by publisher up to X minutes (should be configurable) to prevent that we get too many deltas and RPs fall back to snapshots.

Publishing more frequently than once per minute adds little or no value. Given the update interval of 10 minutes that is widely used, something like publishing a delta every 5 minutes feels about right.

PR #943, opened 2022-09-27
Use write-ahead log for RepositoryContent.

timbru:
[draft] reviewing allowed, but can wait for the following:

Will deploy to benchmark environment to test upgrade and effect.
Still need to update the developer docs.


timbru:
Deployed to benchmarking VM with 5000 CAs with 10 ROAs each.

The system is responsive and CPU load is low, except for when publication happens. With the additional log messages added we can see that this is due to updating the RRDP state and writing new RRDP files for every update.

This should be solved when https://github.com/NLnetLabs/krill/issues/693 is implemented. That will allow us to defer RRDP updates and include multiple publisher deltas in one. Furthermore, this will mean that the RRDP are not blocking publisher.

The RRDP update process itself can also be optimised more. Currently it's creating the snapshot XML twice (and for a 280 MB XML file this is what's blocking the system). Once to get the hash to include in the notification, and once to write the actual file. We may want to optimise this, and perhaps use the RRDP support from rpki-rs rather than the older RRDP XML support in krill itself.

timbru:
This needs more work - it's currently saving the entire new snapshot in an RrdpUpdate change. This naive implementation ends up using way too much data. So, will change this and only then mark this for review. The CPU usage seems quite improved at least.

timbru:
The deltas are now much smaller. Overall performance is still not good enough because creating new RRDP snapshots is still too expensive and happens too often. This will be fixed in a separate PR for issue #693.

This should be good to merge now, unless other issues are raised during review of course.

src/commons/error.rs
```diff
@@ -649,6 +657,9 @@ impl Error {
 
             // internal server error
             Error::AggregateStoreError(e) => ErrorResponse::new("sys-store", &self).with_cause(e),
+            
+            // internal server error
+            Error::WalStoreError(e) => ErrorResponse::new("sys-wal-store", &self).with_cause(e),
```
ximon18, line 662: You distinguish wal store error from aggregate store error here, but above you use the same description string for both. This seems inconsistent.

src/commons/eventsourcing/wal.rs
```diff
@@ -0,0 +1,352 @@
...
+        path.push(name_space);
+
+        let kv = KeyValueStore::disk(krill_data_dir, name_space)?;
+        let cache = RwLock::new(HashMap::new());
+
+        Ok(WalStore { kv, cache })
+    }
+
+    /// Warms up the store: caches all instances.
+    pub fn warm(&self) -> WalStoreResult<()> {
```
ximon18, line 139: Minor: This is doing something (i.e. warming up) rather than being "warm" so wouldn't it be better named `warmup()` or similar?

ximon18 (approved):

src/commons/error.rs
timbru: updated the error text to be a bit more clear (none of this is likely to be seen by users, but still)

src/commons/eventsourcing/wal.rs
timbru: I prefer to keep it like this. This is consistent with `AggregateStore::warm`. Fwiw.. "warm" was intended as a verb, and the imperative at that. For being warm I would use something like "is_warm" or "warmed" (i.e. past tense, not the imperative).

src/commons/eventsourcing/wal.rs
```diff
@@ -0,0 +1,352 @@
...
+    /// Add a new entity for the given handle. Fails if the handle is in use.
+    pub fn add(&self, handle: &MyHandle, instance: T) -> WalStoreResult<()> {
+        let instance = Arc::new(instance);
+        let key = Self::key_for_snapshot(handle);
+        self.kv.store_new(&key, &instance)?; // Fails if this key exists
+        self.cache.write().unwrap().insert(handle.clone(), instance);
+        Ok(())
+    }
+
+    /// Checks whether there is an instance for the given key
```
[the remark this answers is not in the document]
timbru: fixed

src/commons/eventsourcing/wal.rs
timbru: To be clear - I think `warmup` would have been better, but I don't think that `warm` on its own is too problematic. I am happy to change after all the stacked PRs are merged - because I fear conflicts.

Issue #945, opened 2022-09-30
Deprecate keeping old RRDP files and dirs in Krill


timbru:
Krill currently tries to keep old RRDP files for a while after these files are no longer referenced in the notification file.

This adds a fair amount of complexity, and in practice this is managed - and managed better - by krill-sync.

In future we should merge krill-sync into krill itself, but for now we can simply deprecate this and let krill keep current RRDP files only.

PR #946, opened 2022-10-03
Deprecate keeping old RRDP files and dirs in Krill (#945)

timbru:
NOTE: This will simplify the Krill code and help to make it more efficient. Ensuring that old snapshot and delta files (and rsync for that matter) are available for long enough is handled better by krill-sync


src/pubd/repository.rs
```diff
@@ -906,14 +889,6 @@ impl RrdpServer {
         keep
     }
 
-    // Find where to truncate old notification files
```
ximon18, line 909: Shouldn't this PR also remove the no longer used config settings, e.g. `retention_old_notification_files_seconds` ?

ximon18 (approved):

src/pubd/repository.rs
timbru: yes.. now removed

Issue #950, opened 2022-10-07
Use per CA locks in CA aggregate store.


timbru:
We already have `CaLocks` which implements a "Handle" to lock structure. But the `AggregateStore` uses an `outer_lock` for all updates to any aggregate.

This means that changes to any CA can only happen sequentially. Per CA (Aggregate) locks could help to improve performance here.

PR #951, opened 2022-10-10
Combine rrdp deltas

timbru:
Implements support for issue #693.

See image.. load in server where 500 CAs with 20 ROAs are created as a benchmark test. CPU load is high, bound to a single CPU during setup (issue #950 will most likely improve this). When this is done publication of the next RRDP delta is postponed as configured - we used 5 minutes for the min interval). So, now publication of objects is actually very fast...

Screenshot 2022-10-10 at 13 41 39

Before implementing this issue (to have a minimal interval time before the next RRDP delta) CPU usage would be high for a prolonged time as new snapshots would be continuously written. In case we had 5000 CAs CPU usage would remain high as work got added faster than it was resolved. Will run this test now as well. In the meantime this should now be ready for review.


timbru:
Changed base to `simplify-rrdp-cleanup` for review. This should be merged after that one, PR #946, and before *that* actually PR #943 are merged.

defaults/krill-pubd.conf
```diff
@@ -148,33 +148,43 @@
...
+# RRDP deltas will be produced no more frequently than the specified interval. If the server
+# is quiet, then typically the first update sent to it will result in an RRDP update straightaway.
+# But, if the server is busy, then subsequent updates will be held until the interval period has
+# passed since the last RRDP update.
 #
-# retention_old_notification_files_seconds = 600 (10 mins)
+# By default every change gets its own RRDP delta. But on systems with many publishers it is
+# a good idea to limit the amount of deltas a bit. E.g. for a system with 5000 CAs with 20 ROAs
+# each a higher interval is needed - otherwise the system will be too busy writing large snapshot
+# files and essentially won't be able to handle the (normal, mft/crl renewal) publicaion load.
```
ximon18, line 175: s/publicaion/publication/

defaults/krill-pubd.conf
```diff
@@ -148,33 +148,43 @@
...
+#
+# Since Relying Party software (validators, such as routinator) use update intervals in the order
```
ximon18, line 177: s/routinator/Routinator/

defaults/krill-pubd.conf
timbru: fixed

defaults/krill-pubd.conf
timbru: fixed

PR #954, opened 2022-10-10
Per aggregate locks

timbru:
[draft] This should target `prep-0.12.0-rc1` once PR #951 is merged. Branched that because changes are in the same area of the code.

Essentially this PR removed the `CaLocks` we had in `CaManager` and adds an equivalent `AggregateLocks` which will be used by `AggregateStore`. The latter was using a global lock structure which means that only one aggregate could be updated at a time. This should be better now with this move. Furthermore, by having the locks in the AggregateStore we ensure that they are always used - it was easy to forget - and it should be easier to support multi-node locks in future?

I looked into making the locking async, but that change is too complicated at this time. It looks like we don't need this at this time from a performance point of view - so let's postpone async to a later moment.


timbru:
Merging this is hard because it was branched off another branch that was squashed when merged. It's really all in the final two commits.. so what I will do is: create a new branch and PR with those commits cherry picked and remove this PR..

timbru:
Replaced by https://github.com/NLnetLabs/krill/pull/956

Issue #955, opened 2022-10-12
Listen on IPv4+IPv6


ties:
As a krill user, I want to be able to configure krill to listen on multiple addresses (IPv4+IPv6) to prevent issues with dual-stack hostnames and some clients (happy eyeballs) working and others not (embedded http clients) when running on a dual-stack host

timbru:
The usual setup is that Krill listens on IPv4 localhost only, and a proxy server is used in front of it. The proxy server can do IPv4 and IPv6, proper HTTPS (eg with a Letsencrypt certificate, good ciphers and all that) - it can also limit access based on other needs - e.g. source IP addresses.

Point is.. while this can be added to Krill I would like to understand the urgency given the proxy option.

timbru:
As far as I can tell, but I might be mistaken.. the `hyper` server used by Krill can only bind to one address.

I have made an issue asking for a feature or clarification:
https://github.com/hyperium/hyper/issues/3009

This said, I still think that letting Krill listen on IPv4 localhost only and using a proxy for any public IP addresses is best practice.

ties:
I do agree that a reserve proxy is the preferred setup. However in some situations it may be good to listen on either multiple addresses or dual stack.

PR #958, opened 2022-10-14
Support listening on multiple IP addresses. (#955)

timbru:
Allows optional configuration of multiple `ips` but will not change default behaviour and will still accept `ip` if it was set.

This will allow Krill to listen on both IPv4 and IPv6, of multiple of either of course..

Internally the code was reworked a bit so that the futures do not return an error anymore.. they just log an error and come back. This was needed to make the wiring a bit more manageable.

### DIFF
--- a/defaults/krill-pubd.conf
+++ b/defaults/krill-pubd.conf
@@ -91,7 +91,7 @@
 # or Apache if you must make your Krill instance accessible remotely.
 #
 # You can use the 'ip' setting in this config file to override the default. You
-# can specify a singly IP address or an array of addresses.
+# can specify a single IP address or an array of addresses.
 #
 # If you want to support remote delegated CAs to be children under a CA and/or
 # publish their content, then you should set the "service uri" setting described
--- a/defaults/krill-pubd.conf
+++ b/defaults/krill-pubd.conf
@@ -91,16 +91,16 @@
 # or Apache if you must make your Krill instance accessible remotely.
 #
 # You can use the 'ip' setting in this config file to override the default. You
-# can also specify multiple IP addresses if you need to. In this case 'ips' needs
-# to be set instead of 'ip'. If you set both, then 'ips' takes precedence.
+# can specify a singly IP address or an array of addresses.
 #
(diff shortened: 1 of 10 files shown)


defaults/krill-pubd.conf
```diff
@@ -84,13 +84,24 @@
...
+# can also specify multiple IP addresses if you need to. In this case 'ips' needs
+# to be set instead of 'ip'. If you set both, then 'ips' takes precedence.
+#
+# If you do change the IP address(es), then the service uri will use it (or the
+# first in case of multiple) as a hostname for remote CAs and Publishers. If you
+# want to use a proper public hostname, please set 'service_uri' below.
+#
+#
+### ip             = "127.0.0.1"            # default
+### ips            = [ "127.0.0.1", "::1" ] # overrides 'ip'
```
ximon18, line 103: FYI serde_as OneOrMany from the serde_with crate can be used to accept both single string or toml array of string inputs under the same toml key, avoiding the need for two options for the same thing.

timbru: As most users won't need this I prefer that we do not introduce yet another dependency.

timbru: alright.. `serde_with` only depends on `serde` and the rest is optional - so I guess that including it is low risk.. will look.

defaults/krill-pubd.conf
```diff
@@ -84,13 +84,24 @@
...
-# Note: we recommend that you use the defaults and use a proxy if you
-# must make your Krill instance accessible remotely.
+# Note: by default Krill uses "127.0.0.1" (IPv4 localhost) as its IP address.
+# We recommend that you keep this setting and use a proxy server such as NGINX
+# or Apache if you must make your Krill instance accessible remotely.
 #
-### ip             = "localhost"
-### port           = 3000
+# You can use the 'ip' setting in this config file to override the default. You
+# can specify a singly IP address or an array of addresses.
```
ximon18, line 94: s/singly/single/

ximon18 (approved):

PR #966, opened 2022-10-26
Update rpki-rs to accept id-cert with path len constraints.

timbru:
DRAFT: Requires release of rpki-rs.


timbru:
@partim This depended on a branch for rpki-rs 0.15.9-dev but I believe that is merged. Can we have a 0.15.9 release soon-ish?

timbru:
dropping this branch and PR rather than resolving the conflicts - this really only needs an update of the rpki-rs dependency which can be done straight on the prep-0.13.0 branch.

PR #972, opened 2022-11-17
Follow move of reusable pkg workflow to new ploutos repo.

ximon18:
Rather than pollute the org wide central repo with the reusable packaging workflow, it has been moved to its own home in a new ploutos repo. This change follows that move.

### DIFF
--- a/.github/workflows/pkg.yml
+++ b/.github/workflows/pkg.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   package:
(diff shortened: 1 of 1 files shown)


timbru (approved):
I have not reviewed the code at the new repository location, but I trust that it's the same or better..

ximon18:
> I have not reviewed the code at the new repository location, but I trust that it's the same or better..

@timbru: It’s  the same. There is a newer version but it’s backwards incompatible so requires explicit opt-in by bumping the requested version from @v1 to @v4 which will be done as part of a separate PR.

PR #973, opened 2022-11-22
Add cas in bulk

timbru:
Adds an API to support importing CAs in bulk. There is support in the client library (also used for testing), but there is no option added to the CLI. I believe that we will not need this and it might confuse users if the option is added.

The TA related structure (aia and uri) may change a bit when we work on supporting offline TAs and importing existing keys after this.

The existing testbed and benchmarking code has been modified so that it uses this import process as well.


src/cli/client.rs
```diff
@@ -171,6 +171,10 @@ impl KrillClient {
                 post_empty(&self.server, &self.token, "api/v1/bulk/cas/suspend").await?;
                 Ok(ApiResponse::Empty)
             }
+            BulkCaCommand::Import(structure) => {
+                post_json(&self.server, &self.token, "api/v1/bulk/cas/import", structure).await?;
+                Ok(ApiResponse::Empty)
```
ximon18, line 176: Minor observation: as all branches end with `Ok(ApiResponse::Empty)` you could factor that out and as a bonus make it clearer at a glance that all bulk operations yield an empty response.

src/commons/api/import.rs
```diff
@@ -0,0 +1,147 @@
...
+    // the resources for each child CA.
+    pub fn validate_ca_hierarchy(&self) -> Result<(), String> {
+        let mut seen: HashMap<ParentHandle, ResourceSet> = HashMap::new();
+        // ta is implied
+        seen.insert(ta_handle().into_converted(), ResourceSet::all());
+        for ca in &self.cas {
+            let mut ca_resources = ResourceSet::empty();
+            for parent in &ca.parents {
+                if let Some(seen_parent_resources) = seen.get(parent.handle()) {
+                    if seen_parent_resources.contains(&parent.resources) {
```
ximon18, line 55: Shouldn't this check if the parent contains the resources claimed by the CA? It seems to check that the parent contains the parents resources which should always be true...

src/test.rs
```diff
@@ -450,6 +450,55 @@ pub async fn ca_configured_roas(ca: &CaHandle) -> ConfiguredRoas {
...
+    let expected: Vec<_> = expected
+        .iter()
+        .map(|entry| entry.clone().into_explicit_max_length())
+        .collect();
+
+    for configuration in configured_roas.iter().map(|configured| configured.roa_configuration()) {
+        if !expected.contains(configuration) {
+            use std::fmt::Write;
+
+            let mut expected_str = String::new();
```
ximon18, line 469: This seems like a lot of effort to do a join of a collection by commas into a string, isn't there a one-liner for this?

tests/functional_ca_import.rs
```diff
@@ -0,0 +1,116 @@
...
+        let mut expected_files_grandchild = expected_mft_and_crl(&grandchild, &rcn_0).await;
+        expected_files_grandchild.append(&mut expected_mft_and_crl(&grandchild, &rcn_1).await);
+
+        for roa in &grandchild_roas {
+            expected_files_grandchild.push(ObjectName::from(&roa.payload().into_explicit_max_length()).to_string());
+        }
+
+        assert!(
+            will_publish_embedded(
+                "child1 should publish certificate for grandchild and 3 roas",
```
ximon18, line 107: 2 ROAs, no?

tests/functional_ca_import.rs
```diff
@@ -0,0 +1,116 @@
...
+            )
+            .await
+        );
+    }
+
+    {
+        //check child2 exists and has resources
+        assert!(ca_contains_resources(&child2, &child2_resources).await);
+    }
+
```
ximon18, line 88: Should you also check child2 as child1 was checked?

src/daemon/krillserver.rs
```diff
@@ -294,8 +171,78 @@ impl KrillServer {
...
+                }
+
+                let startup_structure = api::import::Structure::new(
+                    testbed.ta_aia().clone(),
+                    testbed.ta_uri().clone(),
+                    testbed.publication_server_uris(),
+                    import_cas,
+                );
+                server.cas_import(startup_structure).await?;
+            } else if config.benchmark.is_some() {
```
ximon18, line 238: I'd find it more readable if these two short if blocks came before the longer block.

src/cli/client.rs
timbru: sure. Done

src/test.rs
timbru: yeah.. fixed it using a join call instead.

src/commons/api/import.rs
timbru: I can see how this was confusing.

That "parent" is the parent as the CA sees it. But, I added some comments and renamed "parent" to "ca_parent". Hopefully that makes it more clear what is going on.

tests/functional_ca_import.rs
timbru: ah yes, and "grandchild" not "child1". Copy/paste.. now fixed

tests/functional_ca_import.rs
timbru: Yeah sure.  I kept it short as it's implied by the checks on the grandchild, but I have added it now for completeness.

ximon18 (approved):

src/daemon/krillserver.rs
timbru: improved (I think)

PR #975, opened 2022-11-29
Use handle locks in WalStore.

timbru:
This change will ensure that locks are used for ``WalSupport``  types (write-ahead log) managed through the ``WalStore`` in the same way as we have been doing for ``Aggregate`` types managed trough the ``AggregateStore``.

This fixes an issue with concurrent creation of CAs as part of the (unreleased) ca_imports function.

### DIFF
--- a/src/commons/eventsourcing/locks.rs
+++ b/src/commons/eventsourcing/locks.rs
@@ -79,9 +79,9 @@ impl HandleLocks {
             }
         }
 
-        // Entry exists now, so return the lock
-        let map = self.locks.read().unwrap();
-        HandleLock { map, handle }
+        // Entry probably exists now, but recurse in case the entry
+        // was dropped immediately after creation.
+        self.for_handle(handle)
     }
 
     pub fn drop_handle(&self, handle: &MyHandle) {
--- a/src/commons/eventsourcing/locks.rs
+++ b/src/commons/eventsourcing/locks.rs
@@ -1,0 +1,91 @@
+//! Support locking on Handles so that updates can be
+//! performed sequentially. Useful for both event sourced
+//! types (Aggregates) as well as write-ahead logging
+//! types.
+
+//------------ HandleLocks ---------------------------------------------------
+
+use std::{
+    collections::HashMap,
+    sync::{RwLock, RwLockReadGuard, RwLockWriteGuard},
+};
+
+use rpki::ca::idexchange::MyHandle;
+
+#[derive(Debug, Default)]
+struct HandleLockMap(HashMap<MyHandle, RwLock<()>>);
+
+impl HandleLockMap {
+    fn create_handle_lock(&mut self, handle: MyHandle) {
+        self.0.insert(handle, RwLock::new(()));
+    }
+
+    fn has_handle(&self, handle: &MyHandle) -> bool {
+        self.0.contains_key(handle)
+    }
+
+    fn drop_handle_lock(&mut self, handle: &MyHandle) {
+        self.0.remove(handle);
+    }
+}
+
+pub struct HandleLock<'a> {
+    // Needs a read reference to the map that holds the RwLock
+    // for the handle.
+    map: RwLockReadGuard<'a, HandleLockMap>,
+    handle: MyHandle,
+}
+
+impl HandleLock<'_> {
+    // panics if there is no entry for the handle.
+    pub fn read(&self) -> RwLockReadGuard<'_, ()> {
+        self.map.0.get(&self.handle).unwrap().read().unwrap()
+    }
+
+    // panics if there is no entry for the handle.
+    pub fn write(&self) -> RwLockWriteGuard<'_, ()> {
+        self.map.0.get(&self.handle).unwrap().write().unwrap()
+    }
+}
+
+/// This structure is used to ensure that we have unique access to an instance for a [`Handle`]
+/// managed in an [`AggregateStore`] or [`WalStore`]. Currently uses a `std::sync::RwLock`, but
+/// this should be improved to use an async lock instead (e.g. `tokio::sync::RwLock`).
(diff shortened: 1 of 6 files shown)


src/commons/eventsourcing/locks.rs
```diff
@@ -0,0 +1,91 @@
...
+                map.create_handle_lock(handle.clone());
+            }
+        }
+
+        // Entry exists now, so return the lock
+        let map = self.locks.read().unwrap();
+        HandleLock { map, handle }
+    }
+
+    pub fn drop_handle(&self, handle: &MyHandle) {
```
ximon18, line 87: In theory is it possible that thread A calls `for_handle()` and creates or finds an existing lock, but then between lines 80 and 83 that thread B calls `drop_handle()` and removes the lock that is thought to exist by line 83?

ximon18: Note I just saw this code is not new but just moved so I guess we've been over this before and you can ignore me, though am curious about the answer.

src/commons/eventsourcing/wal.rs
```diff
@@ -209,8 +225,17 @@ impl<T: WalSupport> WalStore<T> {
...
+            {
+                // First get a lock and remove the object
+                let handle_lock = self.locks.for_handle(handle.clone());
+                let _write = handle_lock.write();
+                self.cache.write().unwrap().remove(handle);
+                self.kv.drop_scope(handle.as_str())?;
+            }
+
+            // Then drop the lock for it as well. We could not do this
+            // while holding the write lock. Note that
```
ximon18, line 237: Incomplete sentence.

src/commons/eventsourcing/wal.rs
```diff
@@ -120,6 +120,7 @@ impl<T: WalSupport> WalSet<T> {
 pub struct WalStore<T: WalSupport> {
     kv: KeyValueStore,
     cache: RwLock<HashMap<MyHandle, Arc<T>>>,
+    locks: HandleLocks,
```
ximon18, line 123: Commenting here since https://github.com/github/roadmap/issues/347 is still open.
Is the purpose of the lock to protect access to `kv`? If so, shouldn't the lock be acquired in more functions that use `kv`, I see some functions do not acquire the lock, e.g. `has()`,  `get_snapshot()` and `list()`?

src/commons/eventsourcing/wal.rs
```diff
@@ -285,6 +314,11 @@ impl<T: WalSupport> WalStore<T> {
     /// This is a separate function because serializing a large instance can
     /// be expensive.
     pub fn update_snapshot(&self, handle: &MyHandle, archive: bool) -> WalStoreResult<()> {
+        // Note: We do not need to keep a read lock for the instance when we update the snapshot.
```
ximon18, line 317: I don't understand how it is okay not to lock here.

src/test.rs
```diff
@@ -725,6 +725,22 @@ pub async fn wait_for_nr_cas_under_testbed(nr: usize) -> bool {
     false
 }
 
+pub async fn wait_for_nr_cas_under_publication_server(publishers_expected: usize) {
+    let mut publishers_found = list_publishers().await.publishers().len();
+    for _ in 0..300 {
+        if publishers_expected == publishers_found {
```
ximon18, line 731: Personally I'd swap the order of these comparison arguments as it reads to me as if `publishers_expected` should change to match `publishers_found`.

src/commons/eventsourcing/locks.rs
timbru: It is an unlikely race condition under normal use, because it implies that a 'thing' (Aggregate or WalSupport type) was removed, and then updated. But it not strictly impossible.

We can use recursion instead. That will work for now since the lock is not (yet) `async`.

src/commons/eventsourcing/wal.rs
timbru: No, the KV store is used for different entities identified by their own handles. So, we specifically want to lock operations on a specific handle only. E.g. otherwise CA1 would have to wait for CA2's ROA changes to complete etc. This is even worse when we have an embedded parent CA that can be busy.

Operations on different handles can never conflict, so I did not want the global lock.

src/commons/eventsourcing/wal.rs
timbru: Will improve the comment then.. but essentially..

This updates the latest snapshot file on disk (or well in the KV store). It is okay to do this with a copy of the data, even if that copy becomes outdated as the snapshot is written. All changes are stored in their own write-ahead-log (wal) file. So, even if we end up saving a now outdated snapshot, we will have all changes after it.

If we would keep the lock, then this would mean that write operations would have to wait until the full snapshot file is written. For a large repository (1000s of CAs and ROAs) this can take a while (up to 30 seconds on my fast machine for a 250 MB repo) because the serialisation is expensive. I think it's better not to block the system for this long if it's not needed.

ximon18 (approved):

src/commons/eventsourcing/wal.rs
timbru: Ah yes, oops.. I remember now that as I was typing a comment that was going to say that a race condition would be unlikely I figured that perhaps.. we should have a `self.locks.remove_handle` instead. Let me take a stab at that in a moment..

src/test.rs
timbru: fixed

src/commons/eventsourcing/wal.rs
timbru: updated the comment

src/commons/eventsourcing/wal.rs
timbru: updated comment

PR #982, opened 2022-12-08
Check uri not hash for publish elements. Fixes #981

### LINKED ISSUE #981: Publication Server should check uri, not hash, in publish elements
Publish elements should use a unique uri. Until now Krill servers have been checking that the hash is unique instead. This is clearly wrong.

This has not been an issue in actual Krill deployments because Krill CAs will always perform a list query first and then send an update. I.e. they would never send publish elements for existing uris - they would send an update element (publish + hash attribute) instead.

This bug can cause interop issues with CAs that simply send updates to the server without checking the content of the server first.

timbru:


src/commons/api/rrdp.rs
```diff
@@ -731,3 +734,79 @@ impl DeltaData {
...
 }
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+    use crate::test::*;
+
+    #[test]
+    fn current_objects_delta() {
```
ximon18, line 745: Where does this test verify the correct behaviour for the scenario described in #981? I would expect to see a test here that attempts to publish an object with URI A and hash X and then to publish an object with URI A and hash Y, i.e. verify that a client that _"sends a publish element for an existing uri" (ed: slightly paraphrased from what was said on #981) will not succeed. I assume that the `fileN_content` variables contain hashes rather than file content, and thus that I should be looking for a test that for example publishes `file1_uri` with `file1_content` and then attempts to publish `file1_uri` with `file2_content`, but I only see a single publish followed by an identical publish rather than a same-by-uri-but-not-hash publish, and then after that only updates and withdraws.

ximon18 (approved):

PR #983, opened 2022-12-09
Purge content by uri

timbru:
[draft] Merge after #PR982

This adds functionality to purge (well, maybe it should be called remove? zap? obliterate?) content from the publication server repository by URI. This will help in case a repository got into a bad state because of issue #981. But it may also help in certain other cases.

To use this from the command line:
```
krillc help pubserver purge
krillc-pubserver-purge 
Delete specific files from the Publication Server

USAGE:
    krillc pubserver purge [FLAGS] [OPTIONS] --base_uri <rsync URI>

FLAGS:
        --api        Only show the API call and exit. Or set env: KRILL_CLI_API=1
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -u, --base_uri <rsync URI>    Remove file matching the base uri if it ends with '/', or the exact uri if it refers
                                  to a file.
    -f, --format <type>           Report format: none|json|text (default). Or set env: KRILL_CLI_FORMAT
    -s, --server <URI>            The full URI to the Krill server. Or set env: KRILL_CLI_SERVER
    -t, --token <string>          The secret token for the Krill server. Or set env: KRILL_CLI_TOKEN
```

### DIFF
--- a/src/commons/api/rrdp.rs
+++ b/src/commons/api/rrdp.rs
@@ -318,7 +318,7 @@ impl CurrentObjects {
 
     /// Returns a copy of self where elements matching the given URI
     /// are removed if there are any matches. Otherwise, returns None.
-    pub fn purge(&self, uri: &uri::Rsync) -> Option<Self> {
+    pub fn with_matching_uri_deleted(&self, uri: &uri::Rsync) -> Option<Self> {
         let mut withdraws = vec![];
 
         // We first loop through the elements to avoid having to clone in case there is no work
--- a/src/commons/api/rrdp.rs
+++ b/src/commons/api/rrdp.rs
@@ -316,6 +316,29 @@ impl CurrentObjects {
         }
     }
 
+    /// Returns a copy of self where elements matching the given URI
+    /// are removed if there are any matches. Otherwise, returns None.
+    pub fn purge(&self, uri: &uri::Rsync) -> Option<Self> {
+        let mut withdraws = vec![];
+
+        // We first loop through the elements to avoid having to clone in case there is no work
+        for (hash, el) in &self.0 {
+            if el.uri() == uri || (uri.as_str().ends_with('/') && el.uri().as_str().starts_with(uri.as_str())) {
+                withdraws.push(hash)
+            }
+        }
+
+        if withdraws.is_empty() {
+            None
+        } else {
+            let mut copy_of_self = self.clone();
+            for hash in withdraws {
+                copy_of_self.0.remove(hash);
+            }
+            Some(copy_of_self)
+        }
+    }
+
     pub fn len(&self) -> usize {
         self.0.len()
     }
--- a/src/commons/api/rrdp.rs
+++ b/src/commons/api/rrdp.rs
@@ -208,6 +208,10 @@ pub struct WithdrawElement {
 }
 
 impl WithdrawElement {
+    pub fn new(uri: uri::Rsync, hash: Hash) -> Self {
+        WithdrawElement { uri, hash }
+    }
+
     pub fn uri(&self) -> &uri::Rsync {
         &self.uri
     }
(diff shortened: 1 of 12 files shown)


ximon18:
> This adds functionality to purge (well, maybe it should be called remove? zap? obliterate?)

Delete?

timbru:
> > This adds functionality to purge (well, maybe it should be called remove? zap? obliterate?)
> 
> Delete? Unpublish?

Yeah, I guess 'delete' is the most neutral. Purge has connotations - which may be applicable in this context, but then again they may not and it's best to use neutral language. Will change later when I have a chance to get back to this.

src/commons/api/rrdp.rs
```diff
@@ -313,6 +316,29 @@ impl CurrentObjects {
...
+            if el.uri() == uri || (uri.as_str().ends_with('/') && el.uri().as_str().starts_with(uri.as_str())) {
+                withdraws.push(hash)
+            }
+        }
+
+        if withdraws.is_empty() {
+            None
+        } else {
+            let mut copy_of_self = self.clone();
+            for hash in withdraws {
```
ximon18, line 335: You could use [`retain()`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.retain) here.

src/commons/api/rrdp.rs
```diff
@@ -313,6 +316,29 @@ impl CurrentObjects {
...
+        for (hash, el) in &self.0 {
+            if el.uri() == uri || (uri.as_str().ends_with('/') && el.uri().as_str().starts_with(uri.as_str())) {
+                withdraws.push(hash)
+            }
+        }
+
+        if withdraws.is_empty() {
+            None
+        } else {
+            let mut copy_of_self = self.clone();
```
ximon18, line 334: You could also construct the `HashMap` using `FromIterator` on only those items matching a filter, instead of clone then remove.

src/commons/api/rrdp.rs
```diff
@@ -313,6 +316,29 @@ impl CurrentObjects {
...
     }
 
+    /// Returns a copy of self where elements matching the given URI
+    /// are removed if there are any matches. Otherwise, returns None.
+    pub fn purge(&self, uri: &uri::Rsync) -> Option<Self> {
+        let mut withdraws = vec![];
+
+        // We first loop through the elements to avoid having to clone in case there is no work
+        for (hash, el) in &self.0 {
+            if el.uri() == uri || (uri.as_str().ends_with('/') && el.uri().as_str().starts_with(uri.as_str())) {
```
ximon18, line 326: This same check is also done by the caller `purge_files()`.... maybe this fn should just purge an exact set that it's given? Or the caller should use the result of this call rather than work out the matching set itself as well?

src/commons/api/rrdp.rs
```diff
@@ -313,6 +316,29 @@ impl CurrentObjects {
         }
     }
 
+    /// Returns a copy of self where elements matching the given URI
+    /// are removed if there are any matches. Otherwise, returns None.
+    pub fn purge(&self, uri: &uri::Rsync) -> Option<Self> {
```
ximon18, line 321: This fn doesn't actually purge anything, it doesn't even change `self`... it could even always return a collection in place of Option and just be empty if there is nothing to "purge", maybe it should be called find matching or find purgeable or ...

ximon18 (approved):

src/commons/api/rrdp.rs
timbru: yeah, but then we get a nested loop - i.e. O-square (well N el *M filter) rather than O-N (I presume) for clone, and O log M I presume for remove. Mostly we will be removing only a a fraction - so I think the allocation and then removal should be fine.

src/commons/api/rrdp.rs
timbru: The context is different. First we remove the matching files from the repo content.  But.. the objects are *also* kept on a per publisher basis and we will need to delete them there to. Presumably most publishers will have nothing to delete. So this is done this way to avoid cloning unless it's really needed - i.e. we found a publisher for which we need to do work: make an event with a copy of the (to be) deleted objects.

src/commons/api/rrdp.rs
timbru: ack - changed it to `with_matching_uri_deleted`

src/commons/api/rrdp.rs
timbru: I think this would be O-N-square (N el * M hashes) vs O N (el) * O log M (hashes).

PR #985, opened 2022-12-22
Support offline ta

timbru:
This adds better TA support. The TA functions are split into a 'Proxy' and a 'Signer'. In test setups both can be active in a Krill instance. The signer is intended to be operated offline in production environments.

We introduce a new binary `krillta` to manage both the proxy and the signer.

This will connect to the API of Krill, like `krillc`, when this tool is used to manage the proxy.
It will require a local directory and configuration file when use to manage the signer.

Documentation can be found here for now (branch will be merged on release):
https://krill.docs.nlnetlabs.nl/en/ta-support/trust-anchor.html


timbru:
The code is in a working state, but I will work on user documentation (on a branch of the krill docs). That may be helpful to review this.. stay tuned.

timbru:
Merged into the `prep-0.13.0` branch. We will make an RC release based on that branch and do a full code review before merging it into main and making the final 0.13.0 release.

PR #998, opened 2023-01-19
Clean up old TA code. Refuse to upgrade if there is an old ta.

timbru:
It's quite hard to do the migration and there are only a handful of test setups expected to use this.. so no migration.

Migration would entail a rather complicated re-writing of history - we cannot simply convert the old TA to a new style TA signer and proxy. In particular with regards to the interaction with their child and the publication server.


timbru:
Merged this into `support-offline-ta`, where it will still be reviewed as part of PR #985

PR #999, opened 2023-01-20
Sign ta proxy-signer requests and responses

timbru:


timbru:
This will be reviewed as part of PR #985

PR #1001, opened 2023-01-23
Upgrade Ploutos packaging workflow from v1 to v5

ximon18:
Changes:
- Now specifies a `cross_max_wait_mins`, needed in newer Ploutos to allow the parallel cross-compilation jobs to take longer than the default maximum of 10 minutes.
- `_path` extensions on input names were removed in newer Ploutos.
- `rpm_extra_build_packages` must manually specify `make` as a dependency because installation of the RPM `Development Tools` package group as a default behaviour was removed in newer Ploutos (to speed up builds that don't need it, i.e. Pure Rust builds).
- `deb_maintainer` was removed as it is now determined from `Cargo.toml`. You can see that it is still set correctly by looking at the action log outputs particularly the Verify step, e.g. [here](https://github.com/NLnetLabs/krill/actions/runs/3987414689/jobs/6837208879#step:24:148).
- `package_test_rules_path` was removed because newer Ploutos can re-use the settings used for building if they are the same as used for testing, and in Krill the only differences we used to have were to skip upgrade testing on certain platforms but this difference was outdated because we now have prior releases on those platforms that we _can_ do upgrade testing against.

Optional still to do:
- Make use of the [new RPM macro replacement functionality](https://github.com/NLnetLabs/ploutos/blob/main/docs/os_packaging.md#maintainer-scriptlets), e.g. [like Routinator does](https://github.com/NLnetLabs/routinator/blob/main/pkg/rpm/scriptlets.toml).

Observations:
- Runtime is reduced compared to past runs with Ploutos v1 where the [vast majority of successful runs exceeded 30 minutes](https://github.com/NLnetLabs/krill/actions/workflows/pkg.yml?query=is%3Asuccess) and the [test run of this PR](https://github.com/NLnetLabs/krill/actions/runs/3987414689) took 20 minutes.


timbru:
@ximon18 should this target `prep-0.13.0` instead of main?

pkg/rules/packages-to-test.yml
```diff
@@ -1,34 +0,0 @@
----
```
timbru, line 1: So, I presume testing is now done in a different way?

timbru (approved):

pkg/rules/packages-to-test.yml
ximon18: No, it's just that that file was a complete duplicate of the build instructions, newer Ploutos doesn't require you to manually specify the same stuff twice if not needed.

timbru: should have read your "Changes" above better.. sorry

PR #1007, opened 2023-02-01
Remove RepositoryManager RwLock

timbru:
[draft] We should make a branch for `prep-0.12.2` and merge to that if this is stable.

This PR removes the RwLock in RepositoryManager.

We saw issues where a busy Krill Publication Server would freeze without any reported issues. We suspect that this may be due to a deadlock or lock starvation issue in RepositoryManager. A version where read locks were no longer acquired (they are actually not needed) - was deployed and so far seems stable.

Then one more commit was done to remove the RwLock altogether in this layer. It is in fact not needed because there are existing RwLocks in the layers below this (`WalStore::send_command`) to ensure that actual updates do happen sequentially.

### DIFF
--- a/src/pubd/manager.rs
+++ b/src/pubd/manager.rs
@@ -211,8 +211,6 @@ impl RepositoryManager {
 
     /// Returns a list reply for a known publisher in a repository.
     pub fn list(&self, publisher: &PublisherHandle) -> KrillResult<ListReply> {
-        let _lock = self.default_repo_lock.read().unwrap();
-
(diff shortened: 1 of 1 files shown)


Koenvh1:
I can understand why the previous deadlock happened, but I am not certain something similar could not still happen with the current locks in the `WalStore`. Let me see if I can create a test for that somehow.

Koenvh1 (approved):
As far as I can see this should not cause any issues

PR #1008, opened 2023-02-03
Allow including TA private key pem for CA import.

timbru:


src/daemon/krillserver.rs
```diff
@@ -576,47 +577,37 @@ impl KrillServer {
...
-            }
-
-            if let Some(import_ta) = structure.ta.clone() {
-                info!("Creating embedded Trust Anchor");
-                let (ta_aia, ta_uris) = import_ta.into_uris();
-                self.ca_manager
-                    .ta_init_fully_embedded(ta_aia, ta_uris, &self.repo_manager, &actor)
-                    .await?;
-            }
+        if let Some(publication_server_uris) = structure.publication_server.clone() {
```
ximon18, line 580: Why did you remove these "enabled" checks?

timbru: apologies for the late response - got side tracked - they are removed because technically we could just import local CAs and not set up a TA. That said, the checks should be done when a TA is imported - so I will add them there now.

PR #1020, opened 2023-02-28
Per best practice don't permit PKCS#11 private keys to be modified (fixes #1018).

### LINKED ISSUE #1018: Mark PKCS#11 private keys as unmodifiable
Per best practice, e.g. from http://secgroup.dais.unive.it/wp-content/uploads/2010/10/Reponse-by-SafeNet.pdf:

> 3. **Follow best practices when setting sensitive key attributes:** If you have secret or private keys
that are particularly sensitive and you want to prevent them from being wrapped off, they can be
generated with their template attributes: CKA_SENSITIVE and CKA_PRIVATE set to True and
CKA_EXTRACTABLE and CKA_MODIFIABLE both set to False. This way, the keys are only
accessible by a user who is logged in, and key values cannot be read by anyone. Also, the keys
cannot be wrapped off and the attribute values cannot be changed at some later time to invalidate
the original settings

Currently when Krill creates an RSA key pair using the PKCS#11 `C_GenerateKeyPair()` function it sets key attributes `CKA_SENSITIVE` and `CKA_PRIVATE` to `CK_TRUE` :heavy_check_mark:  and `CKA_EXTRACTABLE` to `CK_FALSE` :heavy_check_mark: as per the guidance above, but leaves `CKA_MODIFIABLE` to the default value (which is `CK_TRUE` according to [section 4 of the PKCS#11 v2.40 specification](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc416959705)) :x:.

ximon18:
_(This description has been updated following re-work of the PR as described in the comments below)_

This PR adds `krill.conf` configuration settings that give operators control (per #1018) over some of the PKCS#11 attributes set on new public and private keys that Krill creates, rather than assume in Krill that we know what the best defaults / hard-coded values are.

The existing default behaviour is preserved for backward compatibility, i.e. these new settings do nothing if not used.


timbru (approved):

ximon18:
@timbru: I think these new settings also need mentioning in the HSM part of the Krill manual, but I haven't done that yet.

PR #1021, opened 2023-02-28
Permit setting CKA_PRIVATE to CK_FALSE on PKCS#11 RSA public keys (#1019)

### LINKED ISSUE #1019: Permit setting CKA_PRIVATE to CK_FALSE on PKCS#11 RSA public keys
Currently when creating an RSA key pair using the PKCS#11 `C_GenerateKeyPair()` function Krill sets `CKA_PRIVATE` to `CK_FALSE` for both the public and private keys to be generated.

Per [section 4.4 of the PKCS#11 v2.40 specification](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc416959705):
> When the CKA_PRIVATE attribute is CK_TRUE, a user may not access the object until the user has been authenticated to the token.

This seemed like a good value to use. While one could argue that the public key itself is not sensitive, why allow it to be accessed by unauthenticated users of the PKCS#11 interface if authenticated access is anyway required to access the private key (because `CKA_PRIVATE` is set to `CK_TRUE` for the private key) ?

However, there are at least a couple of references to situations in which some HSMs require the public key to have `CKA_PRIVATE` set to `CK_FALSE`:
- https://nshielddocs.entrust.com/api-generic/12.80/pkcs11#functions-supported _("You must set CKA_PRIVATE to FALSE for public keys")_.
- https://jira.hyperledger.org/browse/FAB-5407
- https://jira.hyperledger.org/browse/FAB-17280

Note though that it may not be safe to always set `CKA_PRIVATE` to `CK_FALSE`, e.g. the AWS CloudHSM documentation says in a couple of places that it must be set to `CK_TRUE`:
- https://docs.aws.amazon.com/cloudhsm/latest/userguide/pkcs11-attributes.html#f8 _("This attribute **(ed: CKA_PRIVATE for an RSA public key when calling C_GenerateKeyPair)** is partially supported by the firmware and must be explicitly set only to the default value **(ed: where the default value is listed in the same page as True)**")_
- https://s3.amazonaws.com/cloudhsmv2-software/CloudHsmClient/Docs/PKCS11ComplianceReportSDK3-2-1.pdf _("Objects with CKA_PRIVATE set to CK_FALSE – meaning objects intended to be accessed without authenticating to the HSM – cannot be stored on the HSM")_

We should introduce a way to configure Krill to omit defining the `CKA_PRIVATE` attribute in the public key template when invoking the PKCS#11 function `C_GenerateKeyPair()` to support more HSMs and/or more HSM configurations.

ximon18:
Fixes #1019.

Introduces a new optional PKCS#11 setting in the `[[signers]]` section of `krill.conf` called `pubkey_access`.

The new setting controls the value of the `CKA_PRIVATE` attribute included in the public key generation template passed to the PKCS#11 `C_GenerateKeyPair()` function.

The setting can be set as follows:

| `pubkey_access` | PKCS#11 `CKA_PRIVATE` equivalent |
|----|---|
| `authenticated` (default) | `CK_TRUE` |
| `unauthenticated` | `CK_FALSE` |
| `token-default` | Not specified |

The setting value names are based on the description of `CKA_PRIVATE` in the PKCS#11 v2.40 specification in [section 4.4 Storage Objects](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc416959705).

The default corresponds to the existing behaviour of previous Krill releases.


ties:
I think that retrying key generation when it fails with the default value (= unset) with `CKA_PRIVATE=true` is probably the best solution.

The additional parameter is confusing and formalises what according to AWS their docs is an incompatibility in the CloudHSM API, as default behaviour.

ximon18:
I decided to investigate further what it would take to do the magic fallback approach instead, i.e. on CKR_INCONSISTENT_TEMPLATE retry omitting the CKA_PRIVATE attribute in the public key generation template.

One problem I hadn't foreseen with this approach is that the current signer registration model is signer implementation independent, there's no way to store a specific flag for a PKCS#11 signer as the same common fields are stored for OpenSSL, KMIP and PKCS#11 signers. I could force the flag into a string type, maybe even force it into the descriptive strings held for each signer, but strong typing is much preferred over basing behaviour on string content.

Of course it could be done, but it would be even more work than I thought. For example we would also have to support existing signer registration details stored on disk that lack the extra field and assume that the absence of the field means that CKA_PRIVATE should be set to CK_TRUE, or "upgrade" the signer registration store on first start of the newer version of Krill to include the missing field.

I can see the attraction of the automatic fallback, hence why I proposed it as an option in the first place, but I'm still wary of the magic quality of this behaviour vs the explicit nature of a config setting.

ximon18:
One improvement I could make to the current implementation is on CKR_INCONSISTENT_TEMPLATE error when probing a signer, to log a helpful error such as _"One possible cause for this error is that the signer doesn't support authenticated pubkey_access. If this is the case, setting pubkey_access to token-default or unauthenticated in your Krill configuration file may be enough to make the signer usable by Krill."_.

ties:
> One improvement I could make to the current implementation is on CKR_INCONSISTENT_TEMPLATE error when probing a signer, to log a helpful error such as _"One possible cause for this error is that the signer doesn't support authenticated pubkey_access. If this is the case, setting pubkey_access to token-default or unauthenticated in your Krill configuration file may be enough to make the signer usable by Krill."_.

This sounds like the right fix.

timbru (approved):

ximon18:
> > One improvement I could make to the current implementation is on CKR_INCONSISTENT_TEMPLATE error when probing a signer, to log a helpful error such as _"One possible cause for this error is that the signer doesn't support authenticated pubkey_access. If this is the case, setting pubkey_access to token-default or unauthenticated in your Krill configuration file may be enough to make the signer usable by Krill."_.
> 
> This sounds like the right fix.

I have pushed a commit to this PR to add this error message. It was more work than I expected because I discovered that (a) the underlying error was not accessible to higher level logic even within the context of the PKCS#11 signer, and that once augmented the modified error message was being wrongly dropped and never included in the logs. I also made some warnings about signer registration/detection failure into errors to be consistent with other nearby logged messages which were already errors, and because if you configure a signer it is a pretty serious error if it cannot be used.

ximon18:
@tim: It might be good if you look at this again given my last commit and comment above.

timbru (approved):

ximon18:
Note that the config setting added in this PR was replaced by different settings added in PR #1020.

PR #1023, opened 2023-03-17
Remove published objects duplication

timbru:
The RepositoryContent component in Krill was keeping separate lists of published objects. One list for all the objects as they are published in RRDP and rsync. And one list per publisher to keep track of the current files for each publisher. This was done because.. well.. wrong reasons.. it was mostly easy and efficient when answering RFC 8181 list queries because there was no need to look at both published, and not yet published, staged elements per publisher.

But of course, despite testing and defensive coding.. just the fact that we had duplication of the same files in separate list bears the risk that we get inconsistencies in case of unforeseen corner cases or bugs. This did not happen much (testing, and defensive coding), but has happened at least once that we know of.

In this PR we change this logic. We make sure that there is only one source of published files which is used both for RRDP and to answer RFC 8181. This way we can no longer have inconsistencies where a publishing CA *thinks* they publish the right set of files (because that is what they were told), but the reality is different.

When migrating (code still in progress) we will take the files per publisher as the leading source of truth. Most likely there are no inconsistencies between the sets, but if there are then the files per publisher reflects the publisher's intent.

Because of code complexity and effort needed we dropped support for upgrading pre 0.9.0 versions of Krill. This could in theory be done as well, but the use case is small because 0.9.0 has been around for quite some time and even if people did not upgrade to 0.9.0 yet, they can upgrade to 0.12.3 first to then upgrade to 0.13.0 and above.


src/pubd/repository.rs
```diff
@@ -764,20 +767,172 @@ pub struct RrdpServer {
...
+        for pbl in publishes {
+            let uri = pbl.uri().clone();
+            match self.0.get_mut(&uri) {
+                Some(DeltaElement::Publish(_)) => {
+                    // This should never happen as the new publish would have
+                    // been an update instead. This will have been checked
+                    // at this point. Still, rather than failing which is
+                    // not allowed in this code path, we can use the publish
+                    // instead of the staged publish.
+                    self.0.insert(uri, DeltaElement::Publish(pbl));
```
ximon18, line 830: While we cannot fail, should we at least warn or leave some sign somewhere that something unexpected happened?

src/pubd/repository.rs
```diff
@@ -764,20 +767,172 @@ pub struct RrdpServer {
...
+                }
+
+                Some(DeltaElement::Update(staged_update)) => {
+                    // This should never happen as the new publish would have
+                    // been an update instead. This will have been checked
+                    // at this point. Still, rather than failing which is
+                    // not allowed in this code path, we can update the
+                    // existing update with the content of the new publish
+                    // element.
+                    staged_update.with_updated_content(staged_update.base64().clone());
```
ximon18, line 840: This looks very strange. It looks like it says set A to A... shouldn't it use `pbl` here?

src/pubd/repository.rs
```diff
@@ -764,20 +767,172 @@ pub struct RrdpServer {
...
+                }
+                Some(DeltaElement::Withdraw(staged_withdraw)) => {
+                    // A new publish that follows a withdraw for the same URI should be
+                    // an Update of the original file.
+                    let hash = *staged_withdraw.hash();
+                    let update = UpdateElement::new(uri.clone(), hash, pbl.base64().clone());
```
ximon18, line 846: What is "hash" on UpdateElement meant to be the hash of? The thing being updated, in which case maybe this is correct? Or the update being supplied, in which case it looks a bit odd to take the hash of the previous withdraw and associate it with the new update... ?

src/pubd/repository.rs
```diff
@@ -764,20 +767,172 @@ pub struct RrdpServer {
...
+    /// Merge a new DeltaElements into this existing (unpublished) StagedElements.
+    ///
+    /// It is assumed that both sets make sense with regards to the current
+    /// files as published in the RRDP snapshot. I.e. *this* StagedElements
+    /// is assumed to be verified and can be applied to the current snapshot.
+    ///
+    /// The new StagedElements is assumed to be verified against the current
+    /// snapshot after *this* existing StagedElements has been applied.
+    ///
+    /// The reason for this merging is that while publishing CAs will always
```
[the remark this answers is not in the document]
timbru: Thanks, Rephrased. Hopefully clearer now.

src/pubd/repository.rs
timbru: added better logging in these cases

src/pubd/repository.rs
timbru: indeed. fix pushed

src/pubd/repository.rs
timbru: The hash matches the thing being updated - i.e. the state this StagedElements will be applied to, so we always need to preserve the hash when merging.

src/upgrades/mod.rs
```diff
@@ -344,7 +348,11 @@ pub fn prepare_upgrade_data_migrations(mode: UpgradeMode, config: Arc<Config>) -
         Some(versions) => {
             info!("Preparing upgrade from {} to {}", versions.from(), versions.to());
             if versions.from < KrillVersion::release(0, 6, 0) {
-                let msg = "Cannot upgrade Krill installations from before version 0.6.0. Please upgrade to any version ranging from 0.6.0 to 0.8.1 first, and then upgrade to this version.";
+                let msg = "Cannot upgrade Krill installations from before version 0.6.0. Please upgrade to 0.8.1 first, then upgrade to 0.12.3, and then upgrade to this version.";
```
ximon18, line 351: Is 0.8.1 -> 0.12.3 possible given that below you say go from 0.9.0 -> 0.12.3?

src/commons/api/rrdp.rs
```diff
@@ -457,29 +495,60 @@ impl SnapshotData {
...
+    /// any issue as deltas are verified when they are submitted.
+    pub fn apply_delta(&mut self, publisher: &PublisherHandle, delta: DeltaElements) {
+        if let Some(objects) = self.current_objects.get_mut(publisher) {
+            objects.apply_delta(delta);
+            if objects.is_empty() {
+                self.current_objects.remove(publisher);
+            }
+        } else {
+            let mut objects = CurrentObjects::default();
+            objects.apply_delta(delta); // Can only happen for new publishers
```
ximon18, line 533: What can only happen for new publishers?

ximon18 (approved):
Approved, with my limited understanding, no testing, and without trying to determine if there is other code that should also have been modified.

src/commons/api/rrdp.rs
timbru: pushed clarification

PR #1024, opened 2023-03-20
Update krill-ui + fix index.html prefix in routing

marlonbaeten:


marlonbaeten:
Please squash the commits :-)

ximon18:
Hi @marlonbaeten,

LGTM, login and logout work with KeyCloak per the [instructions here](https://krill.docs.nlnetlabs.nl/en/stable/multi-user/openid-connect-provider.html#using-keycloak) in the Krill manual, and user attributes are correctly detected and shown in the user head-and-shoulders icon popup.

Potential issues:
- Unlike the current "Lagosta" UI, and potentially confusing or even problematic for end users, I see that the built-in Krill login form appears "briefly" before the OIDC provider login form is shown. This may not even be brief if the user has a slow connection or if the OIDC provider is slow to respond, and then users may even start entering login credentials into the built-in Krill login form. Ideally this form would not be shown at all.

Still to test:
- Token refresh, that the UI correctly picks up the new token after refresh and uses it in subsequent requests to the Krill API.
- Inactive login session expiration.

src/daemon/http/auth.rs
```diff
@@ -33,7 +33,7 @@ fn build_auth_redirect_location(user: LoggedInUser) -> Result<String, Error> {
     let attributes = b64_encode_attributes_with_mapped_error(&user.attributes)?;
 
     Ok(format!(
-        "/index.html#/login?token={}&id={}&attributes={}",
+        "/ui/login?token={}&id={}&attributes={}",
```
ximon18, line 36: We should probably extract this common prefix to a constant, though this issue existed prior to this PR.

timbru: yeah, I prefer to leave this for now.. we should probably evaluate whether to use a web framework (axum maybe, is that the one you mentioned? @marlonbaeten ) rather than doing our own home grown routing at some point

timbru:
> The Cypress test suite no longer passes due to failures in finding expected content in the UI. @marlonbaeten / @timbru: Is it expected that it should pass? Is it superceded by something from tweedegolf and should be removed?

It should be okay to remove this now. There are new tests in krill-ui as far as I know.

ximon18:
> > The Cypress test suite no longer passes due to failures in finding expected content in the UI. @marlonbaeten / @timbru: Is it expected that it should pass? Is it superceded by something from tweedegolf and should be removed?
> 
> It should be okay to remove this now. There are new tests in krill-ui as far as I know.

I'll remove the Cypress tests.

PR #1028, opened 2023-04-06
No session reset on 0.13 upgrade

timbru:
With this we no longer need to do an RRDP session reset after migration to 0.13.0. This will reduce the impact significantly on busy repositories.

In the process the code was reorganised a bit. Hopefully cleaner now.. in particular: in theory the same object could be published in multiple locations (names) by a publisher. It does not make sense to do so, but it's allowed by the spec. By using the URI as the object key instead of the hash this is now possible.

### DIFF
--- a/src/commons/api/rrdp.rs
+++ b/src/commons/api/rrdp.rs
@@ -4,7 +4,7 @@
 use std::{
     convert::{TryFrom, TryInto},
     fmt, io,
-    ops::{Add, AddAssign},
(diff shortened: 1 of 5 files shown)


timbru:
Will merge this into [remove-published-objects-duplication](https://github.com/NLnetLabs/krill/tree/remove-published-objects-duplication) - it changes quite a few other changes introduced there, so it's probably best to review the final result after merging this into the other PR.

Issue #1030, opened 2023-04-08
Make sure that the CSR uses a trailing slash for id-ad-caRepository


timbru:
It seems that behaviour in rpki-rs changed and our trick to enforce that a trailing slash is added no longer works.

This issue results in certificates being requested without a trailing slash for id-ad-caRepository. This is incorrect.

Object validity does not seem to be affected because RP software generally follows id-ad-rpkiManifest instead and then resolves object names.

This should be fixed of course, but given that object validity is not affected it should be enough to fix this in the coming 0.13.0 release.

We should also build in a check to force krill to request a new certificates if its current certificate lacks the trailing slash.

PR #1034, opened 2023-04-12
Enforce that id-ad-caRepository uses a trailing slash. (#1030)

timbru:
nevermind.. the change I made that fxes this is actually in rpki-rs. Hold on..

### DIFF
--- a/src/commons/crypto/signing/dispatch/krillsigner.rs
+++ b/src/commons/crypto/signing/dispatch/krillsigner.rs
@@ -251,7 +251,7 @@ impl KrillSigner {
         let signed_and_encoded_csr = Csr::construct_rpki_ca(
             &self.router,
             key,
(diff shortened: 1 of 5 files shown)


src/commons/crypto/signing/dispatch/krillsigner.rs
```diff
@@ -251,7 +251,7 @@ impl KrillSigner {
         let signed_and_encoded_csr = Csr::construct_rpki_ca(
             &self.router,
             key,
-            &base_repo.ca_repository(name_space).join(&[]).unwrap(), // force trailing slash
+            &base_repo.ca_repository(name_space),
```
timbru, line 254: Note: the trailing slash is now enforced in both `ca_repository` and `Csr::construct_rpki_ca` in the rpki-rs 0.16.0.

PR #1035, opened 2023-04-12
Remove Cypress based UI tests

ximon18:
As the new Krill UI 2.0 has its own test suite.

Note that this PR also removes the OIDC mock and related tests for which we have no replacement at present, especially for the cases where incorrect OIDC provider behaviour or incorrectly configured OIDC responses are tested. Manual testing will suffice for now, and later we should perhaps factor out the OIDC stuff entirely to a separate crate complete with its own tests, and not depend on a real user agent (i.e. Cypress driven FireFox/Chrome as it is now) for integration testing but test more at the unit test level.


timbru (approved):
Looks good to me. Indeed.. it would be good to split Krill into sub-crates now that it's mostly mature.

This should also help to bring down compilation times, and allow us to test things in isolation.

ximon18:
@timbru: This PR only makes sense to merge if the pending Krill UI 2.0 PRs are merged so I won't merge it yet unless you're sure you'll merge the Krill UI 2.0 PRs and want this one closed to keep things organized?

timbru:
@ximon18 makes sense.. I will go over the Krill UI 2.0/2.1 PRs - I hope we can merge them.

timbru:
Merging - Krill UI 2.0 will be merged shortly.

PR #1036, opened 2023-04-12
Improve child parent publisher logging

timbru:


timbru:
Will close this and make a clean PR with just the last commit

PR #1047, opened 2023-05-02
Server side password hashing + basic auth

marlonbaeten:


timbru (approved):

timbru:
tested and works - there is a remaining issue that should be fixed before releasing 0.13.0, but it does not have to stop the RC3 release: https://github.com/NLnetLabs/krill-ui/issues/39

Issue #1063, opened 2023-05-18
Krill MUST NOT create only a single AFI ASPA


jayb-github:
Per this SIDROPS-list discussion: https://mailarchive.ietf.org/arch/msg/sidrops/QtMzLIz0yJwNCUVdvC_1cJjEIHc/

... an ASPA specified for only one AFI will result in RPs signaling an AS0 VAP for the other AFI.

It would better for Krill to not allow an operator to create such a counter-intuitive and probably unexpected situation, but instead to require that after performing the requested operation, explicit ASPAs covering both AFIs MUST exist.

timbru:
Indeed.. okay, the ASPA support in Krill was made at a time when the restrictions had not yet been fully crystallised.

I will make a fix that forces the operator to make an explicit choice in new / updated objects. If the user intended to have an AS0 only for one address family then they can specify this explicitly, rather than getting this result implicitly, which may well surprise them indeed.

I will first finalise release 0.13.0 this Monday based on 0.13.0-rc6 without further changes. This is needed as it will allow Publication Servers to overcome issue #1023.

Then I will make a very quick follow up 0.13.1-rc1 with a fix for this, and #1061 and #1058.

timbru:
Contrary to the earlier message.. we decided to include this in release 0.13.0 which is going out in a few moments

PR #1064, opened 2023-05-22
Krill MUST NOT create only a single AFI ASPA #1063

timbru:


src/commons/api/aspa.rs
```diff
@@ -100,70 +100,28 @@ impl AspaDefinition {
...
-                                if removed_limit == existing_limit {
-                                    // there is a match to remove
-                                    return true;
-                                }
-                                // else -> there is no match to remove
-                            }
-                        }
-                    }
-                }
+    /// Returns true if this is a single AFI AspaDefinition. Technically,
```
partim, line 103: Your doc string and the method name contradict each other – the doc string suggests this returns true if there is only one AFI, the method name suggests it returns true if it has both AFIs.

partim (changes requested):

src/commons/api/aspa.rs
ximon18, line 103: The fn name says "both" but the comment says "single" - this sounds contradictory.

src/commons/api/aspa.rs
```diff
@@ -100,70 +100,28 @@ impl AspaDefinition {
...
+    /// it is allowed to omit one address family entirely, but this will
+    /// be interpreted as thought the user specified an AS0 provider for
```
ximon18, line 105: s/as thought/as though/

src/daemon/ca/certauth.rs
```diff
@@ -1841,21 +1845,40 @@ impl CertAuth {
...
-    /// Verifies whether the update needs to be applied
+    /// Verifies whether the update is allowed and needs to be applied.
     ///
     /// The update does not need to be applied if there would be no change in
     /// the configured AspaDefinition. I.e. this gives us idempotence and e.g. allows
     /// an operator just issue a command to add a provider for a customer ASN, and
     /// if it was already authorised then no work is needed.
-    fn updated_needed(&self, customer: AspaCustomer, update: &AspaProvidersUpdate) -> KrillResult<bool> {
-        if update.is_empty() {
+    fn updated_allowed_and_needed(&self, customer: AspaCustomer, update: &AspaProvidersUpdate) -> KrillResult<bool> {
```
ximon18, line 1854: s/updated/update/ ?

src/daemon/ca/certauth.rs
```diff
@@ -1841,21 +1845,40 @@ impl CertAuth {
         Ok(update_events)
     }
 
-    /// Verifies whether the update needs to be applied
+    /// Verifies whether the update is allowed and needs to be applied.
     ///
     /// The update does not need to be applied if there would be no change in
     /// the configured AspaDefinition. I.e. this gives us idempotence and e.g. allows
     /// an operator just issue a command to add a provider for a customer ASN, and
```
ximon18, line 1852: s/an operator just/an operator to just/

ximon18: Ah, ignore.

src/commons/api/aspa.rs
timbru: yeah, oops sorry.. refactored and then did not change this text

timbru: yeah, oops sorry.. refactored and then did not change this text

src/daemon/ca/certauth.rs
timbru: The function argument is the 'update', but the result is (potentially) something 'updated'. So, leaving this as-is.

src/commons/api/aspa.rs
timbru: fixed

timbru: fixed

src/commons/api/aspa.rs
timbru: fixed

timbru:
> I can't speak to whether or not other parts of the code base should be changed too or if this makes sense from a business logic point of view.

Discussion is still ongoing, but this seems safest for now.

This does not change existing ASPAs as far as they have been defined, but prevents new instances or changes that would result in this.

partim (approved):

PR #1082, opened 2023-06-21
parse error - 

kansal15:
adding double quotes to avoid the following error under 'bgp_risdumps_enabled' secrtion -- 

`
Could not parse config: Error parsing config file: /etc/krill.conf, error: invalid TOML value, did you mean to use a quoted string? at line 295 column 22
`


timbru:
Thanks. Apologies for the delay - this is now merged into the development branch.

Issue #1078, opened 2023-06-07
Let AggregateStore use 'ChangeSets' rather than separate Commands and Events


timbru:
Currently we save a Command key, and then 0 (in case of an error) or more Events (in case there was a change).

We can simplify this by saving a single ChangeSet structure that includes all this. Just like WalStore does.

This will simplify the logic around finding these files. Each change will be in a single atomic stored key. It will also reduce the number of keys (files for most deployments).

timbru:
We will also need to re-work command history when we do this:
- [ ] keep a history structure instead of relying on command-key name syntax
- [ ] use predictable command keys (e.g. command-<seq>.json)
- [ ] keep the time of the command inside the command (not the key)
- [ ] add an optional comment/reason field for commands (while we're at it, useful for system commands related to re-issuance and upgrades)

Issue #1086, opened 2023-06-26
Remove no longer needed 'always_recover_data' function 


timbru:
This is no longer required now that the transactionality of changes is much improved: #1067, #1078, #1076, #1077, #1085, #1108, #1090.

In particular, #1078 ensures we cannot have partially written change sets. So, the recovery function that we had no longer makes sense.

Issue #1076, opened 2023-06-07
Use transactions for fn command in AggregateStore


timbru:
Currently still uses an outer RwLock structure to enforce sequential updates. And then uses the Krill KeyValueStore primitives, which wrap kvx::KeyValueStore

This should be changed to use `kvx::KeyValueStore::transaction` instead. Then this will work across nodes if database storage is used, and we can remove the RwLock (which also has a risk of lock starvation when we have many reads).

We should also rethink how we use the in-memory cache for aggregates and update the contained arcs. There may be a `Mutex` and/or `Cow` we want to use...

PR #1087, opened 2023-06-30
Aggregate command sets

timbru:
See #1078


src/daemon/config.rs
```diff
@@ -448,6 +448,9 @@ pub struct Config {
     )]
     pub storage_uri: Url,
 
+    #[serde(default)]
+    pub disable_history_cache: bool,
```
ximon18, line 452: Should this be documented in krill.conf?

src/commons/api/history.rs
```diff
@@ -118,19 +58,20 @@ impl CommandHistory {
 
 impl fmt::Display for CommandHistory {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        writeln!(f, "time::command::key::success")?;
+        writeln!(f, "time::command::sequence::success")?;
```
ximon18, line 61: You say `sequence` here but below output a value called `version` - is `sequence` the right term to use here?

src/commons/api/history.rs
```diff
@@ -118,19 +58,20 @@ impl CommandHistory {
...
             let success_string = match &command.effect {
-                StoredEffect::Error { msg } => format!("ERROR -> {}", msg),
-                StoredEffect::Success { .. } => "OK".to_string(),
+                CommandHistoryResult::Init() => "INIT".to_string(),
+                CommandHistoryResult::Ok() => "OK".to_string(),
+                CommandHistoryResult::Error(msg) => format!("ERROR -> {}", msg),
             };
             writeln!(
                 f,
                 "{}::{} ::{}::{}",
```
ximon18, line 71: Given that above you say:
> `"time::command::sequence::success`

Is it correct that there is a space in `"{}::{} ::{}::{}"` (though this was also in the original code) ?

src/commons/api/history.rs
```diff
@@ -405,13 +328,40 @@ impl Default for CommandHistoryCriteria {
...
+        writeln!(f, "Time:   {}", self.time().to_rfc3339_opts(SecondsFormat::Secs, true))?;
+        writeln!(f, "Actor:  {}", self.actor())?;
+        writeln!(f, "Action: {}", self.details().summary().msg)?;
+
+        match self.effect() {
+            StoredEffect::Error { msg, .. } => writeln!(f, "Error:  {}", msg)?,
+            StoredEffect::Success { events } => {
+                writeln!(f, "Changes:")?;
+                for evt in events {
+                    writeln!(f, "  {}", evt)?;
```
ximon18, line 345: This used to write out `evt.details()`, is this a deliberate change?

src/commons/api/history.rs
```diff
@@ -538,144 +493,167 @@ impl WithStorableDetails for StorableCaCommand {
...
 }
 
-impl fmt::Display for StorableCaCommand {
+impl fmt::Display for CertAuthStorableCommand {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
+            // ------------------------------------------------------------
+            // Initialisation
+            // ------------------------------------------------------------
+            CertAuthStorableCommand::Initialise => write!(f, "Create CA"),
```
ximon18, line 604: Stupid question: Is initialisation really the same thing as "creating" a CA? Initialisation isn't an internal detail that can happen more than once for a CA or is separate to a CA?

ximon18: Also for the equivalent below you wrote:
```
            StorableRepositoryCommand::Initialise => {
                write!(f, "Initialise server")
            }
```
So why write "Initialise" there and "Create" here?

src/commons/crypto/signing/dispatch/signerinfo.rs
```diff
@@ -131,6 +152,7 @@ enum SignerInfoCommandDetails {
 impl fmt::Display for SignerInfoCommandDetails {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
+            SignerInfoCommandDetails::Initialise => write!(f, "Create Signer"),
```
ximon18, line 155: Same comment as in history.rs, should this be "Initialise" instead of "Create"?

src/commons/eventsourcing/cmd.rs
```diff
@@ -20,6 +19,95 @@ use crate::commons::{
 /// CommandSummary for use in history.
 pub trait WithStorableDetails: Storable + Send + Sync {
     fn summary(&self) -> CommandSummary;
+
+    /// Create an instance representing the initialisation command/event.
+    ///
+    /// Should not contain any data, so that we can keep this signature
```
ximon18, line 25: Which signature? The function signature? Why is it important to keep it "simple" (and assuming we even agree on the meaning of "simple" in this context...)

src/commons/eventsourcing/cmd.rs
```diff
@@ -20,6 +19,95 @@ use crate::commons::{
...
+    }
+
+    pub fn into_details(self) -> I {
+        self.details
+    }
+}
+
+impl<I: InitCommandDetails> fmt::Display for SentInitCommand<I> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "initialise '{}'", self.handle)
```
ximon18, line 99: Other implementations like this call `self.store.fmt()`. Is it right that that is different here?

src/commons/eventsourcing/evt.rs
```diff
@@ -1,66 +1,6 @@
...
-        self.version
-    }
-}
-
-impl<E: fmt::Display + Eq + PartialEq + Storable + Send + Sync + 'static> fmt::Display for StoredEvent<E> {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "id: {} version: {} details: {}", self.id, self.version, self.details)
-    }
-}
+pub trait Event: fmt::Display + Eq + PartialEq + Send + Sync + Storable + 'static {}
```
ximon18, line 6: `04_es_krill.md` needs updating as `StoredEvent` is discussed but here it is removed.

src/commons/eventsourcing/mod.rs
```diff
@@ -294,65 +337,69 @@ mod tests {
 
     #[test]
     fn event_sourcing_framework() {
-        test_under_tmp(|data_dir| {
-            let counter = Arc::new(EventCounter::default());
-            let storage_uri = storage_uri_from_data_dir(&data_dir).unwrap();
-            let mut manager = AggregateStore::<Person>::create(&storage_uri, segment!("person")).unwrap();
-            manager.add_post_save_listener(counter.clone());
+        // crate::test::test_under_tmp(|data_dir| {
```
ximon18, line 340: Is there a reason to leave this here? Were you done and this is can be removed, or where you in the middle of something and it's actually still relevant?

ximon18: Ah I guess this is obsoleted by the KVX changes and should be removed?

src/commons/eventsourcing/store.rs
```diff
@@ -1,155 +1,69 @@
...
 };
 
+use super::InitCommand;
+
 pub type StoreResult<T> = Result<T, AggregateStoreError>;
 
 //------------ Storable ------------------------------------------------------
 
 pub trait Storable: Clone + Serialize + DeserializeOwned + Sized + 'static {}
 impl<T: Clone + Serialize + DeserializeOwned + Sized + 'static> Storable for T {}
```
ximon18, line 30: Removing this impacts `06_ca_manager.md`.

src/commons/eventsourcing/store.rs
```diff
@@ -1,155 +1,69 @@
...
-    }
-}
-
 //------------ AggregateStore ------------------------------------------------
 
 /// This type is responsible for managing aggregates.
 pub struct AggregateStore<A: Aggregate> {
     kv: KeyValueStore,
     cache: RwLock<HashMap<MyHandle, Arc<A>>>,
+    history_cache: Mutex<Option<HashMap<MyHandle, Vec<CommandHistoryRecord>>>>,
```
ximon18, line 37: I'm not sure if I'm thinking about this right but wouldn't that be better as `Option<Mutex>` ?

src/commons/eventsourcing/store.rs
```diff
@@ -178,152 +92,94 @@ where
     pub fn warm_aggregate(&self, handle: &MyHandle) -> StoreResult<()> {
```
ximon18, line 92: Is the fn comment about "surplus entries will be archived" still accurate given the code removed and changes made to `fn recover()` ?

ximon18: Ah I see in #1078 you haven't yet ticked the `Fill cache on warmup` box, so maybe that relates to this.

src/commons/eventsourcing/store.rs
```diff
@@ -178,152 +92,94 @@ where
...
+
+            let aggregate_opt = match self.get_command(&handle, 0) {
+                Err(_) => None,
+                Ok(init_command) => {
+                    if let Some(init_event) = init_command.into_init() {
+                        // Initialise
+                        let mut aggregate = A::init(handle.clone(), init_event);
+
+                        // Find the next command and apply it until there
+                        // is no (valid) next command.
```
ximon18, line 127: Did you deliberately drop the progress reporting, is that an accident, or on the TODO list for later?

src/commons/eventsourcing/store.rs
```diff
@@ -178,152 +92,94 @@ where
...
-                snapshot_version,
-            };
+                    self.kv.drop_scope(&scope)?;
+                }
+                Some(aggregate) => {
+                    // Archive any and all keys that are not command keys
+                    // for versions we just applied.
+                    for key in self.kv.keys(&scope, "")? {
+                        // command keys use: command-#.json
+                        let keep = if let Some(pfx_removed) = key.name().as_str().strip_prefix("command-") {
```
ximon18, line 154: You could simplify this with a `let mut keep = false` and only set it to `true` in the deepest if level.

src/commons/eventsourcing/agg.rs
```diff
@@ -19,22 +21,41 @@ use crate::commons::eventsourcing::WithStorableDetails;
...
+    /// it's quite convenient to store it inside an Aggregate as well.
+    ///
+    /// More importantly, the handle is not typically included in the
+    /// InitEvent itself.
+    fn init(handle: MyHandle, event: Self::InitEvent) -> Self;
+
+    /// Tries to initialise a new InitEvent for a new instance. This
+    /// can fail. The InitEvent is not applied here, but returned so
+    /// that we can re-build state from history.
+    fn process_init_command(command: Self::InitCommand) -> Result<Self::InitEvent, Self::Error>;
```
ximon18, line 51: I'm finding the name of this fn confusing. It doesn't "process" anything, it's more of a From impl it seems.

src/commons/eventsourcing/store.rs
```diff
@@ -356,19 +212,31 @@ where
...
+        let init_event = A::process_init_command(cmd).map_err(|_| AggregateStoreError::InitError(handle.clone()))?;
+        let aggregate = A::init(handle.clone(), init_event.clone());
+
+        // Store the init command. It is unlikely that this should fail, but
+        // if it does then there is an issue with the storage layer that we cannot
+        // recover from. So, exit.
+        let processed_command = processed_command_builder.finish_with_init_event(init_event);
+        if let Err(e) = self.store_command(&processed_command) {
+            error!("Cannot save state for '{}'. Got error: {}", handle, e);
+            error!("Will now exit Krill - please verify that the disk can be written to and is not full");
```
ximon18, line 233: As kvx is now (starting to be?) used is a reference to "disk" here still correct?

src/commons/eventsourcing/store.rs
```diff
@@ -417,99 +281,86 @@ where
...
-                self.save_info(&handle, &info)?; // last_command was updated
+                // Store the processed command with the error.
+                //
+                // If persistence fails, then complain loudly, and exit. Krill should not keep running, because this would
+                // result in discrepancies between state in memory and state on disk. Let Krill crash and an operator investigate.
+                // See issue: https://github.com/NLnetLabs/krill/issues/322
+                let processed_command = processed_command_builder.finish_with_error(&e);
+                if let Err(e) = self.store_command(&processed_command) {
+                    error!("Cannot save state for '{}'. Got error: {}", handle, e);
+                    error!("Will now exit Krill - please verify that the disk can be written to and is not full");
```
ximon18, line 302: Same comment about "disk" as above.

src/commons/eventsourcing/store.rs
```diff
@@ -417,99 +281,86 @@ where
...
+                    std::process::exit(1);
```
ximon18, line 303: Maybe move the exit code to an application wide constant?

src/commons/eventsourcing/store.rs
```diff
@@ -417,99 +281,86 @@ where
...
+                }
+
+                // Update the cached aggregate so that its version is incremented
+                let agg = Arc::make_mut(&mut latest);
```
ximon18, line 307: `make_mut()` may clone the inner value, i.e. your version increment may not be done on the original inner value - is that definitely okay here?

src/commons/eventsourcing/store.rs
```diff
@@ -417,99 +281,86 @@ where
...
-                        if event.version() != expected_version || event.handle() != &handle {
-                            error!("Unexpected event: {}", event);
-                            return Err(A::Error::from(AggregateStoreError::WrongEventForAggregate(
-                                handle,
-                                event.handle().clone(),
-                                expected_version,
-                                event.version(),
-                            )));
+                    // Unfortunately, this means that we will need to clone the command.
+                    let agg = Arc::make_mut(&mut latest);
```
ximon18, line 327: Same question about `make_mut()` as above.

src/daemon/properties/mod.rs
```diff
@@ -52,12 +78,16 @@ impl fmt::Display for PropertiesCommandDetails {
...
 pub enum StorablePropertiesCommand {
+    Initialise,
     UpgradeTo { krill_version: KrillVersion },
 }
 
 impl fmt::Display for StorablePropertiesCommand {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
+            Self::Initialise => {
+                write!(f, "create properties")
```
ximon18, line 89: Same comment as elsewhere about whether this should say "create" or "initialise".

src/daemon/properties/mod.rs
```diff
@@ -146,32 +179,37 @@ impl Aggregate for Properties {
...
+            krill_version: command.into_details().krill_version,
+        })
+    }
+
     fn version(&self) -> u64 {
         self.version
     }
 
-    fn apply(&mut self, event: Self::Event) {
+    fn increment_version(&mut self) {
```
ximon18, line 206: Should this be a default fn body in the trait, as you seem to do this in a lot of places...

src/daemon/ta/proxy.rs
```diff
@@ -186,6 +219,9 @@ impl fmt::Display for TrustAnchorProxyCommandDetails {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         // note that this is a summary, full details are stored in the json.
         match self {
+            TrustAnchorProxyCommandDetails::Initialise => {
+                write!(f, "Create TA proxy")
```
ximon18, line 223: Same comment as elsewhere as to whether this should be "create" or "initialise".

src/daemon/ta/proxy.rs
```diff
@@ -435,187 +484,186 @@ impl eventsourcing::Aggregate for TrustAnchorProxy {
...
+            TrustAnchorProxyCommandDetails::Initialise => {
+                // This can't happen really.. we would never send this command
+                // to an existing TrustAnchorProxy.
+                //
+                // This could be solved more elegantly, and more verbosely, if
+                // we create a separate TrustAnchorProxyStorableCommand that
+                // implements 'WithStorableDetails' - like we have in other cases -
+                // because then our initialisation command could map to that type
+                // instead of having this additional variant for storing.
+                Err(Error::custom("Trust Anchor Proxy already created"))
```
ximon18, line 499: Same comment as elsewhere as to whether this should say "created" or "initialised".

src/commons/eventsourcing/agg.rs
ximon18: Ah, ignore me, I see that in some impls it indeed does actual work.

src/upgrades/pre_0_10_0/cas_migration.rs
```diff
@@ -77,192 +81,79 @@ impl CasMigration {
...
-        .prepare_new_data(mode)
+        .upgrade(mode)
     }
 }
 
-impl UpgradeStore for CasMigration {
-    fn prepare_new_data(&self, mode: UpgradeMode) -> Result<(), PrepareUpgradeError> {
-        // check existing version, wipe if needed
-        self.preparation_store_prepare()?;
+impl UpgradeAggregateStorePre0_14 for CasMigration {
```
ximon18, line 88: Errm, I'm confused. This is called `Pre0_14` but is inside the `pre_0_10_0` directory? What am I missing?

src/upgrades/mod.rs
```diff
@@ -220,79 +223,335 @@ impl UpgradeMode {
...
+            // changes were applied.
+            self.update_data_upgrade_info(&scope, &data_upgrade_info)?;
+
+            info!("Verified migration of '{}'", handle);
+        }
+
+        match mode {
+            UpgradeMode::PrepareOnly => {
+                info!(
+                    "Prepared migrating CAs to Krill version {}. Will save progress for final upgrade when Krill restarts.",
```
ximon18, line 443: Above you no longer refer to CA but instead this fn has become more generic. Is it correct to still refer to CA here?

src/upgrades/pre_0_10_0/cas_migration.rs
```diff
@@ -77,192 +81,79 @@ impl CasMigration {
...
-    fn deployed_store(&self) -> &KeyValueStore {
-        &self.current_kv_store
+        Ok(Some(new_command))
     }
 
-    fn preparation_store(&self) -> &KeyValueStore {
-        &self.new_kv_store
+    /// Override post migration, we need to do extra stuff.
+    fn post_migration(&self, handle: &MyHandle) -> UpgradeResult<()> {
+        info!("Will migrate the current repository objects for CA '{}'", handle);
```
ximon18, line 156: How can "will migrate" which seems to refer to a future migration be correct in a *post* migration step? Should this say "Will verify" instead of "Will migrate"? And should it refer to "CA" here or be more generic?

src/commons/api/history.rs
timbru: Missed that one. I will do a search for 'sequence' and replace it with 'version'.

src/commons/api/history.rs
timbru: Yes. `evt` was a stored event containing meta information and the actual `details`. Now, the `evt` is just the details.

src/commons/api/history.rs
timbru: Yeah, you're right. As with "version" and "sequence", I should choose one name for this thing. I will go for "Initialise".

src/upgrades/mod.rs
```diff
@@ -480,87 +773,103 @@ pub fn finalise_data_migration(
...
-        // upgrade-data/pubd exists
-        if pubd_upg.exists() {
-            // pubd -> arch-pubd-{old-version}
-            // upgrade-data/pubd -> pubd
-            move_dir_if_exists(&pubd, &pubd_arch)?;
-            move_dir_if_exists(&pubd_upg, &pubd)?
-        }
+                let version_file = current_dir.join("version");
+                if version_file.exists() {
+                    debug!("Removing excess version file: {}", version_file.to_string_lossy());
```
ximon18, line 843: Is it old or excess, you refer to it differently in these two log messages.

ximon18 (approved):

src/commons/eventsourcing/cmd.rs
timbru: fixed

src/commons/eventsourcing/store.rs
timbru: Maybe... not sure. I changed it but then had to move things around a bit because of lifetimes. It will be slightly more efficient in the non-default case where no cache is used.

src/upgrades/mod.rs
```diff
@@ -220,79 +223,335 @@ impl UpgradeMode {
...
+        );
+
+        // Migrate the event sourced data for each scope and create new snapshots
+        for scope in self.deployed_store().scopes()? {
+            // Getting the Handle should never fail, but if it does then we should bail out asap.
+            let handle = MyHandle::from_str(&scope.to_string())
+                .map_err(|_| UpgradeError::Custom(format!("Found invalid handle '{}'", scope)))?;
+
+            // Get the upgrade info to see where we got to.
+            // We may be continuing from an
```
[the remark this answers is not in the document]
timbru: fixed

src/upgrades/mod.rs
timbru: good catch, changed to data

src/upgrades/pre_0_10_0/cas_migration.rs
timbru: It's part of the initial migration but post *command* migration. Changed the fn name to post_command_migration.

src/upgrades/mod.rs
```diff
@@ -480,87 +773,103 @@ pub fn finalise_data_migration(
...
+                let current_dir = data_dir.join(ns.as_str());
+
+                if upgraded_dir.exists() {
+                    // Data was prepared. So we archive the current data and
+                    // then move the prepped data.
+                    move_dir(&current_dir, &archive_dir)?;
+                    move_dir(&upgraded_dir, &current_dir)?;
+                } else if current_dir.exists() {
+                    // There was no new data for this directory. But, we make
+                    // make a backup so that we can have a consistent data set
```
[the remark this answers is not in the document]
timbru: fixed

src/upgrades/mod.rs
timbru: fixed

src/upgrades/pre_0_10_0/cas_migration.rs
timbru: Well, it's both.. but the trait is generic for all pre0_14 migrations.

src/commons/eventsourcing/cmd.rs
timbru: Yeah, this is correct. This is one level higher - those other implementations are on the level of the "command details".

src/commons/eventsourcing/store.rs
timbru: I no longer know the length of the list in advance. And.. we may need to / want to drop this anyway. See issue #1086. In a nutshell: recover is not reliable, it's based on replaying past command/events to a last 'good' state, but that does not work for types that are not event sourced - such as the CaObjects struct.

src/commons/eventsourcing/store.rs
timbru: I extracted this to a separate function and improved the error message.

src/commons/eventsourcing/store.rs
timbru: This should work. It was not changed but moved around.

That said, see issue #1076. We need to rethink how we do transactions, and this will affect how we use the cache.

src/commons/eventsourcing/store.rs
timbru: This should work. It was not changed but moved around.

That said, see issue https://github.com/NLnetLabs/krill/issues/1076. We need to rethink how we do transactions, and this will affect how we use the cache.

src/daemon/properties/mod.rs
timbru: fixed

src/daemon/properties/mod.rs
timbru: It needs to modify the data of the implementing struct, so I am unsure how.

src/commons/eventsourcing/evt.rs
timbru: Created issue: [Revise module structure and update rustdoc #1091](https://github.com/NLnetLabs/krill/issues/1091)

src/commons/eventsourcing/store.rs
timbru: Created issue: [Revise module structure and update rustdoc #1091](https://github.com/NLnetLabs/krill/issues/1091)

Issue #1094, opened 2023-07-07
Support migrations using non-disk storage


timbru:
No longer require that "upgrade_storage_uri" is set. This will enable migrations that use non-disk storage in future, but it should be noted that the migration from disk to non-disk (postgresql) storage is a separate step that should only be done *after* upgrading to 0.14.0 - see issue #1079.

- [x] Update `kvx` to support namespace migrations.
- [x] Remove upgrade_storage_uri from `Config`
- [x] Remove all direct file / directory access in upgrade code (use KV instead)
- [x] Move `data_dir_from_storage_uri` to `Config` (only used for `tls_dirs`, `repo_dir` and `pid_file`)
- [x] Move `storage_uri_from_data_dir` to `Config`
- [x] Use the in-memory data store for automated upgrade tests (make this configurable)

Note: issue #1098 was made to improve version upgrades in clustered setups.

PR #1099, opened 2023-08-30
Support migrations using non-disk storage #1094

timbru:
Note: we need to release kvx first.


src/commons/util/mod.rs
```diff
@@ -36,6 +35,11 @@ impl KrillVersion {
         Self::from_str(KRILL_VERSION).unwrap()
     }
 
+    /// Make a notation friendly to namespaces for upgrades.
+    pub fn hyphen_notated(&self) -> String {
+        format!("{}-{}-{}{}", self.major, self.minor, self.patch, self.release_type)
```
Koenvh1, line 40: There's in principle nothing that guarantees that `self.release_type` will be in a format friendly to namespaces, but I guess we are not going to ever put something silly in there, and if we do we will probably notice it anyway.

Koenvh1 (approved):
Looks good to me. There are two very very very minor things that you should feel free to ignore.

src/commons/util/mod.rs
timbru: In principle, these values come from the `version` entry in Cargo.toml. And we use a sane scheme for the release_type postfix.

src/upgrades/mod.rs
```diff
@@ -918,82 +815,71 @@ pub fn finalise_data_migration(
...
+                    if let Some(signer_handle) = &openssl_signer_handle {
+                        let internal_key_id = key_id.to_string();
+                        if let Some(mapper) = krill_signer.get_mapper() {
+                            mapper.add_key(signer_handle, &key_id, &internal_key_id)?;
+                            num_recorded_keys += 1;
+                        }
+                    }
+                }
+            } else {
+                debug!("Could not parse key as key identifier: {}", key);
```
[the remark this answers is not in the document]
timbru: Yeah, that's okay - it means that someone mucked about with files on disk manually. So, they are kind of on their own then anyway.

PR #1100, opened 2023-09-01
Migrate data store

timbru:
[draft] should probably be after #1099 

This PR adds support to `krillup` to migrate data from one store to another. It seems to work both in unit test and by manual testing of the actual `krillup` binary.

However, in doing this work I discovered that the (hidden) runtime that is used in the synchronous `kvx` lib will not work with Krill - as it already has a (tokio) runtime. So that will need thought, and a separate fix and PR.


Koenvh1 (approved):
Looks good.

This is a bit of a nit, but I think the flow of `main` could be made a bit clearer. I would abstract the actual matching of matches (i.e. line 19 of `src/bin/krillup.rs` and down) to another function. Now after `parse_matches` the code goes back to the top again. Alternatively you could inline it, or leave it as-is. I don't consider it a blocking issue, but something to perhaps consider.

PR #1104, opened 2023-09-06
Use transactions/locks (kvx::execute) for AggregateStore.

timbru:
[draft] merge to `prep-0.14.0-rc1` after PR #1100


src/daemon/ca/manager.rs
```diff
@@ -116,7 +116,7 @@ impl CaManager {
             // This is slow, but it will ensure that all commands and events are accounted for,
             // and there are no incomplete changes where some but not all files for a change were
             // written to disk.
-            ca_store.recover()?;
+            todo!("issue #1086");
```
Koenvh1, line 119: `todo!` panics, which I am not quite sure is what we want in this case even if the config option no longer makes sense. Then again it is a nit and #1117 solves that anyway.

Koenvh1 (approved):

src/daemon/ca/manager.rs
timbru: This is removed in #1117 indeed. Just left this here as things were in progress.

PR #1110, opened 2023-09-08
Kvstore tx

timbru:
This PR makes the `krill::commons::eventsourcing::kv::KeyValueStore` use transactions/locking for all functions, except for migration functions where there is no clear scope or more than one scope.


src/bin/krillup.rs
```diff
@@ -1,82 +1,145 @@
...
-                    eprintln!();
-                    eprintln!("Note that your server data has NOT been modified. Do not upgrade krill itself yet!");
-                    eprintln!("If you did upgrade krill, then downgrade it to your previous installed version.");
-                    ::std::process::exit(1);
-                }
-            };
+fn main() {
+    let matches = make_matches();
+
+    match parse_matches(matches) {
```
Koenvh1, line 19: This flow feels illogical. I am down the file and then continuing up here again. Then again this is probably a personal preference so feel to ignore it :-)

Koenvh1 (approved):

src/bin/krillup.rs
timbru: I usually like to have `main` as the first thing. But the whole match could be in a separate function as well. It's personal preference territory.

PR #1144, opened 2023-10-26
Fix prometheus type for krill_cas_bgp_roas_total

steffann:
My Zabbix server noticed an error in the /metrics endpoint :)


timbru:
Ah! Oops, thanks! I think I can sneak this one into the release this Monday.

steffann:
image

Credits to Zabbix for a helpful error message :)

PR #1152, opened 2023-11-06
Support async storage

timbru:
This makes the `KeyValueStore` enum async, which, of course, has a rather big impact on the code base.

The enum delegates to `KeyValueStoreImpl` which is still sync. The actual `Disk` and `Memory` implementations are also sync. We can make this implementation async in a future a PR. We may be able to remove the indirection to the `KeyValueStoreImpl` enum then as well. We will see. When this is done, we can also port the `kvx` implementation for postgresql - albeit, using async.

Just mentioning this here now so that it is clear that there is more work, but I intentionally wanted to keep this PR a bit smaller so that it can be reviewed better.


partim:
The code has diverged from this PR quite a bit, so there’d be hell in trying to merge. Probably easier to redo this when it is time,

Issue #1160, opened 2023-11-19
Always write to temp file first, then move


timbru:
I think this used to work before kvx, but I am not 100% sure. In any case... we should write all new files to a tempfile first, and _then_ rename. Otherwise we can end up with halfwritten file if the system is rebooted.

Current behaviour of the KeyValueStore, now imported from kvx, is to only use a tempfile if the target file already exists.

ximon18:
This was included in [v0.14.3](https://github.com/NLnetLabs/krill/releases/tag/v0.14.3).

PR #1162, opened 2023-11-23
Recover from issues with corrupt files on upgrade to 0.14.3 (#1160)

timbru:
This should fix issues. It still needs more manual testing. Unfortunately, it's not easy to test this in automated way currently as our upgrade test code assumes that values can be read before putting them into a memstore (which only accepts valid json) for testing. This needs some thought - or a very thorough review and manual testing.

There is a happy case test, but that one essentially does nothing - it covers the case where an upgrade is done and there were no corrupt files.


test-resources/migrations/v0_14_2/tasks/pending/1700751070000-sweep_login_cache
```diff
@@ -0,0 +1,3 @@
+{
```
ximon18, line 1: Why don't these files have `.json` extensions?

Cargo.lock
```diff
@@ -1081,7 +1081,7 @@ dependencies = [
 
 [[package]]
 name = "krill"
-version = "0.15.0-dev"
+version = "0.14.3"
```
ximon18, line 1084: Shouldn't this be left as 0.15.0-dev?

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 # Note: some of these values are also used when building Debian packages below.
 name = "krill"
-version = "0.15.0-dev"
+version = "0.14.3"
```
ximon18, line 4: Shouldn't this be left as 0.15.0-dev?

src/upgrades/mod.rs
```diff
@@ -1134,6 +1165,26 @@ mod tests {
         );
     }
 
+    #[test]
+    fn prepare_then_upgrade_0_14_2() {
+        test_upgrade(
+            "test-resources/migrations/v0_14_2/",
+            &[
+                "ca_objects",
+                "cas", // testbed/command-6.json was intentionally removed
```
ximon18, line 1174: "Removed" - is that right? I don't see it in earlier test-resources/migrations/xxx/cas/ directories so it wasn't removed compared to them, but it is _missing_ in the sequence as there is command-5.json and command-7.json but no command-6.json.

src/upgrades/pre_0_14_3/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+            warn!("Pending task could not be parsed. Dropping: {}", pending_key);
+            task_store.delete(&pending_key).map_err(|e| {
+                UpgradeError::Custom(format!("Cannot delete corrupt task: {}. Error: {}", pending_key, e))
+            })?;
+        }
+    }
+
+    for running_key in task_store
+        .list_keys(&running_scope)
+        .map_err(|e| UpgradeError::Custom(format!("Cannot read running tasks: {}", e)))?
```
ximon18, line 131: Given the description on this fn, why can this fail?

src/upgrades/pre_0_14_3/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+pub fn upgrade_tasks(config: &Config) -> UpgradeResult<()> {
+    let task_store = kvx::KeyValueStore::new(&config.storage_uri, TASK_QUEUE_NS)
+        .map_err(|e| UpgradeError::Custom(format!("Cannot create task store: {}", e)))?;
+
+    let pending_scope = Scope::from_segment(segment!("pending"));
+    let running_scope = Scope::from_segment(segment!("running"));
+
+    for pending_key in task_store
+        .list_keys(&pending_scope)
+        .map_err(|e| UpgradeError::Custom(format!("Cannot read pending tasks: {}", e)))?
```
ximon18, line 119: Given the description on this fn, why can this fail?

src/upgrades/mod.rs
```diff
@@ -769,6 +777,29 @@ pub fn prepare_upgrade_data_migrations(
...
+            } else if versions.from < KrillVersion::release(0, 14, 3) {
+                // Check all possibly affected stores for corrupted files resulting
+                // from issue #1160 and fix them if needed.
+                let mut migrated = false;
+
+                pre_0_14_3::upgrade_tasks(config)?;
+                pre_0_14_3::upgrade_status(config)?;
+
+                migrated = pre_0_14_3::upgrade_agg::<CertAuth>(CASERVER_NS, config)? || migrated;
+                // note: ca_objects is not affected by #1160, as it replaces existing objects and in that
```
ximon18, line 789: Does this comment refer to the line above?

src/upgrades/mod.rs
```diff
@@ -769,6 +777,29 @@ pub fn prepare_upgrade_data_migrations(
...
                 Ok(Some(UpgradeReport::new(aspa_configs, true, versions)))
+            } else if versions.from < KrillVersion::release(0, 14, 3) {
+                // Check all possibly affected stores for corrupted files resulting
+                // from issue #1160 and fix them if needed.
+                let mut migrated = false;
+
+                pre_0_14_3::upgrade_tasks(config)?;
+                pre_0_14_3::upgrade_status(config)?;
+
+                migrated = pre_0_14_3::upgrade_agg::<CertAuth>(CASERVER_NS, config)? || migrated;
```
ximon18, line 788: Could these `migrated = N || migrated` lines not be done as `migrated |= N` ?

src/upgrades/pre_0_14_3/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+    {
+        if task_store.get(&pending_key).is_err() {
+            warn!("Pending task could not be parsed. Dropping: {}", pending_key);
+            task_store.delete(&pending_key).map_err(|e| {
+                UpgradeError::Custom(format!("Cannot delete corrupt task: {}. Error: {}", pending_key, e))
```
ximon18, line 124: Rather than error out here, if this fails could you not try and wipe the whole store as is done in `upgrade_status()`?

src/upgrades/pre_0_14_3/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+                }
+            }
+        }
+
+        Ok(true)
+    }
+}
+
+/// Check the task store for corrupted tasks due to issue #1160, and if present
+/// drop them in place. This is safe to do because missing tasks are re-added at start up.
```
ximon18, line 109: Why, here and in the warning messages below, do you use the term "drop"? "Drop" has specific meaning in Rust but that isn't the meaning that applies here, rather items are being removed/deleted from the store, and in the most common (currently the only) end-user scenarion this will be deletion of files on disk. Wouldn't delete be a word that operators will understand better than drop?

src/upgrades/mod.rs
```diff
@@ -769,6 +777,29 @@ pub fn prepare_upgrade_data_migrations(
                 pre_0_14_0::UpgradeAggregateStoreTrustAnchorProxy::upgrade(TA_PROXY_SERVER_NS, mode, config)?;
 
                 Ok(Some(UpgradeReport::new(aspa_configs, true, versions)))
+            } else if versions.from < KrillVersion::release(0, 14, 3) {
+                // Check all possibly affected stores for corrupted files resulting
+                // from issue #1160 and fix them if needed.
```
ximon18, line 782: This comment, to me at least, is misleading. Nothing will be "fixed", if problematic store items are found they are deleted which to me is quite different than fixing them.

ximon18: Correction: Deletion only happens in the `upgrade_tasks()` and `upgrade_status()` calls, I see that the `upgrade_agg()` call for example does attempt a sort of fix via replay (though replay as far as it can which presumably is also "deleting" the bits that it can't).

Cargo.toml
timbru: Perhaps it would be more clean to make this change in a separate PR. If you prefer I can do that. But, this is going to be in a bug fix release, so 0.14.3 is more appropriate than 0.15.0.

Cargo.lock
timbru: Perhaps it would be more clean to make this change in a separate PR. If you prefer I can do that. But, this is going to be in a bug fix release, so 0.14.3 is more appropriate than 0.15.0.

test-resources/migrations/v0_14_2/tasks/pending/1700751070000-sweep_login_cache
timbru: no strong reason, except that we already had a database (and memory) store in mind, so the extension which seems more appropriate to filesystems was dropped along the way.

Addding it now means that we should either wipe the task queue on upgrade startup, or map the existing keys. So, I'd like to avoid it for now.

src/upgrades/mod.rs
```diff
@@ -1134,6 +1165,26 @@ mod tests {
...
     }
 
+    #[test]
+    fn prepare_then_upgrade_0_14_2() {
+        test_upgrade(
+            "test-resources/migrations/v0_14_2/",
+            &[
+                "ca_objects",
+                "cas", // testbed/command-6.json was intentionally removed
+                "keys",
```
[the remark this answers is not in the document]
timbru: No, not for this test. Generally speaking it may be better to leave the keys dir out actually as it contains private keys. Even if this is all test data, it's a good habit not to include them.

src/upgrades/mod.rs
timbru: The left-hand function must be evaluated first, so I don't think so.

timbru:
After discussion - we will close this. It's only useful to do this under very specific circumstances: i.e. there is a 0.14.0/.1/.2 installation that runs into the issue that a tempfile is not used, and a resulting file is half-written due to a hard server crash or disk full situation, and then the user upgrades to this. Otoh it could cause issues in other situations. So, it was decided that it's best not to try any automated recover here. Just let users update to 0.14.3, and talk to use if they get hit by the issue in 0.14.0-0.14.2

PR #1163, opened 2023-11-27
Fix aspa migration issues

timbru:


src/upgrades/pre_0_14_0/cas_migration.rs
```diff
@@ -147,12 +147,33 @@ impl UpgradeAggregateStorePre0_14 for CasMigration {
...
                         for old_event in events {
-                            full_events.push(old_event.into());
+                            match old_event {
+                                Pre0_14_0CertAuthEvent::AspaConfigAdded { .. }
+                                | Pre0_14_0CertAuthEvent::AspaConfigRemoved { .. }
+                                | Pre0_14_0CertAuthEvent::AspaConfigUpdated { .. }
+                                | Pre0_14_0CertAuthEvent::AspaObjectsUpdated { .. } => {
+                                    // we only expect AspaObjectsUpdated to be possible outside of
+                                    // Aspa related commands, e.g. because of a key rollover, but
+                                    // to be sure.. we do not migrate any of the ASPA events in
```
ximon18, line 157: When you say "we do not migrate" do you mean "we keep without modifiction" ?

src/upgrades/pre_0_14_0/old_commands.rs
```diff
@@ -133,7 +133,7 @@ pub enum Pre0_14_0CertAuthStorableCommand {
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct Pre0_14_0AspaDefinitionUpdates {
     add_or_replace: Vec<Pre0_14_0AspaDefinition>,
-    remove: Vec<Pre0_14_0ProviderAs>, // was using string notation
+    remove: Vec<rpki::resources::Asn>,
```
ximon18, line 136: What is this change?

timbru: It was a bug. Krill > 0.10.0 use a u32 here and not an ASN string, as 0.9.x did

timbru: our regression test code did not include any "delete ASPA" (by "remove") in its history.

src/upgrades/pre_0_14_0/cas_migration.rs
timbru: I mean that these events don't need to be kept. We track the ASPA stuff separately and add back ASPA configs in the end - and re-issue ASPA objects then.

timbru:
@ximon18 improved the tests and fixed an issue so that 0.9.6 can be upgraded. Incidentally, 0.9.6 cannot be upgraded to 0.10.x if it has ASPA. But it can be upgraded to 0.14.3 (when this is merged and released).

ximon18:
> @ximon18 improved the tests and fixed an issue so that 0.9.6 can be upgraded. Incidentally, 0.9.6 cannot be upgraded to 0.10.x if it has ASPA. But it can be upgraded to 0.14.3 (when this is merged and released).

When you say "cannot" - what happens if someone tries?

timbru:
> > @ximon18 improved the tests and fixed an issue so that 0.9.6 can be upgraded. Incidentally, 0.9.6 cannot be upgraded to 0.10.x if it has ASPA. But it can be upgraded to 0.14.3 (when this is merged and released).
> 
> When you say "cannot" - what happens if someone tries? And what happens if they try upgrading to 0.11.x instead of 0.10.x? Is it everything between 0.10.x and 0.14.2 inclusive that won't work in some way?

If people used ASPA in 0.9.x, they can only upgrade to 0.14.3. This was not working in past versions.

If they did not use ASPA, then they could have updated to 0.10.x and above.

I tested that upgrades with ASPA from 0.10.3 to 0.11.0 to 0.12.1 to 0.13.1 work, and used the data dir after each of thos upgrades to test that upgrading to 0.14.3 will work. I can test that on the test machine when I have a package for this PR.

ximon18 (approved):

PR #1164, opened 2023-11-28
Ignore corrupt status files and use default.

timbru:


src/daemon/ca/status.rs
```diff
@@ -96,7 +96,10 @@ impl StatusStore {
     /// so any missing, corrupted, or no longer supported data format - can be ignored.
     /// It will get updated with new status values as Krill is running.
     fn load_full_status(&self, ca: &CaHandle) -> KrillResult<()> {
-        let repo: RepoStatus = self.store.get(&Self::repo_status_key(ca))?.unwrap_or_default();
+        let repo: RepoStatus = match self.store.get(&Self::repo_status_key(ca)) {
```
ximon18, line 99: I can live with this pattern though it gets repetitive given that you do it more than once.

An alternative syntax for this could be (but still requires that the pattern be repeated on each use):
```
let repo: RepoStatus = self.store.get(&Self::repo_status_key(ca))
    .unwrap_or(None)
    .unwrap_or_default();
```

It might be better to just add a `get_or_default()` fn to the store and use that.

ximon18: Wow even uglier would be calling `unwrap_or_default()` twice in a chain... ;-)

ximon18: I wonder, is there an issue here when there is nothing wrong with the data on disk or in whatever backend KV store, but there's just a temporary issue accessing it? Then you'd proceed anyway ignoring the good but temporarily unavailable state rather than reporting the access issue - would it be good to at least warn that the underlying "execute()" call failed?

ximon18: I suppose if there is a problem accessing the underlying store then failure is going to happen soon anyway...

ximon18 (approved):

timbru:
Commit cherry-picked into PR #1167. Will close this, and merge that.

PR #1166, opened 2023-11-30
Update dependencies and code to allow using rpki-rs 0.18.0

timbru:


timbru:
openssl was not updated actually, I propose to squash this all when merging.

src/lib.rs
```diff
@@ -17,6 +17,7 @@ extern crate reqwest;
 extern crate rpki;
```
partim, line 17: I don’t think you need any of the `extern crate` things any more.

partim (approved):

src/lib.rs
timbru: This was cleaned up. The `#[macro_use]` instances were kept for now, as removing them would require a lot of fixes to use statements. So, it's better to do this when there are no big open PRs.

PR #1178, opened 2023-12-19
Allow overriding the TA manifest number.

timbru:
Allows overriding the initial manifest number, either by specifying `--initial_manifest_number` if the CLI is used to initialise the TA signer, or by including `ta_mft_nr_override: #nr` in the ImportTa JSON.

Allows overriding the TA manifest number when signing a TA proxy request by specifying `--ta_mft_number_override` in the CLI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .cargo
 .idea/
+.locks
 .vscode/
 .DS_Store
 data/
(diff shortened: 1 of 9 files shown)


ximon18 (approved):
This has been used and no issues were reported.

Issue #1177, opened 2023-12-18
Login ajax request taking forever / timeout


straussmarkus:
Hi,

after upgrading to 0.14.4 on Debian 12, i have a weird issue, where logging in (config auth) takes forever (load indicator at top of page spinning)

In dev tools i can see that it waits for an ajax request to:
https://myurl.com/api/v1/cas/MY-CA/routes/analysis/full

Which never retourns and then eventually my reverse proxy times out for this request.

Is there anything i can do about that?

timbru:
Is this still blocking?

It looks like it might have been stuck waiting for the BGP Ris files to be downloaded. @ximon18 I think the write lock in this code should be moved until after the download happens:
https://github.com/NLnetLabs/krill/blob/46feb1e64a29e5bfe7b3db91fb0227602011522e/src/commons/bgp/analyser.rs#L47

straussmarkus:
I still have that problem. Restarting the krill instance may solve it sometimes. But on next login it happens again

timbru:
Do you have a large amount of ROAs configured, or do you have many announcements for the space on your certificate? I am wondering why you see this issue and others do not.

If it's the downloading BGP RIS dumps that is causing this, then you could try disabling the BGP preview functionality in your config file by setting ``bgp_risdumps_enabled = false``. You would no longer have ROA suggestions. I.e. you only see your actual configurations and no information about announcements.

If this improves things for you then we are one step closer to knowing the cause.

straussmarkus:
I have 4 parent CAs and 3 ROAs, so not mu h.I set  bgp_risdumps_enabled = false and it seems to work at the moment, but i will see tomorrow as it always took some time before the problem occuredMit freundlichen GrüßenIng. Markus Strauß MSc-------- Ursprüngliche Nachricht --------Von: Tim Bruijnzeels ***@***.***> Datum: 20.12.23  09:51  (GMT+01:00) An: NLnetLabs/krill ***@***.***> Cc: "Ing. Markus Strauß MSc" ***@***.***>, Author ***@***.***> Betreff: Re: [NLnetLabs/krill] Login ajax request taking forever / timeout (Issue #1177) 

Do you have a large amount of ROAs configured, or do you have many announcements for the space on your certificate? I am wondering why you see this issue and others do not.
If it's the downloading BGP RIS dumps that is causing this, then you could try disabling the BGP preview functionality in your config file by setting
[...]

PR #1179, opened 2023-12-21
Fix locking when downloading RIS dump files (#1177)

timbru:
This should fix the issue observed in #1177.

The code locked the 'seen' announcements while downloading new RIR dump files. This prevented read access to the announcement during download. This change postpones getting the write lock until after the download is completed, so it only keeps it for the shortest possible time.


ximon18 (approved):
LGTM

Issue #1180, opened 2023-12-22
rrdp content not matching XML schema


ties:
One of my tools fails to parse RRDP content generated by what I expect is krill. One example delta from nicbr their repository:
```
<delta xmlns="http://www.ripe.net/rpki/rrdp" version="1" session_id="aa4bfebe-4ea2-44f5-9da7-9492753e8e98" serial="285013">
</delta>
```

The relaxng schema requires at least one publish or withdraw element
```relaxng
# Delta segment: think DNS IXFR.

start |= element delta {
  attribute version    { version },
  attribute session_id { uuid },
  attribute serial     { serial },
  delta_element+
}

delta_element |= element publish  {
  attribute uri  { uri },
  attribute hash { hash }?,
  base64
}

delta_element |= element withdraw {
  attribute uri  { uri },
  attribute hash { hash }
}
```

timbru:
Thank you for reporting this!

I am not quite sure how this happens. As far as I can tell Krill CAs will not send empty RFC 8181 publication deltas to the server, but if this should happen then the Publication Server could produce an empty RRDP delta. I created a PR with a fix for this, but merging and releasing will have to wait until after the holidays at least.

ties:
By now we see this happening in a test environment where we have automated (API-driven) actions:
```
2024-01-26T13:14:41.362521000Z DEBUG:rpkiclientweb.rpki_client.stderr:rpki-client: https://rrdp.paas.rpki.prepdev.ripe.net/notification.xml: pulling from network
2024-01-26T13:14:41.362550000Z DEBUG:rpkiclientweb.rpki_client.stderr:rpki-client: https://rrdp.paas.rpki.prepdev.ripe.net/notification.xml: downloading 3 deltas (f610c790-8ceb-4711-bddd-938d0b16fa39#54425)
2024-01-26T13:14:41.362583000Z DEBUG:rpkiclientweb.rpki_client.stderr:rpki-client: parse failed - empty delta
2024-01-26T13:14:41.362616000Z DEBUG:rpkiclientweb.rpki_client.stderr:rpki-client: https://rrdp.paas.rpki.prepdev.ripe.net/notification.xml: parse error at line 2: parsing aborted
2024-01-26T13:14:41.362652000Z DEBUG:rpkiclientweb.rpki_client.stderr:rpki-client: https://rrdp.paas.rpki.prepdev.ripe.net/notification.xml: delta sync failed, fallback to snapshot
[...]

PR #1181, opened 2023-12-27
Don't create an RRDP delta for publishers entries without content #1180

timbru:


timbru:
This should prevent that empty deltas are produced, both by not staging empty publication (RFC8181) deltas for publisher and skipping them in the decision process to decide whether to produce a new RRDP delta.

As far as I can tell Krill CAs [do not send empty publication deltas ](https://github.com/NLnetLabs/krill/blob/46feb1e64a29e5bfe7b3db91fb0227602011522e/src/daemon/ca/manager.rs#L2052) but issue #1180 shows that empty RRDP deltas were produced. Possibly because non Krill publishers sent them.

ximon18:
Notes to self:
- Ideally I think this PR should include a test that shows that the issue occurs and is fixed.
- Given that the RFC defines an XML schema to validate against, do the Krill tests include anywhere XML schema compliance validation and if not, should they? (though perhaps not as part of this PR)

timbru:
> Notes to self:
> 
> * Ideally I think this PR should include a test that shows that the issue occurs and is fixed.

Ideally, yes. But it was not trivial to do this...

There are tests in place (under the tests directory) that have CAs publish at a Krill Server. However, Krill CAs do not send empty publication requests. Setting up an intentionally broken (or well, at least different) publishing CA is not trivial. Unit testing at the current function level is even less trivial because it would require a lot of set up and new code.

Extracting `if self.staged_elements.values().any(|el| !el.0.is_empty()) {` to a function could allow for some unit testing of that function with specific instances of `self`, but when implementing this I felt that was taking things a bit too far, as the test then because very tied to the implementation and become rather trivial.

Anyway... not saying any of this in protest against adding a test. Just trying to explain why there isn't one.

> * Given that the RFC defines an XML schema to validate against, do the Krill tests include anywhere XML schema compliance validation and if not, should they? (though perhaps not as part of this PR)

I am not sure, this kind of leads to formal schema validation in the XML library - which is not supported. Maybe, but it would need case by case evaluation. In this particular case it would not help as the internal state with an empty delta should be prevented. The XML is just a representation of that that state. So failing the XML generation would not help, and could even lead to irrecoverable issues (theoretically).

ximon18:
1. Thanks for the insights, appreciated.

2. Re XML validation I was referring more generally to showing that the XML produced is RFC compliant, not that XML representation is the right level to test this change at.

partim (approved):

PR #1196, opened 2024-05-03
Make cryptoki-sys optional

botovq:
cryptoki itself is marked optional and only needed for hsm builds. cryptoki-sys is still compiled unconditionally, also in builds disabling the hsm feature. This causes build failures if cryptoki isn't installed or on systems unsupported by cryptoki-sys's build script.


partim:
Thank you for the PR and sorry for the long silence!

partim (approved):

PR #1199, opened 2024-06-11
Remove deprecated chrono functions.

partim:
This PR replaces a now deprecated use of _chrono_ with its current equivalent.


ximon18 (approved):
LGTM.

PR #1211, opened 2024-06-18
Update hyper to 1.3.1 and reqwest to 0.12.5.

partim:
This PR updates hyper to 1.3.1 and reqwest to 0.12.4 and rearranges the HTTP server to the changes in hyper 1.0.


src/daemon/http/mod.rs
```diff
@@ -433,67 +452,23 @@ impl Request {
...
 
-    /// See hyper::body::to_bytes
-    ///
-    /// Here we want to limit the bytes consumed to a maximum. So, the
-    /// code below is adapted from the method in the hyper crate.
     pub async fn read_bytes(self, limit: u64) -> Result<Bytes, Error> {
-        let body = self.request.into_body();
-
-        futures_util::pin_mut!(body);
+        // We’re going to cheat a bit. If we know the body is too big from
```
ximon18, line 456: It's not clear to me if this code, or even the old code, actually prevents reading too much data into memory from the network.

src/daemon/http/tls.rs
```diff
@@ -1,322 +1,380 @@
...
-        let mut config = ServerConfig::builder()
-            .with_safe_default_cipher_suites()
-            .with_safe_default_kx_groups()
-            .with_safe_default_protocol_versions()
-            .map_err(TlsConfigError::Rustls)?
-            .with_no_client_auth()
-            .with_single_cert(cert, tokio_rustls::rustls::PrivateKey(key))
-            .map_err(TlsConfigError::Rustls)?;
-
-        // See: https://wiki.wireshark.org/TLS#tls-decryption
```
ximon18, line 185: Is there a reason that this comment wasn't carried over to the new code?

src/daemon/http/tls.rs
```diff
@@ -1,322 +1,380 @@
...
+        Accept { #[pin] fut: Accept<TcpStream> },
+
+        /// We have a working TLS stream.
+        Stream { #[pin] fut: TlsStream<TcpStream> },
+
+        /// TLS handshake has failed.
+        ///
+        /// Because hyper still wants to do a clean flush and shutdown, we
+        /// need to still work in this state. For read and write, we just
+        /// keep returning the clean shutdown indiciation of zero length
```
ximon18, line 129:
```suggestion
        /// keep returning the clean shutdown indication of zero length
```

src/daemon/http/tls.rs
```diff
@@ -1,322 +1,380 @@
...
-
-impl std::fmt::Display for TlsConfigError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            TlsConfigError::Io(err) => err.fmt(f),
-            TlsConfigError::Pkcs8ParseError => write!(f, "pkcs8 parse error"),
-            TlsConfigError::RsaParseError => write!(f, "rsa parse error"),
-            TlsConfigError::EmptyKey => write!(f, "key contains no private key"),
-            TlsConfigError::Rustls(err) => write!(f, "rustls error: {}", err),
+/// Reads the first private key from the given PEM file.
```
ximon18, line 60: "First" here conflicts with "if there isn't exactly one private key" below.

src/daemon/http/tls.rs
```diff
@@ -1,322 +1,380 @@
...
+///
+/// The key may be a PKCS#1 RSA private key, a PKCS#8 private key, or a
+/// SEC1 encoded EC private key. All other PEM items are ignored.
+///
+/// Errors out if opening or reading the file fails or if there isn’t exactly
```
ximon18, line 65: The while loop seems to try to the next item if the current one is not supported... so this comment is incorrect I think?

src/daemon/http/tls.rs
```diff
@@ -1,322 +1,380 @@
...
+                        Poll::Ready(Err(err))
+                    }
                 }
             }
-        };
-
-        let mut config = ServerConfig::builder()
-            .with_safe_default_cipher_suites()
-            .with_safe_default_kx_groups()
-            .with_safe_default_protocol_versions()
```
ximon18, line 179: Is there a reason that `.with_safe_default_protocol_versions()` wasn't carried over to the new code?

ximon18 (approved):
Added some comments but looks mostly fine.

src/daemon/http/mod.rs
partim: My understanding is that Hyper will only start reading the body when you explicitly tell it to. The first test here basically looks at the Content-Length header and rejects the request if that’s too large before even beginning to read the body. The new `Limited` thing is only relevant for chunked transfers.

src/daemon/http/tls.rs
partim: Nope. I added it back in again.

src/daemon/http/tls.rs
partim: Changed it to “… if there isn’t exactly one _recognized_ private key in the file.”

src/daemon/http/tls.rs
partim: `ServerConfig::builder` now returns a builder with all these things already set.

Issue #901, opened 2022-08-24
Simplify PKCS#11 "is initialized" logic


ximon18:
By using the new `is_initialized()` function provided by the soon to be released next version of the `cryptoki` crate.

See: https://github.com/parallaxsecond/rust-cryptoki/issues/77

ximon18:
FYI [crpytoki 0.4.0](https://github.com/parallaxsecond/rust-cryptoki/blob/main/CHANGELOG.md#cryptoki-040-2022-09-07) has just been released which adds a `Pkcs11::is_initialized` function call which we can use to implement this feature.

Issue #902, opened 2022-08-24
Check supported PKCS#11 capabilities


ximon18:
The KMIP signer tests the supported capabilities of the KMIP provider to see if it supports the minimum required set of operations. The PKCS#11 signer doesn't do this because it wasn't possible with the old `pkcs11` crate nor with the new `cryptoki` crate. However, the soon to be released new version of `cryptoki` makes this possible.

Note: If implementing this we should also implement a 'force' flag as exists for the KMIP signer, in case a provider doesn't claim support for a particular operation but does actually support it.

See: https://github.com/parallaxsecond/rust-cryptoki/issues/78

ximon18:
FYI [crpytoki 0.4.0](https://github.com/parallaxsecond/rust-cryptoki/blob/main/CHANGELOG.md#cryptoki-040-2022-09-07) has just been released which adds a `Pkcs11::is_fn_supported()` function call which we can use to implement this feature.

PR #1212, opened 2024-06-19
Upgrade cryptoki to 0.7.

partim:
This PR upgrades cryptoki – the library for PKCS11 support – to version 0.7. This requires a few minor changes related to handling of user PINs.


Cargo.toml
```diff
@@ -27,8 +27,8 @@ basic-cookies   = { version = "^0.1", optional = true }
 bytes           = "1"
 chrono          = { version = "^0.4", features = ["serde"] }
 clap            = "^2.33"
-cryptoki        = { version = "^0.3", optional = true }
-cryptoki-sys    = { version = "=0.1.4", optional = true } # pin cryptoki-sys because of compilation issues on various systems
+cryptoki        = { version = "0.7", optional = true}
+#cryptoki-sys    = { version = "=0.1.4", optional = true } # pin cryptoki-sys because of compilation issues on various systems
```
ximon18, line 31: Why not delete this line rather than comment it out?

src/commons/crypto/signing/signers/pkcs11/signer.rs
```diff
@@ -452,7 +469,7 @@ impl Pkcs11Signer {
     ) -> Result<UsableServerState, ProbeError<SignerError>> {
         fn slot_label_eq(ctx: &RwLockReadGuard<Pkcs11Context>, slot: Slot, slot_label: &str) -> bool {
             match ctx.get_token_info(slot) {
-                Ok(info) => String::from_utf8_lossy(&info.label).trim_end() == slot_label,
+                Ok(info) => info.label().trim_end() == slot_label,
```
ximon18, line 472: It looks to me like this change potentially introduces a new panic, if the underlying token info contains a non-UTF8 compliant value. Does the `label()` call ensure that that is not possible? This change doesn't seem to be related to upgrading cryptoki or did I miss something?

ximon18 (approved):
LGTM, but FYI it doesn't take advantage of the changes in v0.4 of Cryptoki to resolve #901 and #902 (neither of which are critical).

Cargo.toml
partim: That is a very good question!

partim:
Let’s keep this PR strictly the upgrade and then do the issues in separate PRs.

PR #1228, opened 2024-07-25
Refactor cli to use clap’s derive.

partim:
This PR changes how the clients – `krillc`, `krillta`, as well as the integration tests – work to better fit the derive model provided by _clap_. This results in basically everything in the `cli` module and all the integration tests being different now.

The PR slightly changes the options for both `krillc` and `krillta`. For `krillc`, the `--server`, `--token`, `--format`, and `--api` options are now before the first subcommand (since they affect all commands). For `krillta`, those options are now after `krillta proxy` but before the next subcommand, while `--format` is now after `krillta signer`.

This PR also removes client support and integration tests for RTA.

This is a breaking change.


src/cli/ta/options/signer.rs
```diff
@@ -0,0 +1,307 @@
...
+
+//------------ Command -------------------------------------------------------
+
+#[derive(clap::Args)]
+pub struct Command {
+    /// Path to config file.
+    #[arg(
+        long, short,
+        value_name = "path",
+        default_value = "/etc/krillta.conf"
```
Koenvh1, line 27: Do we want to make this a constant like `KRILL_DEFAULT_CONFIG_FILE`?

ximon18:
I see that the new `krillc --help` behaviour no longer shows the version in the first line but instead a description string.

Also I see that the new code sorts command line subcommands alphabetically, whereas before they were grouped and ordered deliberately.

Are these deliberate changes?

ximon18:
I note that error messages no longer contain tips, e.g.

Old:
```
Missing argument: --token, alternatively you may use env var: KRILL_CLI_TOKEN
```

New:
```
error: the following required arguments were not provided:
  --token <TOKEN>
```

Is this a deliberate change?

ximon18:
The text printed for `krillc info --help` no longer shows relevant command line arguments:

Old:
```
krillc-info
Show server info

USAGE:
    krillc info [FLAGS] [OPTIONS]

FLAGS:
        --api        Only show the API call and exit. Or set env: KRILL_CLI_API=1
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -f, --format <type>     Report format: none|json|text (default). Or set env: KRILL_CLI_FORMAT
    -s, --server <URI>      The full URI to the Krill server. Or set env: KRILL_CLI_SERVER
    -t, --token <string>    The secret token for the Krill server. Or set env: KRILL_CLI_TOKEN
```

New:
```
Show server info

Usage: krillc --token <TOKEN> info

Options:
  -h, --help  Print help
```

The purpose of the command is no longer shown.
Various command line arguments are no longer shown, with `--server` and `--api` being very relevant and now missing.

ximon18:
One other thing I don't get, though I'm not sure that this is caused by this PR, is that the `/health` endpoint is an unprotected endpoint, so that load balancers for example can query it without needing credentials, and this works localhost, e.g.:

```
root@krill:~# wget --no-check-certificate -qSO- https://127.0.0.1:3000/health
  HTTP/1.1 200 OK
  content-type: text/plain
  content-length: 0
  date: Thu, 15 Aug 2024 13:19:17 GMT
```

But it doesn't work via the `krillc` command, either with an earlier version or this version:

Old:
```
root@krill:~# krillc -V
Krill Client 0.14.4
root@krill:~# krillc health
Missing argument: --token, alternatively you may use env var: KRILL_CLI_TOKEN
```

New:
```
$ target/release/krillc health -s https://krill.nlnetlabs.nl/
Missing argument: --token, alternatively you may use env var: KRILL_CLI_TOKEN
```

ximon18:
Another oddity, problem user error on my part:

Old:
```
$ krillc --help
...
SUBCOMMANDS:
    add          Add a new CA
    aspas        Manage ASPAs for a CA (experimental)
...
```

The code in this branch doesn't report `aspas`, and I see that the `aspa` feature is gone from `Cargo.toml` but that the `rta` feature remains (even though this PR description says RTA is removed) ??

ximon18:
There seems to be an exit code regression:

Old:
```
$ krillc roas list --ca nlnetlabs
2a04:b907::/48-48 => 0 # Invalid more specific at Coloclue, valid less specific at Vultr
...
$ echo $?
0
```

New:
```
$ target/release/krillc roas list --ca nlnetlabs
2a04:b907::/48-48 => 0 # Invalid more specific at Coloclue, valid less specific at Vultr
...
$ echo $?
1
```

src/bin/krillup.rs
```diff
@@ -1,89 +1,161 @@
...
+                Ok(mgr) => mgr,
+                Err(e) => {
+                    eprintln!("*** Error Preparing Data Migration ***");
+                    eprintln!("{}", e);
+                    eprintln!();
+                    eprintln!(
+                        "Note that your server data has NOT been modified. \
+                         Do not upgrade krill"
+                    );
+                    eprintln!("itself yet!");
```
ximon18, line 50: This looks a bit strange.

src/bin/krillup.rs
```diff
@@ -1,89 +1,161 @@
...
+
+
+//------------ Command -------------------------------------------------------
+
+#[derive(clap::Subcommand)]
+pub enum Command {
+    /// Prepare a migration, leave current data unmodified
+    Prepare(Prepare),
+
+    /// Migrate data to different storage. Stop Krill before use!
```
ximon18, line 138: What do you mean here by "different storage"? Migration is about upgrading the stored content to the new serialized format, right?

src/bin/krillup.rs
```diff
@@ -1,89 +1,161 @@
...
+//------------ Migrate -------------------------------------------------------
+
+#[derive(clap::Args)]
+pub struct Migrate {
+    /// The storage target as a URI string.
+    #[arg(short, long, value_name = "URI")]
+    pub target: Url,
+}
+
+/*
```
ximon18, line 158: Why leave this commented out code here?

src/cli/options/aspa.rs
```diff
@@ -0,0 +1,192 @@
...
+
+
+//------------ Command -------------------------------------------------------
+
+#[derive(clap::Subcommand)]
+pub enum Command {
+    /// Show current ASPAs
+    List(List),
+
+    /// Add or replace an ASPA.
```
ximon18, line 20: Nit: Inconsistent use of trailing period ".".

src/cli/options/aspa.rs
```diff
@@ -0,0 +1,192 @@
...
+
+
+//------------ Add -----------------------------------------------------------
+
+#[derive(clap::Parser)]
+pub struct Add {
+    #[command(flatten)]
+    ca: ca::Handle,
+
+    /// the ASPA formatted like: 65000 => 65001, 65002, 65003
```
ximon18, line 66: Nit: Comment starting with lowercase instead of initial capital letter.

src/cli/options/aspa.rs
```diff
@@ -0,0 +1,192 @@
...
+    CustomerAsProvider,
+    EmptyProviders,
+}
+
+impl fmt::Display for CleanAspaDefinitionError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::Format(err) => err.fmt(f),
+            Self::CustomerAsProvider => {
+                f.write_str("Customer AS may not be used as provider.")
```
ximon18, line 182: Higher up in this file the references all seem to refer to "ASN" but here the term "AS" is used. Is this correct?

src/cli/options/aspa.rs
```diff
@@ -0,0 +1,192 @@
...
+pub struct Update {
+    #[command(flatten)]
+    ca: ca::Handle,
+
+    /// Customer ASN of an existing ASPA
+    #[arg(long, value_name = "ASN")]
+    customer: Asn,
+
+    /// Provider ASN to add
+    #[arg(long, value_name = "ASN")]
```
ximon18, line 119: Should this perhaps be "ASN(s)" ?

src/cli/options/aspa.rs
```diff
@@ -0,0 +1,192 @@
...
+    add: Vec<Asn>,
+
+    /// Provider ASN to remove.
+    #[arg(long, value_name = "ASN")]
```
ximon18, line 123: Should this perhaps be "ASN(s)" ?

src/cli/options/bgpsec.rs
```diff
@@ -0,0 +1,176 @@
...
+#[derive(clap::Parser)]
+pub struct Add {
+    #[command(flatten)]
+    ca: ca::Handle,
+
+    /// The ASN to authorize the router key for
+    #[arg(long, short, value_name = "ASN")]
+    asn: Asn,
+
+    /// Path to the DER-encoded certificate signing request.
```
ximon18, line 70: Nit: Inconsistent trailing period ".".

src/cli/options/bgpsec.rs
```diff
@@ -0,0 +1,176 @@
...
+pub struct Remove {
+    #[command(flatten)]
+    ca: ca::Handle,
+
+    /// The ASN of router key to be removed
+    #[arg(long, short, value_name = "ASN")]
+    asn: Asn,
+
+    /// The hex encoded key identifier of the router key
+    #[arg(long, value_name = "key identifier")]
```
ximon18, line 102: Nit: Inconsistent case: "ASN" starts with a capital letter while "key identifier" does not.

src/cli/options/bulk.rs
```diff
@@ -0,0 +1,76 @@
+//! Trigger events for all CAs.
```
ximon18, line 1: These are commands, not trigger events, right?

src/cli/options/ca.rs
```diff
@@ -0,0 +1,216 @@
...
+
+//-------- HistoryCommands ---------------------------------------------------
+
+#[derive(clap::Parser)]
+pub struct HistoryCommands {
+    #[command(flatten)]
+    ca: Handle,
+
+    /// Number of rows (max 250)
+    #[arg(long, value_name = "number")]
```
ximon18, line 115: Nit: Inconsistent case: "number" starts with a lowercase letter while "RFC" starts with an uppercase letter.

src/cli/options/children.rs
```diff
@@ -0,0 +1,335 @@
...
+    /// Remove an existing child from a CA
+    Remove(Remove),
+
+    /// Show the RFC 8183 Parent Response XML
+    Response(Response),
+
+    /// Show connections stats for children of a CA
+    Connections(Connections),
+
+    /// Suspend a child CA: hide certificate(s) issued to child
```
ximon18, line 43: What does it mean to "hide" here?

src/cli/options/children.rs
```diff
@@ -0,0 +1,335 @@
...
+//------------ Handle --------------------------------------------------------
+
+#[derive(clap::Args)]
+struct Handle {
+    /// Name of the CA to control
+    #[arg(long, short, env = "KRILL_CLI_MY_CA")]
+    ca: CaHandle,
+
+    /// The name of the child CA you wish to control
+    #[arg(long, value_name = "name")]
```
ximon18, line 75: Nit: Inconsistent case: "name" starts with a lowercase letter but "RFC" (elsewhere) starts with an uppercase letter.

src/cli/options/children.rs
```diff
@@ -0,0 +1,335 @@
...
+#[derive(clap::Parser)]
+pub struct Add {
+    #[command(flatten)]
+    handle: Handle,
+
+    #[command(flatten)]
+    resources: ChildResources,
+
+    /// Path to the RFC 8183 Child Request XML file
+    #[arg(long, short, value_name = "path")]
```
ximon18, line 91: Nit: Inconsistent case: "path" starts with a lowercase letter but "RFC" (elsewhere) starts with an uppercase letter.

src/cli/options/children.rs
```diff
@@ -0,0 +1,335 @@
...
+#[derive(clap::Parser)]
+pub struct Update {
+    #[command(flatten)]
+    handle: Handle,
+
+    #[command(flatten)]
+    resources: ChildResources,
+
+    /// Path to the RFC 8183 Child Request XML file
+    #[arg(long, short, value_name = "path")]
```
ximon18, line 135: Nit: Inconsistent case: "path" starts with a lowercase letter but "RFC" (elsewhere) starts with an uppercase letter.

src/cli/options/children.rs
```diff
@@ -0,0 +1,335 @@
...
+}
+
+
+//------------ Connections ---------------------------------------------------
+
+#[derive(clap::Args)]
+pub struct Connections {
+    /// Name of the CA to control
+    #[arg(long, short, env = "KRILL_CLI_MY_CA")]
+    ca: CaHandle,
```
ximon18, line 271: Why not use ca::Handle here?

src/cli/options/config.rs
```diff
@@ -0,0 +1,284 @@
...
+
+//-------- Simple --------------------------------------------------------
+
+#[derive(clap::Parser)]
+pub struct Simple {
+    /// Override the default data directory of ./data.
+    #[arg(short, long)]
+    pub data: Option<String>,
+
+    /// Override the default log file path of ./krill.log.
```
ximon18, line 39: Nit: Inconsistent use of trailing period ".".

src/cli/options/config.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+}
+
+
+//-------- User ----------------------------------------------------------
+
+#[cfg(feature = "multi-user")]
+#[derive(clap::Parser)]
+pub struct User {
+    /// ID (e.g., username, email) to generate configuration for.
```
ximon18, line 129: Nit: Inconsistent use of trailing period ".".

src/cli/options/config.rs
```diff
@@ -0,0 +1,284 @@
...
+    #[arg(long, value_name = "id")]
+    pub id: String,
+
+    /// Attributes for the user.
```
ximon18, line 133: Nit: Inconsistent use of trailing period ".".

src/cli/options/config.rs
```diff
@@ -0,0 +1,284 @@
+use std::fmt;
```
ximon18, line 1: Nit: Missing module comment?

ximon18 (commented):
I've done a partial review, will continue reviewing next week.

src/cli/client.rs
```diff
@@ -1,985 +1,886 @@
...
+    }
 
-                Ok(ApiResponse::BgpAnalysisSuggestions(suggestions))
-            }
+    pub async fn child_import(
+        &self, ca: &CaHandle, import: api::import::ImportChild
+    ) -> Result<api::Success, Error> {
+        self.post_json(
+            ca_path(ca).into_iter().chain(
+                [
```
ximon18, line 337: Are there missing `/` separators here?

src/cli/client.rs
```diff
@@ -1,985 +1,886 @@
...
-        let (password_hash, salt) = {
-            use scrypt::scrypt;
+    pub async fn aspas_update_single(
+        &self,
+        ca: &CaHandle, customer: Asn,
+        update: api::AspaProvidersUpdate
+    ) -> Result<api::Success, Error> {
+        self.post_json(
+            ca_path(ca).into_iter().chain(
+                [ "aspas".into(), "as".into(), customer.to_string().into()]),
```
ximon18, line 588: Are there missing `/` separators here?

src/cli/client.rs
```diff
@@ -1,985 +1,886 @@
...
-            //   "TL;DR - either only allow ASCII characters in passwords, or
-            // use            String.prototype.normalize('NFKC')
-            // on any password" So in Lagosta we do the NFKC
-            // normalization and thus we need to do the same here.
-            use unicode_normalization::UnicodeNormalization;
+    pub async fn publishers_stale(
+        &self, seconds: u64,
+    ) -> Result<api::PublisherList, Error> {
+        self.get_json(
+            ["api/v1/pubd/stale".into(), seconds.to_string().into()]
```
ximon18, line 603: Are there missing `/` separators here?

src/commons/api/misc.rs
```diff
@@ -0,0 +1,31 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("Ok")
+    }
+}
+
+impl Serialize for Success {
+    fn serialize<S: Serializer>(
+        &self, serializer: S
+    ) -> Result<S::Ok, S::Error> {
+        let mut serializer = serializer.serialize_struct( "UnitSuccess", 1)?;
```
ximon18, line 26: Why "UnitSuccess" and not "Success"?

tests/auth_check.rs
```diff
@@ -1,39 +1,32 @@
...
-    let mut config =
-        test_config(&storage_uri, None, false, false, false, false);
-    config.admin_token = Token::from("wrong secret");
+    let (server, _tempdir) = common::KrillServer::start().await;
 
-    // Start Krill with the customized config
-    start_krill_with_custom_config(config).await;
+    // Get a client and change its auth token.
+    let mut client = server.client().clone();
+    client.set_token("different token".into());
```
ximon18, line 16: "different" doesn't imply "wrong" to me, perhaps replace "different token" with "deliberately incorrect token"?

tests/client_coverage.rs
```diff
@@ -0,0 +1,160 @@
...
+    server.client().roas_suggest(&child, Some(ca_res.clone())).await.unwrap();
+
+    server.client().bgpsec_add_single(
+        &child, common::asn("AS65000"),
+        rpki::ca::csr::BgpsecCsr::decode(
+            include_bytes!("../test-resources/bgpsec/router-csr.der").as_ref()
+        ).unwrap(),
+    ).await.unwrap();
+    server.client().bgpsec_list(&child).await.unwrap();
+
```
ximon18, line 125: Is `server.client().bgpsec_delete_single()` skipped because it's a variant of `bgpsec_add_single()` in that they both call `bgpsec_update()` ?

src/ta/mod.rs
```diff
@@ -50,7 +50,7 @@ mod tests {
     #[test]
     fn init_ta() {
         test::test_in_memory(|storage_uri| {
-            let cleanup = test::init_logging();
+            let _ = stderrlog::new().verbosity(5).init();
```
ximon18, line 53: Where does 5 come from?

ximon18 (approved):
I played with the CLI in a shell, I read most if not all of the actual changed client/server code, and I scanned the modified tests to see if they looked structurally roughly the same as the originals. I did **NOT** try and work out if the `encode()` HTTP path percent encoding is correct as I need more coffee or something before I can make sense of RFC 3986. I'll approve this PR with the assumption that the comments I made will be checked. Nice cleanup by the way!

partim:
Regarding the changes to the help output: I suppose they are the price for using the derive model. We could probably try and work around some of them, but I’m not sure whether the _krillc_ is used widely enough to make this worthwhile.

partim:
> Why does `health` contact the `/api/v1/authorized` endpoint rather than the `/health` endpoint?

Not sure. I will leave it like it is for now, though.

ximon18:
> > Why does `health` contact the `/api/v1/authorized` endpoint rather than the `/health` endpoint?
> 
> Not sure. I will leave it like it is for now, though.

This was my bad, this behaviour wasn't introduced by this PR.

partim:
> There seems to be an exit code regression:

Nice find! Someone mixed up the two cases …

src/bin/krillup.rs
partim: Yeah, but it is sort of correct (if you want to have each line in its own println command).

src/bin/krillup.rs
partim: I think this is also intended to later move data into a database? In any case, this is basically just taken from the old code, so I want to leave it.

src/cli/options/bgpsec.rs
partim: The value names are intended to be in regular case, i.e., starting small. But ASN is an abbreviation and hence in capitals.

src/cli/report.rs
```diff
@@ -1,317 +1,163 @@
...
         match format {
-            ReportFormat::None => Ok("".to_string()),
-            ReportFormat::Json => Ok(self.json()),
-            ReportFormat::Text => self.text(),
+            ReportFormat::None => { Ok(()) }
+            ReportFormat::Json => {
+                // The &mut here seems to be necessary to avoid a move into
+                // the function.
+                serde_json::to_writer_pretty(&mut *target, self)?;
+                target.write_all(b"\n")?;
```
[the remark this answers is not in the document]
partim: This is a fix of sorts – the missing line break caused the shell prompt to just continue the last line of output.

src/cli/ta/options/proxy.rs
```diff
@@ -0,0 +1,451 @@
...
+            }
+            Self::Parse(path, err) => {
+                write!(
+                    f, "Failed to parse child info file '{}': {}'",
+                    path, err
+                )
+            }
+            Self::Cert(path, err) => {
+                write!(
+                    f, "Failed to parse child info file '{}': {}'",
```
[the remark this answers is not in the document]
partim: I think “parse” is appropriate here. The really helpful information should come from displaying `err`.

src/commons/api/misc.rs
partim: That’s probably leftover from an earlier name of the struct. I’ll fix it.

tests/auth_check.rs
partim: I’ll change it to “wrong secret”.

tests/client_coverage.rs
partim: Let’s go with “Yes” ;)

(I’ll add a comment.)

src/ta/mod.rs
partim: I suppose from me misreading the documentation and thinking “higher than 4” is for trace rather than “4 or higher.” Looking at the source, they same to use `LevelFilter` internally, so I’ll change it to that.

Issue #1229, opened 2024-08-06
Replace use of JMES paths.


partim:
This issue is a proposal to replace the use of JMES paths for translating OpenID claims with system using a small number of configuration attributes. This is necessary since the _jmespath_ crate won’t currently compile together with other dependencies and seems to be mostly unmaintained.

# Current Design

In OpenID, an authenticated user has a number of properties attached to it in what is known as _claims._ These are contained a JSON object with each claim as a member. The values for the claim can be any valid JSON value, i.e., it can be very complex, but in practice they seem to be either simple values or arrays of simple values.

A JMES path is used to select a, possibly nested, value from this claims object, optionally transform it into another value by way of a custom regexp substitution function and assign it to a property that can then be used within Krill.

# Proposed New Design

Instead of using a JMES path to allow users to transform claims in whatever way they like, we use a strict four step approach:

1. Pick a claim based on its name.
2. Match the claim’s value using a regular expression.
3. Substitute the claim’s value using a regexp substitution possibly using submatches from the previous step.
4. Assign the result to a target property.

Only if the named claim is available and the regular expression matches will the value be transformed and assigned. If the claim’s value is an array, the first matching element of the array is used.

This will configure similarly to the current system but with the `jmespath` property split into three properties:

* `claim` is the name of the claim,
* `match` is the regular expression to match the claim’s value against,
* `subst` is the substitution expression to apply to the match.

In addition, the current optional `dest` property allows selecting the Krill property to apply the value to.

# Example

Here’s the new version of the Azure Active Directory example from the Krill documentation:

```
[auth_openidconnect.claims]
ro_role = { claim = "groups", match = "gggggggg-gggg-gggg-gggg-gggggggggggg", subst = "readonly", dest = "role" }
rw_role = { claim = "groups", match = "hhhhhhhh-hhhh-hhhh-hhhh-hhhhhhhhhhhh", subst = "readwrite", dest = "role" }
```

PR #1232, opened 2024-09-12
Restructure authentication policies.

partim:
This PR restructures how authentication policies are used in Krill. It removes the use of Oso and its policy definition language and instead switches to simple, straightforward mappings between permissions, roles, and users.

The existing concept of _roles_ is augmented to serve as the central configuration option for limiting a user’s access to certain action and resources. Roles are now user configurable via the new `auth_roles` configuration directive. For each role, a set of permissions has to be provided. Optionally, a list of resource handles (vulgo: CAs) can be given in which case access is limited to these resources.

The authentication providers now assign one of these roles to each logged in user.

The OpenID Connect provider now only determines claims for “id,” i.e., the user name, and the “role.” Since we replaced the previous use of JMES paths with custom functions with a more stringent model of matching and substitution, the configuration had to change in a non-compatible way, anyway, so we cleaned it up a bit and switched from a map to an array for the claims.

For the config file provider, this was already possible by adding a “role” attribute. This has now been changed into a “role” field of the user details. In order to make upgrading seamless, the “role” attribute is still accepted but a deprecation warning is logged. Since the `auth_users` configuration is not used for the OpenID Connect provider any more, the `password_hash` and `salt` fields of the user details are now mandatory.

Custom policies have been removed.


partim:
At this point, I have removed the _oso_ and _jmespath_ crates and shifted all policy decisions into the new `daemon::auth::policy` module. However, the module is currently just a stub. Next we need to define the actual policy configuration and implement it.

partim:
The `auth::policy` module is gone again and has been split into private `auth::roles` and `auth::permissions` modules with their types re-exported at `auth`.

I have also implemented the new strategy outlined in #1229 with a tweak: the OpenID Connect provider now only resolves the ID and role name from the claims. The roles themselves are defined in the configuration and include both the permissions and possible limitations for the CAs access is allowed to. This means you can’t include the allowed CAs in the OpenID Connect claims any more but hopefully this is fine.

Still missing for this PR to leave draft state are documentation adjustments, both internally and in the manual.

AlexanderBand (approved):

partim:
Alright, I‘m calling this done. Now have at it ;)

src/daemon/auth/providers/config_file.rs
```diff
@@ -0,0 +1,386 @@
...
+        .unwrap();
+
+        let encoded_hash = hex::encode(hashed_hash);
+
+        // And now finally check the user, so that both known and
+        // unknown user code paths do the same work
+        // and don't result in an obvious timing difference between
+        // the two scenarios which could potentially
+        // be used to discover user names.
+        if encoded_hash != user_password_hash {
```
Koenvh1, line 212: I know this was not changed recently, but I don't like the "succeed unless it fails" flow for authentication. I prefer an explicit "fail unless it succeeds" flow, as if authentication becomes more complex in the future, then making a small mistake where the checks are all bypassed becomes easier.

src/daemon/auth/providers/openid_connect/config.rs
```diff
@@ -34,66 +33,30 @@ pub struct ConfigAuthOpenIDConnect {
...
-}
-
-#[derive(Clone, Debug)]
-pub enum ConfigAuthOpenIDConnectClaimSource {
-    ConfigFile,
-    IdTokenStandardClaim,
-    IdTokenAdditionalClaim,
-    UserInfoStandardClaim,
-    UserInfoAdditionalClaim,
+fn default_id_claims() -> Vec<TransformationRule> {
```
Koenvh1, line 41: Perhaps make it more explicit in the config/manual that this is the default (and does not need to be manually configured)

Koenvh1 (changes requested):

defaults/krill-multi-user.conf
```diff
@@ -462,30 +443,41 @@
...
-# tenant, app registration and enterprise application settings (with redirect
-# URI), users, group assignments and optional claim configuration (in the above
-# example AD was configured to expose groups as roles).
-#
-# The JMESPath expression matches on Azure AD group GUID values, taking the
-# first match it finds and then setting the "role" attribute to either readonly
-# or readwrite depending on which GUID was matched. The GUIDs for your groups
-# will be different than those used in this example, see your Krill log for the
-# GUIDs to match on.
+#   [[auth_openidconnect.claims]]
```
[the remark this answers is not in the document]
partim: Fixed.

defaults/krill-multi-user.conf
```diff
@@ -233,146 +200,160 @@
...
+#
+#                              Objects never match.
+#
+#                              If the 'match' field in missing, any simple
+#                              values matches and is used as is, i.e., the
+#                              expression in the 'subst' field is ignored.
+#                              For array claim values, the first element is
+#                              used.
+#
+#   subst               No     This field describes a transformation of a
```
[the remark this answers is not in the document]
partim: That’s actually not true: If there is no ”match” then the claim’s value is used as is and “subst” is ignored.

I decided against a simple way to set a “static role” – I felt like allowing everyone access should be a very rare case?

src/daemon/auth/providers/openid_connect/claims.rs
```diff
@@ -0,0 +1,464 @@
...
+                TransformationRule::Match(rule) => {
+                    if let Some(res) = self.process_match_rule(rule)? {
+                        return Ok(res)
+                    }
+                }
+            }
+        }
+
+        Err(Self::internal_error(
+            format!("OpenID Connect: no value found for '{}' claim.", dest),
```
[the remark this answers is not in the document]
partim: Hm, good point. I’m changing it to generate a better error message one level up.

src/daemon/auth/providers/config_file.rs
partim: I agree with your point, but in this case the code would become super unreadable. It would drift right badly and all the errors would have to be created in else branches far away from the check.

src/daemon/auth/providers/openid_connect/config.rs
partim: I added a heading above the text in the manual to make it easier to find.

Koenvh1 (approved):

PR #1233, opened 2024-10-28
Integrate roto-api in Krill replacing RISwhois

Koenvh1:
This PR replaces the downloading and parsing of RISwhois files in Krill with the roto-api.

Requires https://github.com/NLnetLabs/rpki-rs/pull/306 to work


src/commons/bgp/analyser.rs
```diff
@@ -1,83 +1,143 @@
...
+    bgp_api_uri: String,
 }
 
 impl BgpAnalyser {
     pub fn new(
-        ris_enabled: bool,
-        ris_v4_uri: &str,
-        ris_v6_uri: &str,
+        bgp_api_enabled: bool,
+        bgp_api_uri: &str,
```
partim, line 29: This should probably take a `String` and be cloned if necessary by the caller.

src/commons/bgp/iptree.rs
```diff
@@ -2,9 +2,9 @@ use std::{collections::HashMap, ops::Range};
 
 use intervaltree::IntervalTree;
 
-use rpki::repository::resources::ResourceSet;
+use rpki::repository::resources::{Addr, AddressRange, ResourceSet};
 
-use crate::commons::api::TypedPrefix;
+use crate::commons::api::{Ipv4Prefix, Ipv6Prefix, TypedPrefix};
 
 //------------ IpRange -----------------------------------------------------
```
partim, line 9: Would it make sense to clean up the range code in rpki-rs further and then use its types directly?

Koenvh1: For that there would first need to be a V4/V6 range in rpki-rs (which strikes me as a good idea). However, the Krill counterparts are used far and wide throughout the system and _slightly_ differ implementation-wise, so I am not sure if we want to do that as part of this PR.

partim (approved):

PR #1237, opened 2024-11-12
Improve TA handling

Koenvh1:
This PR improves the interaction with the TA and the "online" TA, as well as the documentation on how to use the trust anchor support in Krill, to align the behaviour of Krill with the expected behaviour.

Because this is all based around timings, I tested it by hand using faketime.


partim:
This PR was superseded by #1281.

PR #1238, opened 2024-11-19
Add PostgreSQL database support

partim:
This PR adds support for using the PostgreSQL database as a storage backend for Krill in preparation for supporting fail-over scenarios for higher availability.

The PR integrates the storage abstraction previously implemented via the _kvx_ crate back into Krill to allow a tight coupling of storage primitives with Krill’s requirements.

We have opted to not move the storage code to async functions as proposed in #1152. We think that storage should generally be fast enough that the gain from using async isn’t all that big. Using a thread pool and sync functions makes it easier to reason about what’s going on and should keep the code simpler.


partim:
I moved the PR back to draft status as I intend to make some changes to how the database backend works. Currently it uses transactions and restarts them if they fail which may result in unexpected data changes if the restarted transaction doesn’t re-read the current status. The file system backend uses locks to prevent concurrent access altogether. This can also be achieved in PostgreSQL with row level locks, so I want to switch to that strategy to have the same semantics in both backends.

PR #1239, opened 2024-12-02
Add support for Ubuntu Noble and remove support for Ubuntu Xenial, Bionic, and Debian Stretch

Koenvh1:
This PR adds support for Ubuntu Noble and removes support for Ubuntu 16.04, 18.04 and Debian 9.


partim:
Do I see it right that the PR also removes packaging for old Debian and Ubuntu? Should that be reflected in the title?

PR #1243, opened 2024-12-10
Improve the feedback provided by `krillta proxy signer show-request`

Koenvh1:
This changes the feedback provided by `krillta proxy signer show-request` to include information about certificate expiration and renewal time:

```
$ krillta proxy signer show-request
-------------------------------
nonce: 06d7845b-61aa-4ce9-97cf-d7942a1a5a9a
-------------------------------

Certificates will be reissued 80 weeks before expiry.
The current certificate expires on 2039-12-05T12:47:15+00:00.
Ergo the certificate is eligible for renewal on 2038-05-24T12:47:15+00:00.

NOTE: Use the JSON output for the signer.
```

Additionally, error when the TA and Krill config don't match:
```
$ krillta signer -c ta.conf --format text process --request request.json
TA timing config between krillta and krill do not match!
```

This does change the format of those request files (two additional fields), which are now required.

### DIFF
--- a/src/ta/common.rs
+++ b/src/ta/common.rs
@@ -7,6 +7,7 @@ use std::{
 };
 
 use bytes::Bytes;
+use chrono::TimeDelta;
 use rpki::{
     ca::{
         idexchange::{ChildHandle, RecipientHandle, SenderHandle},
@@ -35,6 +36,8 @@ use crate::{
         PublishedManifest, PublishedObject, UsedKeyState,
     },
 };
+
+use super::TaTimingConfig;
 
 //------------ TrustAnchorObjects ------------------------------------------
 
@@ -553,6 +556,8 @@ impl fmt::Display for TrustAnchorSignedRequest {
 pub struct TrustAnchorSignerRequest {
     pub nonce: Nonce, // should be matched in response (replay protection)
     pub child_requests: Vec<TrustAnchorChildRequests>,
+    pub timing: TaTimingConfig,
+    pub renew_time: Option<Time>
 }
 
 impl TrustAnchorSignerRequest {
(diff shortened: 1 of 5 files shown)


src/ta/common.rs
```diff
@@ -599,7 +604,17 @@ impl fmt::Display for TrustAnchorSignerRequest {
...
             writeln!(f)?;
-        }
+        }   
+        writeln!(f, "Certificates will be reissued {} weeks before expiry.", 
+            self.timing.issued_certificate_reissue_weeks_before)?;
+        if let Some(rt) = self.renew_time { 
+            writeln!(f, "The current certificate expires on {}.", rt.to_rfc3339())?; 
+            if let Some(weeks) = TimeDelta::try_weeks(self.timing.issued_certificate_reissue_weeks_before) {
+                let t = rt - weeks;
+                writeln!(f, "Ergo the certificate is eligible for renewal on {}.", t.to_rfc3339())?;
```
ximon18, line 614: I would drop the word "Ergo", I would be surprised if it is commonly known amongst non-native English speakers, even among native English speakers.

src/ta/common.rs
```diff
@@ -553,6 +556,8 @@ impl fmt::Display for TrustAnchorSignedRequest {
 pub struct TrustAnchorSignerRequest {
     pub nonce: Nonce, // should be matched in response (replay protection)
     pub child_requests: Vec<TrustAnchorChildRequests>,
+    pub timing: TaTimingConfig,
```
partim, line 559: Would it make sense to make this make this field optional and then allow both this one and renew_time to be absent for compatibility?

partim:
This PR was superseded by #1282.

PR #1246, opened 2025-01-08
Replace once_cell::OnceCell with std::sync::OnceLock.

partim:
This PR removes the dependency on `once_cell` since all the relevant functionality is now in `std`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1187,7 +1187,6 @@ dependencies = [
  "kvx",
  "libflate",
  "log",
- "once_cell",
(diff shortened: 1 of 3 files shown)


Koenvh1 (approved):
I was first wondering why you used `OnceLock` rather than `OnceCell`, but I'm assuming it's because of the thread safety.

partim:
Correct. The crate has `unsync::OnceCell` and `sync::OnceCell`, but `std` decided to rename the latter into `OnceLock`.

PR #1250, opened 2025-01-30
Allow for reinitialising of the TA cert

Koenvh1:
Previously, once the TA cert was created, there was no way to reinitialise it. That meant that any changes regarding the rsync URI, HTTPS URI, validity, etc. could not be made. This PR adds a `--force` switch to allow the signer to be created again, provided the same key is used, as well as a convenient `reinit` shorthand.

```
koen@beta:~/Code/krill$ target/debug/krillta signer reinit --help
Reinitialise an already initialised signer

Usage: krillta signer reinit [OPTIONS] --proxy-id <path> --proxy-repository-contact <path> --tal-rsync <rsync URI> --private-key-pem <path>

Options:
  -i, --proxy-id <path>                   Path to the proxy ID JSON file
  -r, --proxy-repository-contact <path>   Path to the proxy repository contact JSON file
      --tal-rsync <rsync URI>             The rsync URI used for TA certificate on TAL and AIA
      --tal-https <HTTPS URI>             The HTTPS URI used for the TAL
      --private-key-pem <path>            The private key for the already initialised signer in PEM format
      --initial-manifest-number <number>  Set the manifest number [default: 1]
  -h, --help                              Print help
```

PR #1255, opened 2025-03-05
Allow for the reissuance of TA certificates (including HSM support)

Koenvh1:
This is an alternative implementation of https://github.com/NLnetLabs/krill/pull/1250 that allows for TA certificates to be reissued in a different (arguably nicer) way.

```
krillta signer -c ta.conf reinit --help
Reissue the TA certificate

Usage: krillta signer reinit [OPTIONS] --proxy-id <path> --proxy-repository-contact <path> --tal-rsync <rsync URI>

Options:
  -i, --proxy-id <path>                  Path to the proxy ID JSON file
  -r, --proxy-repository-contact <path>  Path to the proxy repository contact JSON file
      --tal-rsync <rsync URI>            The rsync URI used for TA certificate on TAL and AIA
      --tal-https <HTTPS URI>            The HTTPS URI used for the TAL
  -h, --help                             Print help
```

It will create a new certificate using the TA timing config provided in the config. The URIs can be adjusted as well. The public key must stay the same.


src/daemon/ca/manager.rs
```diff
@@ -500,8 +500,10 @@ impl CaManager {
 
     /// Adds the associated signer to the proxy.
     ///
-    /// Errors if there is no proxy or the proxy already has a signer.
-    pub fn ta_proxy_signer_add(
+    /// Errors if:
+    /// - there is no proxy
+    /// - the proxy has a signer
+    pub async fn ta_proxy_signer_add(
```
partim, line 506: The method doesn’t need to be async, so it shouldn’t be. (Also, maybe keep the new docstring)?

src/daemon/ca/manager.rs
```diff
@@ -512,6 +514,22 @@ impl CaManager {
         Ok(())
     }
 
+    /// Updates the associated signer to the proxy.
+    ///
+    /// Errors if:
+    /// - there is no proxy
+    /// - the proxy has no or a different signer
+    pub async fn ta_proxy_signer_update(
```
partim, line 522: Doesn’t need to be async.

src/daemon/taproxy.rs
```diff
@@ -736,6 +767,23 @@ impl TrustAnchorProxy {
         }
     }
 
+    fn process_update_signer(
+        &self,
+        signer: TrustAnchorSignerInfo,
+    ) -> KrillResult<Vec<TrustAnchorProxyEvent>> {
+        if let Some(s) = &self.signer {
+            if s.ta_cert_details.cert().key_identifier() == 
+                signer.ta_cert_details.cert().key_identifier() {
```
partim, line 776: Nitpick: The opening brace should be on its own line if the condition has a line break.

src/daemon/taproxy.rs
```diff
@@ -736,6 +767,23 @@ impl TrustAnchorProxy {
...
+                // It is not possible to add a signer that has a different
+                // public key
+                return Ok(vec![
+                    TrustAnchorProxyEvent::SignerAdded(signer)
```
partim, line 780: Should this have its own event rather than re-using Signer Added? (Honest question, I’m not sure.)

partim (approved):

partim (approved):

PR #1264, opened 2025-03-31
Clearly separate between “business logic” and HTTP service.

partim:
This PR more clearly separates code between Krill “business logic” – now bundled in a `KrillManager` –, and the HTTP server code that serves the API. The former now lives in the `server` module, the latter in the `daemon` module together with all the code to spin up a Krill daemon driving the HTTP server.


src/daemon/http/request.rs
```diff
@@ -0,0 +1,527 @@
...
+                request: self.request,
+                server: self.server,
+                limits: self.limits,
+            },
+            self.auth
+        )
+    }
+
+    /// Splits the request into the server and raw Hyper request.
+    pub fn proceed_raw(self) -> (&'a HttpServer, HyperRequest) {
```
ximon18, line 187: This doesn't feel like it should be called proceed because it doesn't do anything, this is just an into fn, maybe call it into_parts() or something?

src/daemon/http/request.rs
```diff
@@ -0,0 +1,527 @@
...
+            },
+            self.auth
+        ))
+    }
+
+    /// Permits the request to the next processing stage.
+    ///
+    /// Returns [`AuthedRequest`] and [`Actor`] without requiring any
+    /// permissions whatsoever.
+    pub fn proceed_unchecked(
```
ximon18, line 173: When I read this at the call site I was confused, if it doesn't check anything why call it, what does it do? This is more like an `into_authenticated_reqyest()` fn. I can't think of a good name for it though. As it does the same as `proceed_permitted()` without the permission check, maybe an alternative could be to just call `check_permission()` from the call site followed by `into_authenticated_request()`.

src/daemon/http/request.rs
ximon18: After looking at `proceed_unchecked()` and thinking that could be called `into_authenticated_request()` maybe call this `into_raw_request()` ?

src/daemon/http/request.rs
```diff
@@ -0,0 +1,527 @@
...
+    pub fn empty(self) -> Result<&'a HttpServer, Error> {
+        if self.request.body().size_hint().upper() != Some(0) {
+            // XXX Maybe this should have its own error?
+            return Err(Error::PostTooBig)
+        }
+        Ok(self.server)
+    }
+
+    /// Returns the raw bytes of the request body.
+    pub async fn api_bytes(self) -> Result<(&'a HttpServer, Bytes), Error> {
```
ximon18, line 220: Why `api_bytes()` and not `into_bytes()` given that it would work for any request, not just API calls? (aside from the post limit check)

ximon18: Ah I see below `rfc6492_bytes()` ... okay, `into_raw_bytes()` then?

src/daemon/http/request.rs
```diff
@@ -0,0 +1,527 @@
...
+    /// The limits for reading the body of the request.
+    limits: BodyLimits,
+}
+
+impl<'a> AuthedRequest<'a> {
+    /// Ensures the body is empty.
+    pub fn empty(self) -> Result<&'a HttpServer, Error> {
+        if self.request.body().size_hint().upper() != Some(0) {
+            // XXX Maybe this should have its own error?
+            return Err(Error::PostTooBig)
```
Koenvh1, line 214: I think this would be nicer with its own error - also GET request can technically have a body, they're not limited to POSTs

Koenvh1 (approved):
I agree with @ximon18 on the naming, but I also don't have any better suggestions myself (and in practice I'll probably get used to it anyhow).

src/daemon/http/request.rs
partim: I found `proceed_permitted` already too long.

I picked _proceed_ here as meaning “continue with what you were doing after a brief interruption,” in this case for the permission check, and then decided to keep it persistently. I don’t like using `into` since some of those functions actually do something more than just conversion.

src/daemon/http/request.rs
partim: It can’t be called `into_authenticated_request` because that implies that authentication was checked. I want the three methods to have the same naming scheme to make it clear that you have to call one of the three to continue.

src/daemon/http/request.rs
partim: The name is just what is was called before. I didn’t find it important enough to rename it. (They are all not great since they aren’t expressing that they actually do something and not just return something they already have.)

src/daemon/http/request.rs
partim: Fixed. We should probably distinguish here between API errors with a JSON error body and non-API errors with a text body. But we’ll leave this for when we re-design all the errors.

src/daemon/http/request.rs
partim: I have renamed the body reading methods into `read_*`.

PR #1267, opened 2025-04-07
Make Clippy happy

Koenvh1:


src/cli/options/ca.rs
```diff
@@ -1,5 +1,4 @@
 /// Commands related to managing a CA.
-
```
partim, line 2: Ah, the error is actually that the docstring should start with `//!` to apply to the whole module. And then the empty line should remain.

PR #1276, opened 2025-05-27
Randomise the entries on the manifest

Koenvh1:
This adds randomisation of the manifest as discussed in https://mailarchive.ietf.org/arch/msg/sidrops/gaX6rFkiwBfKKeE6PXAu2Pz8BIA/


Koenvh1:
This is redundant as (by chance) entries are already randomised on the manifest

PR #1282, opened 2025-06-05
Improve feedback given by `krillta proxy signer show-request`.

partim:
This changes the feedback provided by `krillta proxy signer show-request` to include information about certificate expiration and renewal time:

```
$ krillta proxy signer show-request
-------------------------------
nonce: 06d7845b-61aa-4ce9-97cf-d7942a1a5a9a
-------------------------------

Certificates will be reissued 80 weeks before expiry.
The current certificate expires on 2039-12-05T12:47:15+00:00.
Ergo the certificate is eligible for renewal on 2038-05-24T12:47:15+00:00.

NOTE: Use the JSON output for the signer.
```

### DIFF
--- a/src/api/ta.rs
+++ b/src/api/ta.rs
@@ -7,6 +7,7 @@ use std::{
 };
 
 use bytes::Bytes;
+use chrono::TimeDelta;
 use rpki::{
     ca::{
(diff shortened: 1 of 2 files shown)


partim:
This attempt breaks the case where both TA proxy and signer are part of the Krill server itself. A new approach is being explored in #1283.

PR #1309, opened 2025-07-31
Initialize properties store to distinguish 0.14 from 0.15.

partim:
This PR adds a step to the start of the Krill daemon that initializes the property store with the current version if it hasn’t been initialized earlier. It also assumes that an uninitialized property store with no per-store version information indicates that the data is from version 0.14.0 and migrates the stores lock directories accordingly.


partim:
I tested upgrading data from 0.13 and 0.14 manually.

PR #1322, opened 2025-09-24
Add support for UNIX sockets

Koenvh1:
The idea is to use a local unix domain socket and authenticate via the local user (root by default, other users are configurable). This allows for the use of `krillc` without explicitly supplying a token.

The configuration looks like this:
```
unix_socket = "/tmp/krill.sock"
unix_users = { root = "admin", koen = "readonly" }
```

This allows for running krillc without KRILL_CLI_TOKEN:
```
koen@beta:~/Code/krill/target/debug$ ./krillc list
testca
```


partim (commented):
I feel like this is the point where we should start having separate HTTP client implementations for the client and server portions. So my suggestion would be to copy `src/commons/httpclient.rs` into a new `src/cli/httpclient.rs` and only modify that to be able to use Unix sockets.

Maybe it would be useful to then also create a `HttpClient` type and move the possible socket and possible token in there (this would also allow to have either a token or a socket). This way you don’t need to do splitting of the URL – I feel uneasy relying on the socket having a filename ending in .sock or else stuff breaking – but rather work with a `http://localhost` prefix.

Koenvh1:
I'm fine with pulling the two clients apart, although I would rename them to avoid mixing them up. Not sure about whether a generic type would really work for this, though a type that contains the socket separately rather than parsing the URI over and over is something I can get behind.

I'm happy to hear what you suggest as URI scheme - keep `unix://something/that/could/be/a/path` and make every request to `http://localhost/api/do/something?

src/cli/client.rs
```diff
@@ -32,80 +41,161 @@ use crate::commons::httpclient::Error;
...
+                {
+                    builder = builder.danger_accept_invalid_certs(true);
+                }
+                uri
+            }
+            #[cfg(unix)]
+            ServerUri::Unix(socket_path) => {
+                builder = builder.unix_socket(socket_path);
+                
+                // XXX Do we actually need this if we use an http: base URI?
```
partim: I think this can go but needs to be tested.

Koenvh1:
Note to self: Fix Krill config generation for packages and configuring the Krill client

PR #1326, opened 2025-10-02
Be more flexible on `relations` existing in the bgp-api JSON

Koenvh1:
This resolves an issue mentioned on [Discord](https://discord.com/channels/818584154278199396/818587059073318912/1423115914005582025) where it would return no announcement info even though there is announcement info (but it is available, but there is no information about its relations).


Koenvh1:
I believe this should be in-line with what you expected to see, or am I missing something else?
image

xrobau:
Looking much better! 
image

PR #1377, opened 2026-05-06
Minor edit to documentation for the cli for krillc config

ctybrds:
Reflects current operation of the krillc config command. Have not been able to get the --attr flag to work yet.


partim (approved):
Thank you for the PR!

PR #1383, opened 2026-06-19
Introduce `StorageSystem`.

partim:
This PR introduces an additional layer for the key value store that can keep global information. This is another step in preparation for introducing support for database storage which will have connection pools and such that are shared system-wide.

Since this storage system isn’t actually used yet, the PR for now is just a massive refactoring and does not introduce any functional changes.

This PR makes the `storage_uri` (aliased as `data_dir`) config option mandatory and drops both it being read from an environment variable or defaulting to `./data`. Because of this, **this is a breaking change.**

### DIFF
--- a/src/commons/storage/backends/memory.rs
+++ b/src/commons/storage/backends/memory.rs
@@ -275,36 +275,6 @@ impl Store {
         self.namespace.scopes().clear();
         Ok(())
     }
-
-        /*
-    pub fn migrate_namespace(
-        &mut self, target: &Ident
-    ) -> Result<(), Error> {
-        if !self.namespace.locks().is_empty() {
-            return Err(Error::PendingLocks);
-        }
-        let mut namespaces = MEMORY.namespaces.lock().expect("poisoned lock");
-        let new_key = (self.namespace.ns_key.0.clone(), target.into());
-        let new = namespaces.entry(new_key.clone()).or_insert_with(|| {
-            MemoryNamespace::new(new_key).into()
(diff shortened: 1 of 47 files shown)


src/commons/storage/backends/memory.rs
```diff
@@ -239,6 +304,7 @@ impl Store {
         self.namespace = new.clone();
         Ok(())
     }
+        */
```
Koenvh1: Want to keep this for later? Seems a bit redundant

src/commons/storage/test.rs
```diff
@@ -130,14 +127,21 @@ testfns! {
...
+
         store.store(None, KEY, &CONTENT).unwrap();
+        eprintln!("3");
         assert!(store.has(None, KEY).unwrap());
+        eprintln!("4");
 
         store.drop_key(None, KEY).unwrap();
+        eprintln!("5");
         assert!(!store.has(None, KEY).unwrap());
+        eprintln!("6");
```
Koenvh1: I assume you did some print debugging, are they worth keeping for the future?

Koenvh1 (commented):
Looks good to me :-)

src/commons/storage/backends/memory.rs
partim: Removed.

src/commons/storage/test.rs
partim: Nope. Removed.

PR #1385, opened 2026-07-09
Remove unused dependencies and cleanup dependency list.

partim:
This PR removes the intervalltree and lazy_static dependencies and moves regex to those only pulled in on multi-user. It also reformats the dependency list to avoid long lines.


partim:
TOML 1.1 (which is needed for line breaks within curly braces) is only available from Rust 1.94, so we need to do this later.

Issue #1397, opened 2026-09-11
Empty resource certificate requested and issued


timbru:
Hi,

We had an occurrence of empty resource certificates appearing under a RIPE NCC member using Krill 0.16.0.

The issue seems to be that in case Krill acts as a parent and delegates to a child, and all the resource entitlements for that child are removed but the parent-child relationship is still intact, that child can continue to request certificates with an empty resource set.

This is not a security issue, but it generates noise as empty resource sets are not allowed (or meaningfull).

There is likely more than one way to fix this. The parent could pro-actively revoke all certificates and refure to issue new empty resource set certificates for a child that has no resources. If this is done re-actively, then the child could conclude it needs to request revocation of all keys under a resource class that has no more resources and stop requesting certificates. Removing the child altogether may be the easiest fix (and this works) - however, that also means that this is permanent - the child would have to be added again if they get new resources in future. Suspending the child may work however (if that functionality is still in there) - so long as it can only be unsuspended when new resources appear.

Koenvh1:
Hmmm, interesting catch. I will look into it at some point (though removing the child automatically seems a bit scary).

timbru:
Yeah, indeed - so re-purposing the suspend logic could get you there more safely - if it unsuspends when new resource entitlements are added.

timbru:
Oh also keep in mind, this is really about the intersection (edit: I said union earlier but that was a mixup on my part) of the resources that the parent CA actually holds (which might have been shrunk) and what they delegate to a child. So it may not be a case where the child resource entitlements were all removed, but a case where the parent no longer holds those.

Koenvh1:
Yes I was thinking about that as well (and I knew you meant intersection). Suspending would not work as the next time the child comes around to ask they would be unsuspended again. I assume most of the time this is because of some temporary state (grandparent changing resources) so I think not issuing a new certificate should be enough(?). 

I started working on it in https://github.com/NLnetLabs/krill/pull/1398. Testing this is difficult though.

Koenvh1:
@timbru I assume you are using a different non-Krill child that causes this behaviour? Could you test whether #1398 fixes it?

PR #1398, opened 2026-09-15
Prevent issuance without resources

Koenvh1:
Relates to https://github.com/NLnetLabs/krill/issues/1397

I _think_ this works but testing it is difficult. The test currently passes but does nothing useful.


Koenvh1:
@ximon18 Maybe you have any ideas on how to best test this?

timbru:
I think this should test the previous certificate is revoked as well. Based on memory as previous implementer this does not happen automatically. It could need to be triggered by changes in resources associated with the child, or the issuing CA.

Refusing to issue an empty resource certificate is also a good idea, but by itself is not enough. Without pro-active revocation this would leave the previous certifcate in place. In addition, it would be better if the child does not request an empty certificate when its entitlements become the empty resource set for a given resource class, but instead drops the resource class altogether (and requests revocation for any key it still had), and only add the resource class again when the resource entitlements are non-empty.

Sorry, I can't be of much more help here, but I hope this is useful.

